PR #43, opened 2016-10-26
optmize VPN scripts

MelvinTo:
1. removed hard code /home/pi/…
2. added global env: FIREWALLA_HOME
3. fix code to get external ip with external-ip library

next step: auto select local DNS


jerrchen:
suggestions:

1. why adding another library for external ip?  there is one already called public-ip.  And also, when the device first awake, it calls the checkin function, which will return the publicIp to the user. see below.  The field is .publicIp

    checkin(ip,eid,data,callback,context) {
        logger.debug("Checkin Device",eid,data,null);
        let key = 'bone:device:checkin:'+eid;
        data.ts = Date.now()/1000;
        data.ip = ip;
        this.redis.hmset(key,data,(err,msg)=> {
            logger.debug("DeviceManager:Checkin",key,data);
            DDNS.create(eid,null,data.ip,data.mac,false,(err,dns)=> {
               let rdata = {
                   'license': {
                   },
                   'needUpgrade':false,
                   'needReboot':false,
                   'features': {
                       'vpnFeature':true,
                       'ipsFeature':true,
                       'idsFeature':true,
                   },
                   'ids': {
                       'version':'1.0',
                   }
                }

                if (err==null && dns) {
                   rdata.ddns = dns;
                }

                if (ip) {
                   **rdata.publicIp = ip;**
                }

And this publicIp is set in SysManager

    checkIn(callback) {
        this.getSysInfo((err,data)=>{
            bone.checkin(this.config,data,(err,data)=>{
                console.log("CheckedIn:", data);
                if (data.ddns) {
                    this.ddns = data.ddns;
                    rclient.hset("sys:network:info", "ddns", JSON.stringify(data.ddns), (err, result) => {
                         if (callback) {
                             callback(null,null);
                         }
                    });
                }
                **if (data.publicIp) {
                    this.publicIp = data.publicIp;
                    rclient.hset("sys:network:info", "publicIp", JSON.stringify(data.publicIp), (err, result) => {
                    });
                }**
            });
        });

    }

2. Also reminder when add node modules, it need to be added in the main-run script.  This is for painful upgrades.

MelvinTo:
this is to decouple fishbone and pi when doing testing. so that i can get the public ip without connecting to fishbone. It will try to use fishbone first, and use external-ip library when failed to connect to fishbone.

MelvinTo:
added migration scripts for external ip

jerrchen:
There is a existing external ip library already, that one not working ?

PR #44, opened 2016-10-28
Replace public-ip with external-ip

MelvinTo:


MelvinTo:
commit "upgrade network module to latest (0.3.0)" is also added.

PR #57, opened 2016-11-11
1. support shadowsocks at backend 2. adding debugging rest-api (swagger) for easier debugging purpose

MelvinTo:


jerrchen:
Will the swagger code be build in the image?  

Also, is there an off for ss?

MelvinTo:
* swagger code will be built in the image, but off by default.
* ss is by default off

TODO: need to add code to turn ss on via app.

Manual steps to turn on ss.

1. sudo apt-get install python-pip
1. sudo pip install shadowsocks
1. cd api
1. npm install
1. npm start # this is to start the firewalla restful api, this api can be stopped after turn ss on
1. curl -XPOST localhost:3000/v1/ss/config/renew # Create configuration file for shadowsocks
1. curl -XPOST localhost:3000/v1/ss/start # Start shadowsocks
1. check if shadowsocks process is running by "ps aux | grep ssserver"
1. curl localhost:3000/v1/ss/config # To get shadowsocks configuration, random password

PR #61, opened 2016-11-21
add authentication to apis except message api

MelvinTo:


jerrchen:
How is authentication done?   Is it over jwt tokens? If it is, what kind of token it is?

jerrchen:
The new npm modules need to be in the upgrade section, otherwise existing deployments will break

jerrchen:
I will merge now, the upgrade section need to be fixed for sure

PR #65, opened 2016-11-26
Refine buildraw

niulin:
Refine buildraw. Now it accepts a parameter "-t" to use npm mirror provided by taobao.org which is much faster in China. Also added "-n" option to bypass the time consume apt-get jobs. 
Also improved script robustness and user friendliness.


MelvinTo:
Bravo!

PR #84, opened 2016-12-20
	load debug from pi to app

MelvinTo:


jerrchen:
nice, the power of redis

PR #95, opened 2017-01-06
use api to replace ui

MelvinTo:


jerrchen:
this will prevent accidental release when we are not ready

PR #99, opened 2017-01-09
improve direct mode

MelvinTo:


jerrchen:
give me one more day on this, need to keep main clean to bring up zero

PR #107, opened 2017-01-22
support running firewalla on docker environment

MelvinTo:
not able to link with app yet.


jerrchen:
your build failed.  need to fix it later.  will merge it now

jerrchen:
also isProduction may not work all the time, the enviroment variable some times don't get passed, so need to check if that production file is there.

MelvinTo:
Failed is because the user used in travis ci is "travis", not "pi". To fix it, will add a minor change to use $USER instead of pi.

MelvinTo:
But not sure whether "travis" is really supported throughout the entire code base, the current build will pass with the minor fix above

MelvinTo:
@xuanxuandog

PR #109, opened 2017-01-23
replace pi with $USER when accessing bro

MelvinTo:


MelvinTo:
typo, tt => $USER

jerrchen:
isProduction()  function, yours is too expensive.  need a cache.  the isProduction function likely to be used many times.  this is the reason the original code cache it in a class var.

PR #133, opened 2017-02-16
support block by mac

MelvinTo:


etc/sysctl.conf
```diff
@@ -32,13 +32,15 @@ kernel.panic=10
 
 # Uncomment the next line to enable packet forwarding for IPv4
 net.ipv4.ip_forward=1
+net.ipv4.conf.all.route_localnet=1
```
jerrchen: which packet will use the 127.0.0.1 as src?  if not necessary, we should not change these configs.

scripts/main-run
```diff
@@ -153,6 +153,9 @@ if [[ ! -f /.dockerenv ]]; then
 fi
 
 cd net2
+
+# this config is necessary for forwarding DNS requests to dnsmasq
+sudo sysctl -w net.ipv4.conf.all.route_localnet=1
```
jerrchen: you already added this to sysctl right?

PR #142, opened 2017-03-06
	add sudo prefix to bitbridge4/6/6a

MelvinTo:


jerrchen:
sudo was intentionally removed due to memory usage once there are a lot of devices.

PR #148, opened 2017-03-11
capability cap_net_admin is required for bitbridge

MelvinTo:


MelvinTo:
the cap_net_admin patch needs to be applied on release branch, or bitbridge4 wont run on 3.4 kernel.

PR #150, opened 2017-03-13
patch to fix NIC performance issue

MelvinTo:


jerrchen:
release_pi_3_0 is now active.

PR #153, opened 2017-03-16
Move upgrade code to a separate systemctl service

MelvinTo:
fireapi will auto start
along with other minor changes.
dhcpdump is not ready yet.

### DIFF
--- a/scripts/fireupgrade.sh
+++ b/scripts/fireupgrade.sh
@@ -24,8 +24,7 @@ done
 # very likely to fail
 
 if [[ $branch =~ release.* ]]; then
-  sudo -u pi git fetch origin $branch
-  sudo -u pi git reset --hard origin/$branch
+  sudo -u pi git fetch origin $branch && sudo -u pi git reset --hard FETCH_HEAD
   echo "on release"
 
   # in case there is some upgrade change on firewalla.service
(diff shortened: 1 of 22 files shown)


scripts/fireupgrade.sh
```diff
@@ -0,0 +1,39 @@
...
+    sleep 1
+done
+
+
+# continue to try upgrade even github api is not successfully.
+# very likely to fail
+
+echo "upgrade on branch $branch"
+
+(sudo -u pi git fetch origin $branch && sudo -u pi git reset --hard FETCH_HEAD) || exit 1
```
MelvinTo: use "git reset --hard FETCH_HEAD" instead of "git reset --hard origin/$branch"

FETCH_HEAD is more accurate than origin/$branch. I had several times that origin/$branch is not latest as FETCH_HEAD (not 100% reproducible), which might be an issue why user's pi didn't get upgraded successfully.

jerrchen: ack

jerrchen: 1. will this blow up old systems?

jerrchen: Is DPCP dump used some where?

PR #195, opened 2017-06-10
Feature: support both spoof and dhcp

MelvinTo:


net2/main.js
```diff
@@ -27,34 +27,19 @@ let bone = require("../lib/Bone.js");
...
-secondaryInterface.create(config,(err,ip,subnet,ipnet,mask)=>{
-    if (err == null) {
-        log.info("Successful Created Secondary Interface");
-        sysManager.secondaryIp = ip;
-        sysManager.secondarySubnet = subnet; 
-        sysManager.secondaryIpnet = ipnet; 
-        sysManager.secondaryMask  = mask; 
-    }
-});
-
```
jerrchen: make sure secondary interface is always created.  Reason for this is, in case our customer have issues, we can always let them try this interface. (static ip)  This is useful for some hosts that just don't like to be compatible.

Issue #230, opened 2017-07-07
Make VPN Logless


jerrchen:
request from Ben

http://hack.training/create-logless-vpn/

btrebach:
If you could use a hand with any part of this, please let me know - I'd be happy to help.

jerrchen:
Super, we welcome all developers.   If you interested feel free to look at the VPN side.  Adding the logless is very simple, probably just a small modification of the vpn template
https://github.com/firewalla/firewalla/tree/master/vpn
The file to change is https://github.com/firewalla/firewalla/blob/master/vpn/server_config.txt
modify the server_config log into the log format specified in the link I posted.

jerrchen:
@btrebach thank you for the commit, issue closed.

hasanalizxc:
Hi All,

http://hack.training/create-logless-vpn/ is closed.. Can you give that article?

jerrchen:
@hasanalizxc what article?

r00t34:
I can't reach this website.

http://hack.training/create-logless-vpn/

Can you share it with me?

PR #231, opened 2017-07-08
disabled logs, fixes issue #230

btrebach:


jerrchen:
@MelvinTo see if you guys can test it out

PR #259, opened 2017-07-29
Create fix_eth0_ip.sh

jimyang2008:
## Requirement
if there is ip address on eth0 interface, save it some where
if there is no ip address on eth0 for n seconds
  configure eth0 with last successful ip address manually
this is the dhcp over lay mode


jerrchen:
possible to save also the default gateway and at least one dns?  if dhcp fails, we need to setup all of these.

scripts/fix_eth0_ip.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+ETH0_IP=$(ifconfig eth0 |grep 'inet addr'|awk '{print $2}' | awk -F: '{print $2}')
+
+SAVED_IP_FILE=/var/run/saved_ip
+
+if [[ -n "${ETH0_IP}" ]]
+then
+    /bin/rm -f ${SAVED_IP_FILE}
```
MelvinTo: Need logic to wait for N seconds before using the saved ip.

jimyang2008: wait for N seconds between *test* and *set* ? or just before *test* ?

MelvinTo: both, test before and after.

jerrchen: it is to ensure that 'dhcp' may be slow in some cases ...

PR #267, opened 2017-08-03
update check_fix_network.sh with more features

jimyang2008:
Now it can check and fix(some of) following
* ethernet
* IP
* default gw
* DNS
* github.com


jerrchen:
looking strange, lets chat off line

PR #317, opened 2017-09-06
fix various issues based on feedback from Jerry and Melvin

jimyang2008:


jerrchen:
Will this fix:

1. walla come up and no ethernet, then reboot?
2. walla come up unable to dhcp and restore old dhcp data.  (only to eth0 and don't keep eth0;0?)

PR #341, opened 2017-09-19
Enable BBR TCP congestion control

starsy:
google/bbr:
https://github.com/google/bbr

tcp_bbr: add BBR congestion control:
https://git.kernel.org/pub/scm/linux/kernel/git/davem/net-next.git/commit/?id=0f8782ea14974ce992618b55f0c041ef43ed0b78


jerrchen:
nice, a note, this will only work with Firewalla, the Raspian image has an older kernel it won't work.

starsy:
correct, only works for kernel > 4.9

PR #351, opened 2017-09-25
fix dhcp not working issue for some routers

starsy:
add delay before walla replies arp spoofing packet to router when an unoccupied ip is probed by arping from router. 

Solve the dhcp failure for China telecom routers and other routers based on udhcpd on busybox.

### DIFF
--- a/net2/SpooferManager.js
+++ b/net2/SpooferManager.js
@@ -81,9 +81,9 @@ function startSpoofing() {
       let logStream = fs.createWriteStream(spoofLogFile, {flags: 'a'});
       
       if(firewalla.isDocker() || firewalla.isTravis()) {
-        spawnProcess = spawn("sudo", [getBinary(), ifName, routerIP, myIP,'-m','-q'])
+        spawnProcess = spawn("sudo", [getBinary(), ifName, routerIP, myIP,'-m','-q','-n'])
       } else {
-        spawnProcess = spawn(getBinary(), [ifName, routerIP, myIP,'-m','-q']);
+        spawnProcess = spawn(getBinary(), [ifName, routerIP, myIP,'-m','-q','-n']);
       }
       log.info("starting new spoofing: ", getBinary(), [ifName, routerIP, myIP], {});
 


jerrchen:
did you commit the binary that support -n ?

MelvinTo:
This is just a workaround, can't be used as official solution. "-n" is already supported in current binary, but has side effect that may cause spoofing failed in some scenarios. (e.g. router port mapping on non-existing IPs)

jerrchen:
bitbrdige7 is 3 months old, it is the initial version.  wasn't the -n added much later? @niulin  verify please

starsy:
@jerrchen the current version in master branch supports -n

starsy:
@MelvinTo agree with the side effect, these scenarios should be rare, comparing to making client's network broken and considering the # of routers using busybox's built-in udhcpd, the current fix worths it.

MelvinTo:
liniu will release a better fix very soon

starsy:
discussed w/ @niulin yesterday, the incoming fix will be a similar delay approach w/ adjustable delay time params as different routers using different timeout for the arping. a one shot fix for all the routers will not be possible.

niulin:
@MelvinTo currently, there's no better solution. Either this simple solution or a "better" b7 will impact the port-mappping in the previous seconds. Already discussed with @MelvinTo , we are OK if port-mapping only disturbs spoofing for a few seconds.

MelvinTo:
1. we can wait for the incoming fix 

2. the delay **must** be deterministic in X seconds, the current fix needs to wait for next loop cycle, which is dynamic and depends on the length of the monitored host list

starsy:
- the incoming fix does not solve or make improvement on the stop spoofing issue, the same effect as -n
- don't agree w/ 'as long as any device uses NAT/static port mapping, it may be impacted.'. it only stops spoofing for 1 - 2 secs when the external ip want to reach out to NAT device's port but the device itself is currently offline, which is rare case.

PR #369, opened 2017-09-28
add fireapi-ping.sh to monitor fireapi

jimyang2008:


jerrchen:
1. integrate fireapi-ping into fireping.   no need another cronjob
2. need to have a small check to see if fireapi service is up before doing anything.  
   if (fireapi service is up) {
      do fireapi ping test
   }
3. when fail, do not reboot, just do a sudo service restart fireapi to recover.

scripts/fireapi-ping.sh
```diff
@@ -0,0 +1,67 @@
...
+    if fireapi_ping; then
+        ping_ok=1
+        break
+    fi
+    sleep $SLEEP_TIMEOUT
+    (( retry++ ))
+done
+
+if [[ $ping_ok -ne 1 ]]; then
+    logger "FireAPI ping FAILED, reboot now"
```
MelvinTo: log needs to be refined

etc/brotab
```diff
@@ -9,5 +9,6 @@
 * * * * * ( sleep 30 ; /home/pi/firewalla/scripts/fire-ping.sh  >/dev/null 2>&1 )
 * * * * * ( sleep 40 ; /home/pi/firewalla/scripts/fire-ping.sh  >/dev/null 2>&1 )
 * * * * * ( sleep 50 ; /home/pi/firewalla/scripts/fire-ping.sh  >/dev/null 2>&1 )
+* * * * * ( /home/pi/firewalla/scripts/fireapi-ping.sh  >/dev/null 2>&1 )
```
MelvinTo: let's make it run once per two minutes to avoid overlap. Also fireapi not working is a very rare situation that we can wait for two minutes if happens.

PR #397, opened 2017-10-12
Add FlowUploadSensor.js

xuanxuandog:
will upload all flows to cloud every 15 minutes


sensor/FlowUploadSensor.js
```diff
@@ -0,0 +1,209 @@
...
+let HostTool = require('../net2/HostTool')
+let hostTool = new HostTool()
+let flowTool = require('../net2/FlowTool')()
+let flowUtil = require('../net2/FlowUtil')
+let Bone = require('../lib/Bone.js')
+
+let INTERVAL_MIN = 10 //10 seconds
+let INTERVAL_MAX = 3600 //1 hour
+let INTERVAL_DEFAULT = 900 //15 minutes
+let MAX_UPLOAD_SIZE = 1073741824 //1G
```
jerrchen: I assume this is for safety.  If it is, then make the max like 20 megabytes or something.  And if partial, put an indicator somewhere and say, there are total x bytes, we only send y bytes.

And this number should not be bytes either, it should be number of flows.   In case there are a lot of p2p data, we don't want the server to get all that noise if after a certain limit.

sensor/FlowUploadSensor.js
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    getAllFlows(start, end) {
+        return async(() => {
+            let macs = await(this.getAllMacs())
+            let flows = {}
+            macs.forEach((mac) => {
+                let flow = await(this.getFlowByMac(mac, start, end))
+                if (flow != null && flow.length > 0) {
+                    flows[flowUtil.hashMac(mac)] = this.processFlow(flow, true)
```
jerrchen: let hashdebug = sysManager.isSystemDebugOn();
        let _flow = flowUtil.hashFlow(flow,!hashdebug);

This way, if debug is enabled, or people using master, we get some clear text data.   We can do debug. Otherwsie, we can't read

net2/config.json
```diff
@@ -54,6 +54,11 @@
     "FlowAggregationSensor": {},
     "DeviceNameUpdateSensor": {},
     "IPv6DiscoverySensor":{},
+    "FlowUploadSensor" : {
+      "interval":900,
+      "maxLength": 1073741824,
```
jerrchen: max number of flows please.

PR #412, opened 2017-10-16
won't let app bind when license is invalid

xuanxuandog:


sys/invitation.js
```diff
@@ -154,11 +163,10 @@ class FWInvitation {
...
             status: "expired",
           };
-        } else {
-          return {
-            status: "pending"
-          };
-        }
+        } 
+      }
+      return {
```
MelvinTo: need to continue supporting legacy license mode, since many early adopters are still using old license mode

PR #420, opened 2017-10-19
support parse multiple dhcpdump events

xuanxuandog:


extension/dhcpdump/dhcpdump.js
```diff
@@ -103,28 +103,35 @@ OPTION:  12 ( 12) Host name                 Great-Room-3
   parse(output) {
      let o =  output.split(/\r?\n/);
      let obj = {};
+     var foundMac = null
```
MelvinTo: minor suggestion: var has flaw, should use let or const instead.

extension/dhcpdump/dhcpdump.js
```diff
@@ -103,28 +103,35 @@ OPTION:  12 ( 12) Host name                 Great-Room-3
...
+      //for multiple events,
+      //when match the split line '---------------------------' between two adjacent events,
+      //need reset the foundMac in case one dhcp event has mac without hostname,
+      //then the next event has hostname without mac
+      if (line.indexOf("------------------------------------------------") >= 0) {
+        foundMac = null
+      }
     }
+
+    return obj
```
MelvinTo: general comment: this implementation is okay, and a better design is spliting data to multiple sub-datas (each sub data contains a single dhcp event), and have this.parse(subdata) to extract the mac and hostname. This way the parse function will specifically focus on parsing single dhcp event, which makes the logic simple.

MelvinTo:
```javascript
let msgs = message.split("------------------------------------------------")

msgs.forEach((m) => {
  let obj = this.parse(m)
  callback(obj)
})
```

PR #439, opened 2017-10-30
avoid creating duplicated policies

xuanxuandog:


alarm/PolicyManager2.js
```diff
@@ -195,8 +198,17 @@ class PolicyManager2 {
...
     callback = callback || function() {}
-
-    this.savePolicy(policy, callback);
+    async(()=>{
+      //FIXME: data inconsistence risk for multi-processes or multi-threads
+      let policies = await(this.getSimilarPolicies(policy))
+      if (policies && policies.length > 0) {
+        log.info("policy with type:" + policy.type + ",target:" + policy.target + " already existed")
+        policy.pid = policies[0].pid
+        callback(null, policy.pid)
```
MelvinTo: this might introduce a issue that multiple alarms are bound to same policy, need to think about how to handle this use case.

alarm/PolicyManager2.js
```diff
@@ -230,6 +242,25 @@ class PolicyManager2 {
...
+  getSimilarPolicies(policy) {
+    let pm2 = this
+    return async(() => {
+      return new Promise(function (resolve, reject) {
+        pm2.loadActivePolicys(200, (err, policies)=>{
+          if (err) {
+            reject(err)
+          } else {
+            if (policies) {
+              resolve(policies.filter(p => p.type == policy.type && p.target == policy.target))
```
MelvinTo: better use === instead of ==

alarm/PolicyManager2.js
xuanxuandog: What about let client create and save duplicate policies each time, but will check whether there are similar policies with same type and target when doing enforcement? Also, when client need delete a policy, need delete all similar policies.

MelvinTo: this will make it even more complex, need a simpler solution.

alarm/PolicyManager2.js
```diff
@@ -64,11 +67,24 @@ class PolicyManager2 {
         log.info("got policy enforcement event:" + event.action + ":" + event.policy.pid)
 
         if (event.action && event.action == 'enforce') {
-          this.enforce(event.policy)
+          try {
```
MelvinTo: try catch doesn't work here, have to be inside async scope

MelvinTo: outside of async scope, "try" will not catch the errors from promise

MelvinTo: and better change this section code to async/await, i'm slowly changing all callback/promise to async/await.

PR #445, opened 2017-10-31
take 192.168.218.0/24 as an invalid IP

jimyang2008:
This is for https://github.com/firewalla/firewalla.ios/issues/170


scripts/check_fix_network.sh
```diff
@@ -28,7 +28,7 @@ get_value() {
     kind=$1
     case $kind in
         ip)
-            /sbin/ip addr show dev eth0 | awk '$NF=="eth0" {print $2}' | fgrep -v 169.254. | fgrep -v 192.168.218.
+            /sbin/ip addr show dev eth0 | awk '$NF=="eth0" {print $2}' | fgrep -v 169.254. | fgrep -v 192.168.218.1
```
MelvinTo: may need exact match here??

this fgrep will match 192.168.218.11??

PR #545, opened 2017-11-27
 invoke main-run at the end of soft upgrade but without restarting firekick

jimyang2008:


scripts/fireupgrade_soft.sh
```diff
@@ -19,15 +19,21 @@
...
 
 # call upgrade
-/home/pi/firewalla/scripts/firelog -t cloud -m  "INFO: Upgrade completed with services restart in soft mode $commit_before $commit_after"
+${FIREWALLA_HOME}/scripts/firelog -t cloud -m  "INFO: Upgrade completed with services restart in soft mode $commit_before $commit_after"
 touch /tmp/FWUPGRADING
 touch /home/pi/.firewalla/managed_reboot
 sleep 5
 for svc in api main mon
   do
       sudo systemctl restart fire${svc}
```
MelvinTo: these systemctl restart will be covered by main-run, no need to execute here.

PR #573, opened 2017-12-08
issue500: adding health check script and schedule in cron

jimyang2008:


jerrchen:
you sure this will work?  When the MMC is dead ... it is dead the file system is dead.  I don't think we can even run dmesg

MelvinTo:
not just for firemain.. need a general script to update repository before starting fire{main,mon,kick,api}

Issue #614, opened 2017-12-22
Device error when fetching traffic details


starsy:
```
ERROR 2017-12-22 03:39:47 Netbot: Got error before simpleTxData: TypeError: Cannot read property 'country' of null
    at async (/home/pi/firewalla/net2/FlowTool.js:440:33)
    at tryBlock (/home/pi/.node_modules/asyncawait/src/async/fiberManager.js:39:33)
    at runInFiber (/home/pi/.node_modules/asyncawait/src/async/fiberManager.js:26:9) TypeError: Cannot read property 'country' of null
    at async (/home/pi/firewalla/net2/FlowTool.js:440:33)
    at tryBlock (/home/pi/.node_modules/asyncawait/src/async/fiberManager.js:39:33)
    at runInFiber (/home/pi/.node_modules/asyncawait/src/async/fiberManager.js:26:9)
```

starsy:
closed w/ https://github.com/firewalla/firewalla/pull/616 & https://github.com/firewalla/firewalla/pull/615

PR #615, opened 2017-12-22
Fix null object reference when displaying traffic

starsy:
Issue: https://github.com/firewalla/firewalla/issues/614


jerrchen:
seems already been fixed in master

MelvinTo:
fixed already in my code, so i'll just accept

PR #625, opened 2017-12-26
Natpmp support 

jerrchen:
Natpmp should only be called (via upnp.js module) in main.  This little thing will open a listening port,  which can only be one in the system.


MelvinTo:
reviewed

PR #632, opened 2017-12-27
Enabled HW watchdog / Code to send configs to cloud

jerrchen:
HW watchdog is needed for situations if there is anything that may cause the kernel to hang.  This is experimental until we see it is stable, should go into production.

Checkin data now will contain user configuration and alarms/... see commit comments on what's is clear text.  This data can eventually be used to recover a firewalla, as well as aid in debugging problems.

### DIFF
--- a/net2/HostManager.js
+++ b/net2/HostManager.js
@@ -1851,6 +1851,39 @@ module.exports = class {
     return Promise.all(ipList.map((ip) => flowManager.migrateFromOldTableForHost(ip)));
   }
 
+    /* 
+     * data here may be used to recover Firewalla configuration 
+     */
+    getCheckInAsync() {
+        return async(() => {
+          let json = {};
+          let requiredPromises = [
+            this.policyDataForInit(json),
+            this.modeForInit(json),
+            this.policyRulesForInit(json),
+            this.exceptionRulesForInit(json),
+            this.natDataForInit(json),
+            this.ignoredIPDataForInit(json),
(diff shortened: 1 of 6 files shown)


scripts/fire-reboot-normal
```diff
@@ -2,12 +2,14 @@
 logger "FIREWALLA REBOOT NORMAL"
 #sudo sh -c 'echo V > /dev/watchdog'
 #sudo sh -c 'echo V > /dev/watchdog'
-sudo pkill bitbridge
-sudo pkill fire-ping
-sudo pkill sleep
+sudo service firemain stop
```
MelvinTo: what about the other services? fireapi, firemon, firekick?

net2/HostManager.js
```diff
@@ -1851,6 +1851,39 @@ module.exports = class {
     return Promise.all(ipList.map((ip) => flowManager.migrateFromOldTableForHost(ip)));
   }
 
+    /* 
+     * data here may be used to recover Firewalla configuration 
+     */
+    getCheckInAsync() {
```
MelvinTo: all these data uploaded to cloud?

scripts/fire-reboot-normal
jerrchen: fireapi can't be stopped, otherwise rest of the process won't execute, and never reboot.  Rest services don't need stop.  We stopping fireman is to recover network

PR #635, opened 2017-12-30
fixing previously race condition of cloud checkin

jerrchen:


sensor/BoneSensor.js
```diff
@@ -67,16 +65,23 @@ class BoneSensor extends Sensor {
       let sysInfo = await (sysManager.getSysInfoAsync());
 
       log.info("Checking in Cloud...",sysInfo,{});
-/*
+ 
+      // First checkin usually have no meaningful data ...
```
MelvinTo: ugly code..

PR #637, opened 2017-12-31
Fixed dns policy growth problem

jerrchen:
- added rsyslog file, we may need to override the default.


etc/50-default.conf
```diff
@@ -0,0 +1,70 @@
+#  Default rules for rsyslog.
```
MelvinTo: what this file is about?

jerrchen: to remove all the unwanted syslog.  today we pick and edit lines.  Copying the file over is more efficient. I have added removal of the console stuff

PR #653, opened 2018-01-07
Creating /script directory to store critical scripts outside of git

jerrchen:


scripts/main-start
```diff
@@ -268,6 +269,25 @@ fi
 
 sudo setcap cap_net_raw+eip $(eval readlink -f `which nodejs`)
 
+# Take care of upgrade differently
+
+
+mkdir $PI_HOME/scripts
+
+[ -s /home/pi/firewalla/scripts/fireupgrade.sh ] && cp /home/pi/firewalla/scripts/fireupgrade.sh $PI_HOME/scripts/.
```
MelvinTo: need add md5sum check to make sure file is correct. these scripts are absolutely critical

jerrchen: let me note it down.  Need some help, my bash is limited.

PR #657, opened 2018-01-09
 adjust video alarm's threshold

starsy:


monitor/FlowMonitor.js
```diff
@@ -183,7 +183,7 @@ module.exports = class FlowMonitor {
                if (ignore == false) {
                 log.info("######## flowIntel Processing",JSON.stringify(flow));
                 if (this.checkIntelClass(flow['intel'],"av")) {
-                    if ( (flow.du && Number(flow.du)>60) && (flow.rb && Number(flow.rb)>5000000) ) {
+                    if ( (flow.du && Number(flow.du)>5) && (flow.rb && Number(flow.rb)>1000000) ) {
```
jerrchen: This will make the video detection very sensitive.   We don't want anything to trigger a video.  I think the old number was targeted at people watching video >5 min or so.

MelvinTo: how did you adjust the old number to meet > 5min criteria?

jerrchen: A lot of manual trials.    Today's numbers already pretty aggressive.  We need to use our cloud algorithms for detection and remove this hardcode in the future.

starsy: My experience is never saw a video alarm triggered at my place, by adjusting this parameter at least the video alarms can be triggered. i think the current balance is too much towards precision but the recall is too low, which renders the video alarm to be not useful.

jerrchen: video alarm is different than porn or hacks ... :)

drahnieR:
we probably want a remote configuration solution on this?

PR #724, opened 2018-02-02
watchdog can be disabled by a flag

starsy:


scripts/main-run
```diff
@@ -55,16 +55,21 @@ $FIREWALLA_HOME/scripts/main-start
...
-sleep 1
-$FIREWALLA_HOME/scripts/fire-ping.sh
-sudo service watchdog stop
-echo "WatchDog Stopped"
-$FIREWALLA_HOME/scripts/fire-ping.sh
-sleep 3 
-sudo service watchdog start
-echo "WatchDog Started"
+if [ -e /home/pi/.firewalla/config/disable_watchdog ]; then
+    logger "FIREWALLA:WATCHDOG:DISABLED"
```
MelvinTo: should call "sudo systemctl stop watchdog" when the disable flag exists, in case it's already started.

starsy: added

jerrchen:
Ticket #388, we have one customer's router gateway ping is not stable.  Which causes fire-ping to fail.  Is it possible to add a flag somewhere, once that's enabled, fire-ping will not reboot.   The problem likely is only with pings

MelvinTo:
this is a separate feature, will merge this first.

PR #774, opened 2018-02-24
fix issue 500 - mgit recover may corrupt system when another mgit pro…

jimyang2008:
…cess is running

## Notice


scripts/mgit
```diff
@@ -5,41 +5,118 @@
...
+    case $rname in
+        firewalla)
+        case $item in
+            url)    echo 'https://github.com/firewalla/firewalla.git' ;;
+            branch) echo 'release_6_0' ;;
+        esac
+        ;;
+        .node_modules)
+        case $item in
+            url)    echo 'https://github.com/firewalla/firewalla.git' ;;
```
MelvinTo: this is not updated to the correct url

scripts/mgit
```diff
@@ -5,41 +5,118 @@
...
+    sync
+    echo $toplevel > $mgit_repo/toplevel
+    sync
+
+    # lay down the marker file
+    date > $marker_file
+    sync
+
+    # execute actual git command
+    git "$@" || exit 1
```
MelvinTo: we should continue deleting the mark file even the git command is failed.

the purpose for mgit recover is to recover if user unplugs power during git operations, not to recover if git fails due to like network issue or github down..

jerrchen:
Should the checksum also be changed?

PR #794, opened 2018-03-02
Consolidate policy for dnsmasq and firehttpd

starsy:


jerrchen:
how big is the HashSet stored in Redis?  possible not to.
I also see bone intel being called, how often is this call

 How much is this process adding memory? both in redis and system memory.
 How much interactions with bone cloud?

MelvinTo:
* this hashset will be the minimized hashset for porn only, it won't take much memory.
* hashset will be loaded from bone one time only per node run, or maybe periodically pull in the future
* Reason to put in redis is for FireWeb to dynamic query in the runtime, so that we can redirect traffic correctly (e.g. if porn, go to nicewebsite.com, if others, go to "xx is blocked by firewalla" page.)

MelvinTo:
the memory footprint of FireWeb will be reduced...

starsy:
local policy is to minimize the realtime interactions with bone, which both reduce bone usage and also reduce latency.

starsy:
current porn list is 1M, adblock list is 100K. very small.

starsy:
over-estimated the size:

pi@Firewalla:~/firewalla$ awk -F '/' '{print $2}' ~/.firewalla/config/dns/adblock_filter.conf | wc -c
109710
pi@Firewalla:~/firewalla$ awk -F '/' '{print $2}' ~/.firewalla/config/dns/family_filter.conf | wc -c
617580

in total 727,290 characters.

PR #820, opened 2018-03-10
In DHCP mode, unmonitored hosts should be allocated with original subnet ip

starsy:
https://github.com/firewalla/firewalla.ios/issues/531

originally in dhcp mode, there is a dnsmasq to allocate ip in 192.168.218.x subnet, now adding the other dnsmasq to handle original subnet ip allocation. 

2 dnsmasqs collaborate through filter and hosts file configuration.

no modification to original dnsmasq binary.


starsy:
as master is frozen for beta now.
new PR to supermaster: https://github.com/firewalla/firewalla/pull/822 
closing this PR here.

PR #1006, opened 2018-07-12
Add device back online alert for digital presence

jasonlyc:
# Notice

* If change mgit, fire-ping, fireupgrade, check_reset, fireupgrade2.service, check_fix_network.sh  ... PLEASE CHANGE bootstrap.sha256sum
  * As an option, you can run command **cd .git; ln -s ../.githooks hooks** to enable sha256sum auto validation during commit

firewalla/firewalla.ios#766


jerrchen:
how are you control which device to send alarms? some devices do go on / off ... wake up and sleep, and you don't want them notify users.  And some people don't care ... so this need to be a setting and only target at human used devices.

jasonlyc:
I do see there is a 'dtype' field in device information which I think reflects whether it is a human used device to some extent. Below are my thoughts:
Presences that are not triggered by human may also need to be taken care of sometimes. What if my robotic sweeper suddenly awakens when nobody is at home... I think it's hard for us to tell whether a presence of device is on purpose or unexpected. Even some devices go on/off routinely, we can still provide the visibility in case it is not awakened sometimes... However, the user can still disable digital presence monitoring on a specific device as long as he doesn't care about it.

jerrchen:
Do the reverse.  So user can select which device to monitor presence.   The only device that user may need presence are phone'/tablets, they are much less in number.  Plus  ... the use for this feature very likely be for kids device only.  

1. add a button in UI, default off
2. if user need to know if the device moves, they can turn on. 

future we can optimize.

PR #1035, opened 2018-07-20
Change watchdog interval to 120 seconds

jasonlyc:
# Notice

* If change mgit, fire-ping, fireupgrade, check_reset, fireupgrade2.service, check_fix_network.sh  ... PLEASE CHANGE bootstrap.sha256sum
  * As an option, you can run command **cd .git; ln -s ../.githooks hooks** to enable sha256sum auto validation during commit

firewalla/firewalla.ios#795

### DIFF
--- a/etc/watchdog.conf
+++ b/etc/watchdog.conf
@@ -3,7 +3,7 @@
 #ping			= 172.26.1.255
 #interface		= eth0
 file			= /tmp/watchdog
-change			= 60
+change			= 120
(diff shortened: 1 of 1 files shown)


jasonlyc:
After fire-ping.sh was added to cron, cronjob was stopped somehow. If it happens around 58 or 59 second in a minute, the next cronjob will be scheduled to run after more than 1 minute, causing the starvation of watchdog.

PR #1042, opened 2018-07-23
Add uptime record at the end of main-start

jasonlyc:
firewalla/firewalla.ios#801


sys/invitation.js
```diff
@@ -249,6 +251,19 @@ class FWInvitation {
       intercomm.bpublish(gid, obj.r, config.serviceType);
     }
 
+    const cmd = "awk '{print $1}' /proc/uptime";
```
MelvinTo: You can use async version of exec.

https://www.npmjs.com/package/child-process-promise

PR #1060, opened 2018-08-06
Expire stats:last24 automatically

jasonlyc:
firewalla/firewalla.ios#820


MelvinTo:
this may not be enough, need the cleanup sensor to cleanup existing stats keys.

jasonlyc:
stats:hour:* is already taken care by OldDataCleanSensor, and data in stats:last24:* is a hash whose key represents the hour of day, thereby it should at most contain 24 key-value pairs in each entry.

PR #1115, opened 2018-09-11
Make VPN server configurable and remove many hard codes

jasonlyc:


jasonlyc:
firewalla/firewalla.ios#852

PR #1148, opened 2018-09-30
Generate dummy crl file if not present

jasonlyc:
firewalla/firewalla.ios#907


jasonlyc:
Openvpn seems to report error for empty crl file

PR #1156, opened 2018-10-02
fire-ping try backup domains if default gateway is not responding

jasonlyc:
firewalla/firewalla.ios#906

### DIFF
--- a/scripts/fire-ping.sh
+++ b/scripts/fire-ping.sh
@@ -37,10 +37,19 @@ for i in `seq 1 10`; do
 #      /home/pi/firewalla/scripts/firelog -t debug -m"FIREWALLA PING WRITE"
        exit 0
     else
-      echo "Ping Failed"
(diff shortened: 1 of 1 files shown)


scripts/fire-ping.sh
```diff
@@ -37,10 +37,19 @@ for i in `seq 1 10`; do
 #      /home/pi/firewalla/scripts/firelog -t debug -m"FIREWALLA PING WRITE"
        exit 0
     else
-      echo "Ping Failed"
-      /home/pi/firewalla/scripts/firelog -t debug -m "FIREWALLA PING NO Local Network $DEFAULT_ROUTE"
-      sleep 1
-      touch /tmp/watchdog 
+      BACKUP_DOMAIN="firewalla.com"
+      if [ $((i % 2)) -eq 0 ]; then
+        BACKUP_DOMAIN="github.com"
```
jasonlyc: firewalla.encipher.io seems to filter ICMP, therefore github.com is used instead.

PR #1161, opened 2018-10-09
Disable compression in ovpn by default

jasonlyc:
firewalla/firewalla.ios#927


jerrchen:
Make sure this change is compatible with old profile the customer may already using.  Generating an error is also good.

vpn/server_config.txt
```diff
@@ -26,6 +26,7 @@ duplicate-cn
 keepalive 10 120
 tls-auth /etc/openvpn/easy-rsa/keys/ta.key 0
 crl-verify /etc/openvpn/crl.pem
+client-config-dir /etc/openvpn/client_conf
```
jasonlyc: Use this option to enable profile-wise client configs, e.g., compression options...

jasonlyc:
This should be compatible with old profile. Since client-conf-dir can enable runtime profile-wise configuration, compression will actually also be disabled for clients with old profiles.

vpn/client_conf.txt
```diff
@@ -0,0 +1,4 @@
+COMP_LZO_OPT
+push "COMP_LZO_OPT"
+#COMPRESS_OPT
```
jasonlyc: comment this option for backward compatibility

PR #1171, opened 2018-10-11
Skip /8 subnet in NmapSensor

jasonlyc:
firewalla/firewalla.ios#926

### DIFF
--- a/sensor/NmapSensor.js
+++ b/sensor/NmapSensor.js
@@ -213,6 +213,10 @@ class NmapSensor extends Sensor {
     return Promise.all(this.networkRanges.map((range) => {
 
       log.info("Scanning network", range, "to detect new devices...");
+      if (range.endsWith('/8')) {
+        log.info("Subnet " + range + " contains too many ip addresses to scan, skip it.")
+        return;
+      }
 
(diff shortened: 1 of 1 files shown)


sensor/NmapSensor.js
```diff
@@ -213,6 +213,10 @@ class NmapSensor extends Sensor {
     return Promise.all(this.networkRanges.map((range) => {
 
       log.info("Scanning network", range, "to detect new devices...");
+      if (range.endsWith('/8')) {
+        log.info("Subnet " + range + " contains too many ip addresses to scan, skip it.")
+        return;
```
MelvinTo: should return a promise? otherwsie Promise.all will never return...

or better way is use async/await.

sensor/NmapSensor.js
```diff
@@ -213,6 +213,10 @@ class NmapSensor extends Sensor {
     return Promise.all(this.networkRanges.map((range) => {
 
       log.info("Scanning network", range, "to detect new devices...");
+      if (range.endsWith('/8')) {
```
MelvinTo: shall we force to scan /24 when /8 is specified?

jerrchen: Make sure there is a visual alarm, and then force /24, for the blue in the future we may expend more.

PR #1200, opened 2018-10-26
Sync /etc/hosts at the beginning of main-run and fireupgrade.sh

jasonlyc:
firewalla/firewalla.ios#957


jasonlyc:
> why it need to be done in fire upgrade? try not to touch that file unless absolutely needed. that file should have only what's necessary to upgrade.

Inconsistency between hostname and /etc/hosts may cause significant extra time cost for sudo commands, and there are also a number of sudo commands during the execution of fireupgrade.sh, which may cause unpredictable results

PR #1227, opened 2018-11-05
1.958 Beta Drop

MelvinTo:


vpn/install2.sh
```diff
@@ -1,24 +1,20 @@
 #!/bin/bash
```
jerrchen: There is an old bug where install2.sh was called twice, the original code here was to ensure things are in good order.  The file size check was to do that.  When the bug occurs, the server.conf was set to 0

jasonlyc: File size check is moved to confgen.sh, which should do the same thing. install2.sh will only generate key chains for server.

PR #1229, opened 2018-11-06
Turn off led light if NO_FIREKICK_RESTART is set

jasonlyc:


jasonlyc:
This mainly happens in case of soft upgrade

platform/blue/platform.sh
```diff
@@ -8,10 +8,17 @@ MAX_NUM_OF_PROCESSES=4000
 MAX_NUM_OF_THREADS=20000
 
 function heartbeatLED {
-  sudo sh -c 'echo none > /sys/devices/platform/leds/leds/nanopi:green:status/trigger'
+  sudo sh -c 'echo heartbeat > /sys/devices/platform/leds/leds/nanopi:green:status/trigger'
```
MelvinTo: I intentionally not use the green light as it can barely be seen...

PR #1240, opened 2018-11-07
Set max-mount-count to 1 for fsck, use flock for vpn related scripts

jasonlyc:


jasonlyc:
Also need to enable file system check for mmcblk0p2/3 in /etc/fstab

PR #1251, opened 2018-11-14
Change all ip-based device identity to mac-based

jasonlyc:
firewalla/firewalla.ios#1005

### DIFF
--- a/net2/FlowManager.js
+++ b/net2/FlowManager.js
@@ -760,7 +760,7 @@ module.exports = class FlowManager {
         });
     }
 
-    summarizeActivityFromConnections(flows,callback) {
+    async summarizeActivityFromConnections(flows) {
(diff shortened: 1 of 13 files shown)


net2/FlowManager.js
```diff
@@ -842,18 +838,15 @@ module.exports = class FlowManager {
...
         }
 
         //log.info("### Cleaning",flowobj);
 
-        bone.flowgraph("clean", [flowobj],(err,data)=>{
-            if (callback) {
-                callback(err,data);
-            }
+        return new Promise((resolve, reject) => {
+          bone.flowgraph("clean", [flowobj],(err,data)=>{
```
MelvinTo: there is already an async function called bone.flowgraphAsync :)

PR #1266, opened 2018-11-26
Enable debug log level on startup

drahnieR:
`echo debug> ~/.firewalla/config/FWDEBUG` to enable debug logging on system startup

### DIFF
--- a/net2/logger.js
+++ b/net2/logger.js
@@ -18,7 +18,8 @@ const config = winston.config;
 
 const loggerManager = require('./LoggerManager.js');
 
-let path = require('path');
+const path = require('path');
+const fs = require('fs');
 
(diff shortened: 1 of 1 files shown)


net2/logger.js
```diff
@@ -163,8 +171,10 @@ module.exports = function (component) {
     }
   }
 
+  // winston 1.1.2 use different ordering on log levels
```
MelvinTo: we use 2.4.0 in .node_modules
https://github.com/firewalla/fnm.node8.aarch64/blob/master/winston/package.json

default order:
https://github.com/firewalla/fnm.node8.aarch64/blob/master/winston/lib/winston/config/npm-config.js

PR #1294, opened 2018-12-18
Randomly switch to dev cloud instance for beta branch

jasonlyc:


MelvinTo:
also need to create a isAlpha, beta_7_0 is the alpha.

PR #1295, opened 2018-12-19
upnp check task will only run in FireMain

jasonlyc:
firewalla/firewalla.ios#1112

### DIFF
--- a/extension/upnp/upnp.js
+++ b/extension/upnp/upnp.js
@@ -60,38 +60,39 @@ module.exports = class {
             instance = this;
             this.refreshTimers = {};
             
-            upnpIntervalHandler = setInterval(
-                () => {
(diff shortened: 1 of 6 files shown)


extension/upnp/upnp.js
```diff
@@ -60,38 +60,39 @@ module.exports = class {
...
-                                this.addPortMappingUPNP(protocol, localPort, externalPort, description)
-                            } else {
-                                log.info("Mapping still exists")
-                            }
-                        })
-                    })
-                },
-                upnpCheckInterval
-            )
+            if (process.title === "FireMain") {
```
MelvinTo: use this please :)

https://github.com/firewalla/firewalla/blob/master/net2/Firewalla.js#L279

PR #1297, opened 2018-12-20
Policy and PolicyManager refactor

drahnieR:
This PR aim to build a more Redis friendly Policy implementation.
All policies should now be considered as Policy objects within the box.

### DIFF
--- a/alarm/PolicyManager2.js
+++ b/alarm/PolicyManager2.js
@@ -113,20 +113,22 @@ class PolicyManager2 {
     })
 
     this.queue.process((job, done) => {
-      const event = job.data
-      const policy = new Policy(event.policy)
-      const oldPolicy = new Policy(event.oldPolicy)
+      const event = job.data;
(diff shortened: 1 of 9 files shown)


alarm/PolicyManager2.js
```diff
@@ -115,33 +113,35 @@ class PolicyManager2 {
     })
 
     this.queue.process((job, done) => {
-      const event = job.data
-      const policy = this.jsonToPolicy(event.policy)
-      const oldPolicy = this.jsonToPolicy(event.oldPolicy)
+      const event = job.data;
+      const policy = event.policy instanceof Policy ? event.policy : new Policy(event.policy);
```
MelvinTo: event.policy will never be a class, because it's passed from redis pub/sub

drahnieR: not all, after this change...
https://github.com/firewalla/firewalla/blob/b5b7fdf26d4ca67e5d1163937e7dca4587e1a1a1/alarm/PolicyManager2.js#L712-L726

PR #1300, opened 2018-12-21
Implements alarm:block & alarm:allow for UPNPAlarm

drahnieR:
firewalla/firewalla.ios#1054
Adds some device recognization as well
upnp issued by firewalla should no longer generate alarms


drahnieR:
Add universal check between devicePort policy and alarm. it's no longer bound with UPNPAlarm

drahnieR:
resolves conflict

PR #1306, opened 2018-12-22
support premium and delete eid from gid

MelvinTo:


jasonlyc:
firewalla/firewalla.ios#1130

PR #1310, opened 2018-12-26
Add firewall mode which blocks incoming connections

jasonlyc:
firewalla/firewalla.ios#968


extension/portforward/portforward.js
```diff
@@ -159,6 +162,7 @@ class PortForward {
       }
       
       log.info("PORTMAP: Add",map);
+      await (shieldManager.addIncomingRule(map.protocol, map.toIP, map.dport));
```
MelvinTo: no need to add bracket for native await

extension/portforward/portforward.js
```diff
@@ -178,6 +182,7 @@ class PortForward {
       this.config.maps.splice(old, 1);
 
       log.info("PortForwarder:removePort Found MAP", dupMap);
+      await (shieldManager.removeIncomingRule(map.protocol, map.toIP, map.dport));
```
MelvinTo: no need to add bracket for native await

extension/portforward/portforward.js
jasonlyc: Seems native await is not allowed in 'async' block from 'async/await' library

PR #1316, opened 2019-01-02
Check whether o is null in HostManager.getHosts, fix typo in ShieldManager and listen on 127.0.0.1 in local.bro

jasonlyc:


jasonlyc:
firewalla/firewalla.ios#1158

PR #1331, opened 2019-01-11
when fetching cloud intel fails, still save local intel result. 

drahnieR:
switch to native async/await in multiple places

### DIFF
--- a/hook/DestIPFoundHook.js
+++ b/hook/DestIPFoundHook.js
@@ -114,6 +114,10 @@ class DestIPFoundHook extends Hook {
     // app
     cloudIntelInfos.forEach((info) => {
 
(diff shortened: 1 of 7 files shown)


hook/DestIPFoundHook.js
```diff
@@ -357,30 +362,24 @@ class DestIPFoundHook extends Hook {
         return; // reserved black hole and blue hole...
       }
       
-      this.appendNewFlow(ip,fd);
+      this.appendNewFlow(ip, fd);
     });
 
-    sem.on('DestIP', (event) => {
```
MelvinTo: don't delete this, this is a debugging interface to manually trigger destip action by redis publish

PR #1335, opened 2019-01-14
Some rework on Policy

drahnieR:
- empty `scope`, `expire`, `cronTime` won't be presented in Policy object
- remove redundant code on policy enforcement
- rewrites some async/await
- fix small bug on test API alarm/create

### DIFF
--- a/control/Block.js
+++ b/control/Block.js
@@ -301,12 +301,12 @@ function blockImmediate(destination, ipset) {
 async function advancedBlock(tag, macAddresses, destinations) {
   await setupBlockingEnv(tag);
 
-  await Promise.all(macAddresses.map(mac => {
-    return advancedBlockMAC(mac, getMacSet(tag))
-  }))
-  await Promise.all(destinations.map(addr => {
-    return block(addr, getDstSet(tag))
-  }))
+  for (const mac of macAddresses) {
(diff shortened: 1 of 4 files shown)


drahnieR:
this should go to #1332 , it will cause policy comparison problem in some situation

control/Block.js
```diff
@@ -298,55 +298,51 @@ function blockImmediate(destination, ipset) {
...
-      await (advancedBlockMAC(mac, getMacSet(tag)))
-    })
-    destinations.forEach((addr) => {
-      await (block(addr, getDstSet(tag)))
-    })
-  })()
+async function advancedBlock(tag, macAddresses, destinations) {
+  await setupBlockingEnv(tag);
+
+  await Promise.all(macAddresses.map(mac => {
```
MelvinTo: need to make all blocks sequential.

```
for(const macAddress of macAddresses) {
    await advancedBlockMac(...
}
```

control/Block.js
```diff
@@ -298,55 +298,51 @@ function blockImmediate(destination, ipset) {
...
+    return advancedBlockMAC(mac, getMacSet(tag))
+  }))
+  await Promise.all(destinations.map(addr => {
```
MelvinTo: same here

PR #1339, opened 2019-01-16
Ignore small packets, fix missing top upload/download for new devices, winston debug log level support and local dns fallback in dnsmasq

jasonlyc:


net2/logger.js
```diff
@@ -53,7 +53,7 @@ else if (production) {
 }
 
 function getFileTransport() {
-  let loglevel = globalLogLevel;
+  let loglevel = 'debug';
```
jasonlyc: We use wrapper function of logger to determine log level. loglevel here is used in winston logger itself.

PR #1347, opened 2019-01-22
add cpu temperature monitoring for blue, also temperature is added to…

drahnieR:
… diag.sayHello()

default configuration:
monitor interval: 1min,
report interval 2hrs
report threshold 95 celsius


drahnieR:
firewalla/firewalla.ios#1172

PR #1352, opened 2019-01-23
transactional flowstash clean up

drahnieR:


MelvinTo:
what's the reason to make it transactional?

drahnieR:
To prevent lose of flow data here, as we used to add after remove and that could result in error.
And the previous logic calls expireat multiple times, it not necessary.

PR #1357, opened 2019-01-27
Support whitelist mode and implement all enforcement in nat/filter tables

jasonlyc:


jasonlyc:
This pull request is submitted when whitelist is enabled on W6 office network 😸

PR #1360, opened 2019-01-30
Refine time/duration calculation for rate/trending, fix missing nat rule if dnsmasq is reloaded, verify host in DestIPFoundHook

jasonlyc:


net2/FlowAggrTool.js
```diff
@@ -231,7 +231,7 @@ class FlowAggrTool {
     let end = options.end;
 
     // if working properly, sumflow should be refreshed in every 10 minutes
-    let expire = options.expireTime || 24 * 60; // by default expire in 24 minutes
+    let expire = options.expireTime || 24 * 60 * 60; // by default expire in 24 hours
```
MelvinTo: no need to keep sumflow that long, 24 mins should be enough, or the memory usage will goes high.

sensor/FlowAggregationSensor.js
```diff
@@ -609,6 +614,7 @@ class FlowAggregationSensor extends Sensor {
     })()
   }
   
+  // TODO: Why call it cleanup? This looks confusing. It actually summarize different category flows, i.e., categoryflow:(mac:)?
```
MelvinTo: dont remember...

hook/DestIPFoundHook.js
```diff
@@ -251,25 +265,25 @@ class DestIPFoundHook extends Hook {
     let forceRedisUpdate = options.forceUpdate;
 
     return (async() => {
+      let sslInfo = await intelTool.getSSLCertificate(ip);
+      let dnsInfo = await intelTool.getDNS(ip);
```
MelvinTo: sslInfo and dnsInfo are not guaranteed to be the exact domain this device is really accessing. 

this could potentially reduce the number of entry to be added to category domain

jasonlyc: Current implementation uses these two info to determine domain... I just move this piece of code ahead of cache validation...

hook/DestIPFoundHook.js
```diff
@@ -251,25 +265,25 @@ class DestIPFoundHook extends Hook {
...
         }
       }
 
-      log.info("Found new IP " + ip + " fd " +fd+ " flow "+flow+", checking intels...");
-
-      let sslInfo = await intelTool.getSSLCertificate(ip);
-      let dnsInfo = await intelTool.getDNS(ip);
+      log.info("Found new IP " + ip + " fd " +fd+ " flow "+flow+ " domain " + domains + ", checking intels...");
 
-      let domains = this.getDomains(sslInfo, dnsInfo);
```
jasonlyc: It is moved from here

jasonlyc:
firewalla/firewalla.ios#1214

PR #1368, opened 2019-02-11
safe search

MelvinTo:


sensor/SafeSearchPlugin.js
```diff
@@ -0,0 +1,222 @@
+/*    Copyright 2016 Firewalla LLC
+ *
```
jerrchen: change 2016 to 2019

sensor/SafeSearchPlugin.js
```diff
@@ -0,0 +1,222 @@
...
+  bing: {
+    "strict.bing.com": [
+      "www.bing.com"
+    ]
+  }
+}
+
+class SafeSearchPlugin extends Sensor {
+
+  scheduledJob() {
```
jasonlyc: This is not used... And ipv6 is not declared before...

PR #1380, opened 2019-02-27
add alarm cleanup in OldDataCleanSensor

drahnieR:
firewalla/firewalla.ios#1168

### DIFF
--- a/sensor/OldDataCleanSensor.js
+++ b/sensor/OldDataCleanSensor.js
@@ -275,9 +275,9 @@ class OldDataCleanSensor extends Sensor {
     const invalidMACs = macs.filter((m) => {
       return m.match(/[a-f]+/) != null
     })
-    return Promise.all(invalidMACs.map((m) => {
-      hostTool.deleteMac(m);
-    }))
+    return Promise.all(
+      invalidMACs.map(m => hostTool.deleteMac(m))
+    )
   }
 
   async cleanupAlarmExtendedKeys() {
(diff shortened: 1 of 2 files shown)


sensor/OldDataCleanSensor.js
```diff
@@ -373,38 +374,37 @@ class OldDataCleanSensor extends Sensor {
...
-              if(result) {
-                await (rclient.hsetAsync(key, "blockin", false))
-                log.info("Migrated successfully")
-              } else {
-                log.error("Failed to migrate")
-              }
+  async hostPolicyMigration() {
+    try {
+      const keys = await rclient.keysAsync("policy:mac:*");
+      for (let i in keys) {
```
MelvinTo: use "of"
   for( let key of keys )

sensor/OldDataCleanSensor.js
```diff
@@ -197,104 +194,90 @@ class OldDataCleanSensor extends Sensor {
...
-        await (hostTool.deleteMac(m))
-      })
-    })()
+  async cleanInvalidMACAddress() {
+    const macs = await hostTool.getAllMACs();
+    const invalidMACs = macs.filter((m) => {
+      return m.match(/[a-f]+/) != null
+    })
+    return Promise.all(invalidMACs.map((m) => {
+      hostTool.deleteMac(m);
```
MelvinTo: should use async keyword and return async function.

async (m) => hostTool.deleteMac(m)

PR #1384, opened 2019-03-04
Merge from alpha, beta, production

MelvinTo:


MelvinTo:
to prepare for next commit

PR #1399, opened 2019-03-15
rework on BroNoticeAlarm & notification locales

drahnieR:


locales/en.json
```diff
@@ -7,100 +7,85 @@
   "NOTIF_REBOOT_COMPLETE": "Firewalla reboot completed",
```
drahnieR: Remove ALARM_* entries as their are obsoleted.
Use INFO_ALARM_* for alarm.message
Reorder entries by alarm type for better readability

PR #1406, opened 2019-03-18
prevent enriching local ip from cloud

drahnieR:


net2/SysManager.js
```diff
@@ -767,8 +767,9 @@ module.exports = class {
       }
       return this.inMySubnet6(ip);
     } else {
-      log.debug("SysManager:ERROR:isLocalIP", ip);
-      return true;
+      log.error("SysManager:ERROR:isLocalIP", ip);
+      // TODO: we should throw error here
+      return false;
```
MelvinTo: don't know if any magic could happen here by changing the return value...

PR #1407, opened 2019-03-18
fix upnp timeout error

drahnieR:


vpn/view.js
```diff
@@ -1,36 +0,0 @@
-'use strict';
```
jerrchen: don't remove this, it is extremely useful to query UPNP using this tool.

drahnieR: reverted

PR #1415, opened 2019-03-27
Add optional SNAT for spoof

jasonlyc:


jasonlyc:
monitored_ip_set is always the same as monitored_hosts in redis. In dhcp mode they will be cleared. In node mode, spoof will be turned off, leaving the SNAT rules unused.

PR #1440, opened 2019-04-17
Support multiple bitbridge instances for v4/v6

jasonlyc:


bin/bitbridge4.sh
```diff
@@ -20,9 +17,13 @@ fi
 
 sudo setcap cap_net_admin,cap_net_raw=eip $FIREWALLA_BIN/$BINARY
 
-if [[ ! -z "$BINARY_ARGUMENTS" ]]; then
-  $FIREWALLA_BIN/$BINARY $BINARY_ARGUMENTS
-else
-  exit 1
-fi
+for RC_FILE in $FIREWALLA_BIN/$BINARY.*.rc; do
+  if [[ -e $RC_FILE ]]; then
```
MelvinTo: a minor enhancement here. use bracket () to keep "source" result only for the BINARY command

Example:
```
#file.rc
    PPP=1
$ (source file.rc; echo $PPP) => echo 1
$ echo $PPP #output nothing
```

PR #1442, opened 2019-04-17
add missing DeviceUpdate event source

drahnieR:


MelvinTo:
what bug is this pr trying to fix?

drahnieR:
when device info is updated without "from", incorrect "lastFrom" value in redis could be very confusing

PR #1448, opened 2019-04-23
use isPrimaryInterfaceIP to determine if ip should be spoofed in Spoofer.js

jasonlyc:
…fer.js


jasonlyc:
firewalla/firewalla.ios#1420

PR #1505, opened 2019-05-20
more foreach-async fixes

drahnieR:


net2/SecondaryInterface.js
```diff
@@ -112,7 +112,7 @@ exports.create = function(config, callback) {
             // other interface already occupies ip1, use alternative ip
             _secondaryIpSubnet = config.secondaryInterface.ip2;
             _secondaryIp = _secondaryIpSubnet.split('/')[0];
-            _secondaryMask = ip.cidrSubnet(_secondaryIp).subnetMask;
+            _secondaryMask = ip.cidrSubnet(_secondaryIpSubnet).subnetMask;
```
MelvinTo: is this a critical bug?

drahnieR: this happens when 192.168.218.1/24 network exists before walla. and it disables DHCP mode.

3 hits from 1 box on a single day so far, but it's 1.963. so there could be more coming.

PR #1544, opened 2019-06-11
Refine SS - step 1

MelvinTo:
* use overture instead of chinadns + dns_forwarder
* use cloud country intel instead of hard code one
* remove logic on haproxy
* make ss service systemctlized
* decouple logic on running ss from logic on enabling ss on any device. (make it ready for per device control eventually)


extension/ss_client/ss_client.js
```diff
@@ -33,371 +31,156 @@ const jsonfileWrite = Promise.promisify(jsonfile.writeFile)
...
-  async _stopSSClient() {
-    const cmd = `pkill -f 'fw_ss_client.*${this.getClientPidPath()}'`;
-    log.info("Stopping ss client...", cmd);
-    return exec(cmd).catch((err) => {
-//      log.info("Failed to stop ss client", err);
-    });
+  // prepare the chnroute files
+  async prepareCHNRoute() {
+    log.info("Preparing CHNRoute...");
+    (async () => {
```
drahnieR: redundant async wrap here

PR #1557, opened 2019-06-21
fix zero ip6 regex

drahnieR:


net2/HostTool.js
```diff
@@ -277,7 +277,7 @@ class HostTool {
   async getAllIPs() {
     let allIPs = [];
 
-    let macs = await this.getAllMACs();
+    let macKeys = await this.getAllMACs();
```
MelvinTo: how can this work before...

PR #1607, opened 2019-07-10
BUG: should clean up devicemasq folder when booting up #1612

wzjjack:
clean up leftover config

### DIFF
--- a/extension/dnsmasq/dnsmasq.js
+++ b/extension/dnsmasq/dnsmasq.js
@@ -1285,7 +1285,7 @@ module.exports = class DNSMASQ {
 
   async cleanUpLeftoverConfig(){
     try{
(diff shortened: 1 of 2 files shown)


extension/dnsmasq/dnsmasq.js
```diff
@@ -1282,4 +1282,22 @@ module.exports = class DNSMASQ {
       }
     }
   }
+
+  async cleanUpLeftoverConfig(){
+    try{
+      const userConfigFolder = f.getUserConfigFolder(),
+            dnsConfigFolder = `${userConfigFolder}/dns`,
```
jasonlyc: Clean up whole 'dns' folder may affect other functions based on dnsmasq, e.g., ad-block, policy filter. It's better to cleanup only safe-search specific files in dns folder.

PR #1608, opened 2019-07-11
per device family protect

wzjjack:


extension/dnsmasq/dnsmasq.conf
```diff
@@ -40,8 +40,8 @@ cache-size=1000
 
 # For debugging purposes, log each DNS query as it passes through
 # dnsmasq.
-log-facility=/dev/null
-# log-queries
+log-facility=/home/pi/.forever/dns.log
```
MelvinTo: do not turn on this, it may flood the filesystem

net2/PolicyManager.js
```diff
@@ -255,30 +258,25 @@ module.exports = class {
...
         break;
       case 'v1':
       default:
-        this.familyV1(ip, state, callback);
+        this.familyV1(host, ip, state, callback);
     }
   }
 
-  familyV1(ip, state, callback) {
+  async familyV1(host, ip, state, callback) {
```
MelvinTo: async is not required here, as no awaits inside or calling this function

MelvinTo:
a global feature flag is still required

sensor/FamilyProtectPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+        this.familyDnsAddr((err, dnsaddrs) => {
+            if (this.systemSwitch && this.adminSystemSwitch) {
+                return this.systemStart(dnsaddrs);
+            } else {
+                return this.systemStop(dnsaddrs);
+            }
+        });
+    }
+
+    async applyDeviceFamilyProtect(macAddress) {
```
jasonlyc: applyDeviceFamilyProtect should also use 'adminSystemSwitch' to determine if changes should be applied. If feature is off, neither system nor device family protect should be applied.

sensor/FamilyProtectPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+                } else {
+                    log.warn(`Dnsmasq: Error when remove ${configFile}`, err);
+                }
+            }
+        })
+        dnsmasq.restartDnsmasq();
+    }
+
+    // global on/off
+    async globalOn() {
```
jasonlyc: global on/off switch should invoke applyFamilyProtect to take case of both system level and device level

MelvinTo: device level policy will be triggered by PolicyManager.js

sensor/FamilyProtectPlugin.js
MelvinTo: this could also be a defect of SafeSearch feature?

they have same code structure

sensor/FamilyProtectPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+        this.adminSystemSwitch = true;
+        await this.applySystemFamilyProtect();
```
jasonlyc: I think this.applyFamilyProtect should be called here which will apply both system and device level changes. The feature switch should switch on/off system level and device level simultaneously.

sensor/FamilyProtectPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+            start: this.start,
+            stop: this.stop
+        });
+        await exec(`mkdir -p ${dnsmasqConfigFolder}`);
+        if (fc.isFeatureOn("family_protect")) {
+            await this.globalOn();
+        } else {
+            await this.globalOff();
+        }
+        fc.onFeature("family_protect", async (feature, status) => {
```
jasonlyc: Also need to check fc.isFeatureOn("family_protect") during startup to enable/disable feature immediately.

sensor/FamilyProtectPlugin.js
```diff
@@ -0,0 +1,193 @@
...
+        await fs.unlink(configFile, err => {
+            if (err) {
+                if (err.code === 'ENOENT') {
+                    log.info(`Dnsmasq: No ${configFile}, skip remove`);
+                } else {
+                    log.warn(`Dnsmasq: Error when remove ${configFile}`, err);
+                }
+            }
+        })
+        dnsmasq.restartDnsmasq();
```
drahnieR: is restart necessary here? or a reload will suffice

PR #1614, opened 2019-07-16
per device adblock

wzjjack:


extension/dnsmasq/dnsmasq.js
```diff
@@ -692,9 +694,9 @@ module.exports = class DNSMASQ {
       if (type === "family") {
         targetIP = BLUE_HOLE_IP
       }
-
+      const tag = this.level[type] == "system" ? "" : "$ad_block";
```
jasonlyc: Do not hardcode ad_block here. You can use value of 'type' as tag, here type is 'adblock'.

sensor/AdblockPlugin.js
```diff
@@ -0,0 +1,175 @@
...
+    async run() {
+        this.systemSwitch = false;
+        this.adminSystemSwitch = false;
+        this.enabledMacAddresses = [];
+        extensionManager.registerExtension("adblock", this, {
+            applyPolicy: this.applyPolicy,
+            start: this.start,
+            stop: this.stop
+        });
+        const isPolicyEnabled = await spt.isPolicyEnabled('adblock');
```
MelvinTo: this should be one-time only. After this upgrade, if adblock policy:system flag is on and adblock feature flag is off, the feature flag should be kept as off

sensor/FamilyProtectPlugin.js
```diff
@@ -49,6 +51,10 @@ class FamilyProtectPlugin extends Sensor {
             start: this.start,
             stop: this.stop
         });
+        const isPolicyEnabled = await spt.isPolicyEnabled('family');
```
MelvinTo: this should be one-time only. After this upgrade, if adblock policy:system flag is on and adblock feature flag is off, the feature flag should be kept as off

MelvinTo:
a general comment is old/new family feature should work with old/new app. (fully compatible)

old family <-> old app (yes)
old family <-> new app (can still manage family with policy:system toggle
new family <-> old app (can still manage family with policy:system toggle, feature flag is on)
new family <-> new app (can manage family with feature flag + all devices + specified devices)

this is also applied to adblock

PR #1618, opened 2019-07-19
Enhancement: include spoofMeName when finalizing bname

wzjjack:


util/util.js
```diff
@@ -45,6 +45,10 @@ function getPreferredBName(hostObject) {
     return null;
   }
 
+  if(hostObject.spoofMeName) {
```
MelvinTo: cloudName should be high priority than spoofMeName

PR #1629, opened 2019-07-22
Fix error when firemain startup

wzjjack:
ERROR Dnsmasq: Update adblock filters Failed! { Error: ENOENT: no such file or directory, rename '/home/pi/.firewalla/config/dns/adblock_filter.conf.tmp' -> '/home/pi/.firewalla/config/dns/adblock_filter.conf'

### DIFF
--- a/sensor/AdblockPlugin.js
+++ b/sensor/AdblockPlugin.js
@@ -130,17 +130,17 @@ class AdblockPlugin extends Sensor {
     async applySystemAdblock() {
         if (this.systemSwitch && this.adminSystemSwitch) {
             const adblocktagset = `mac-address-tag=%${systemLevelMac}${dnsTag}\n`;
-            await fs.writeFile(systemConfigFile, adblocktagset);
+            await fs.writeFileAsync(systemConfigFile, adblocktagset);
         } else {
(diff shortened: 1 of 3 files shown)


MelvinTo:
better not use sync function

sensor/AdblockPlugin.js
```diff
@@ -130,38 +130,41 @@ class AdblockPlugin extends Sensor {
...
         dnsmasq.controlFilter('adblock', this.adminSystemSwitch);
     }
 
     async applyDeviceAdblock() {
         const macAddressArr = this.enabledMacAddresses;
         if (macAddressArr.length > 0 && this.adminSystemSwitch) {
-            const adblocktagset = `mac-address-tag=%${macAddressArr.join("%")}${dnsTag}\n`;
-            await fs.writeFile(deviceConfigFile, adblocktagset);
+            let adblocktagset = "";
+            macAddressArr.forEach((macAddress) => {
```
MelvinTo: just FYI: do not use forEach when you are going to call async function in the inner loop of forEach. Here just a string concat is fine.

use 
```
for(const element of array) {
  await ...
}
```

because **forEach** is Pseudo concurrent execution, the async functions can't be "await"-ed by caller.

PR #1632, opened 2019-07-23
detach NetworkManager from managing eth0

jasonlyc:


scripts/prep/09_detach_eth0_network_manager.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+if [[ -e /etc/NetworkManager/NetworkManager.conf ]]; then
```
MelvinTo: should only make change if not yet..

PR #1638, opened 2019-07-24
Fix Bug: show be able to detect if port is already being used

wzjjack:


wzjjack:
Refactor

extension/frp/frp.js
```diff
@@ -313,7 +331,23 @@ module.exports = class {
           this.started = true;
           hasTimeout = false;
           this._startHealthChecker();
-          resolve();
+          if (this.name == "support") {
+            // waiting for service to confirm the port being used or not
+            // if the port already being used, will publish message to channel "System:RemoteSupport"
+            const start = new Date() / 1000;
```
MelvinTo: It's not necessary to periodically check start code every second, use event-based approach to trigger based by event

e.g.
```
sem.once('success', () => { // do something })
sem.once('failure', () => {// do something })
setTimeout(() => ...) // total timeout, e.g. 10 seconds
```

wzjjack: Updated per your requested

PR #1645, opened 2019-07-26
compatibility: adblock and family protect

wzjjack:


sensor/AdblockPlugin.js
```diff
@@ -99,6 +99,10 @@ class AdblockPlugin extends Sensor {
             if (ip === '0.0.0.0') {
                 if (policy == true) {
                     this.systemSwitch = true;
+                    if (!fc.isFeatureOn("adblock")) {
```
jasonlyc: If system level adblock is enabled and then feature is disabled, this will enable the feature unexpectedly.

You can use 
```
if (fc.isFeatureOn("adblock", true)) {
  // enable dynamic feature flag
}
```

PR #1650, opened 2019-07-29
Use new dnsmasq config directory for latest dnsmasq binaries

jasonlyc:


jasonlyc:
firewalla/firewalla.ios#1700

PR #1652, opened 2019-07-29
empty legacy dnsmasq conf dir as well in main.js

jasonlyc:


jasonlyc:
firewalla/firewalla.ios#1700

PR #1662, opened 2019-08-05
Expose .lan domains to firewalla managed devices 

wzjjack:


extension/dnsmasq/dnsmasq.js
```diff
@@ -1323,4 +1328,53 @@ module.exports = class DNSMASQ {
...
+          }
+        }
+      }
+      if (needUpdate || isInit) {
+        await rclient.setAsync(LOCAL_DEVICE_DOMAIN_KEY, JSON.stringify(deviceDomainMap));
+        let localDeviceDomain = "";
+        for (const key in deviceDomainMap) {
+          const deviceDomain = deviceDomainMap[key]
+          //replace space to dot
+          const domain = deviceDomain.name.replace(/\s+/g, ".").toLowerCase();
```
jerrchen: there may be other characters in the host name that may not be domain name friendly.  Like ","  "-" likely not allowed as first character. 

also deal with duplicated names?  host names can be duplicated

wzjjack: How to deal with duplicated names?  override?

jerrchen: Anything is fine:
1. override:  what ever we see last get that IP address ...  User will figure out and change the host name themselves.
2. add something unique to the name:  If we do this, we need to have UI present the name ... which is ugly, since this feature is for power users ... most 'normal' users won't care what a dot LAN is.
3. ignore both:  user will find out ... and change one of the names

I would go to 1, pick the criteria, like “latest" based on timestamp or something get the IP mapping. the other one ... don't. Absolutely avoid (2)

extension/dnsmasq/dnsmasq.js
```diff
@@ -1323,4 +1329,60 @@ module.exports = class DNSMASQ {
...
+          }
+        }
+        for (const key in domainMap) {
+          const domain = getCanonicalizedHostname(key.replace(/\s+/g, ".")) + '.lan';
+          localDeviceDomain += `address=/${domain}/${domainMap[key].ipv4Addr}\n`
+        }
+        await fs.writeFileAsync(LOCAL_DEVICE_DOMAIN, localDeviceDomain);
+      }
+      if (needUpdate && restart) {
+        this.restart()
```
jasonlyc: Do not use this.restart(). Use this.restartDnsmasq() instead. It can introduce a cooldown time between each restart

wzjjack: I use this.restartDnsmasq() before, but it looks like cannot work when I update device name.

wzjjack: Figure out the issue. I used dnsmaq in Netbot.js before, it is in FireApi progress. So restartDnsmasq can not work. Refactor code in Netbot.js, trigger device update by the event.

PR #1665, opened 2019-08-07
use customized device name in alarm, remove alexa intel

drahnieR:


util/util.js
```diff
@@ -45,6 +45,10 @@ function getPreferredBName(hostObject) {
     return null;
   }
 
+  if (hostObject.name) {
```
MelvinTo: ".name" is a special key for user customized name, should not be included in getPreferredBName function.

PR #1666, opened 2019-08-07
patch beta 7 staging

drahnieR:


MelvinTo:
patch to include:
1. block port bug
2. offline notification improvement
3. device name issue in notification
4. resolv.conf bug
5. 8863

PR #1676, opened 2019-08-09
Alpha - Follow up patch

MelvinTo:


wzjjack:
Patch: Add timestamp in alarm is right.

PR #1689, opened 2019-08-14
upnp exception match fix

drahnieR:


sensor/UPNPSensor.js
```diff
@@ -123,15 +123,15 @@ class UPNPSensor extends Sensor {
...
+                    'p.upnp.private.port' : current.private.port.toString(),
                     'p.upnp.protocol'     : current.protocol,
-                    'p.upnp.enabled'      : current.enabled,
+                    'p.upnp.enabled'      : current.enabled.toString(),
                     'p.upnp.description'  : current.description,
-                    'p.upnp.ttl'          : current.ttl,
-                    'p.upnp.local'        : current.local,
-                    'p.device.port': current.private.port,
+                    'p.upnp.ttl'          : current.ttl.toString(),
+                    'p.upnp.local'        : current.local.toString(),
```
jerrchen: are there cases where any of the current.public or current.private ... etc ... null or undefined?

PR #1690, opened 2019-08-14
Need DNS based block in rules

wzjjack:


wzjjack:
to do:  add/remove a domain in a category; block domain expired
doing: category block

alarm/PolicyManager2.js
```diff
@@ -1020,13 +1023,19 @@ class PolicyManager2 {
...
       case "category":
-        await Block.setupRules(scope && pid, target, "hash:ip", whitelist);
-        await Block.addMacToSet(scope, Block.getMacSet(pid));
-
-        if (!scope && !whitelist && target === 'default_c') try {
-          await categoryUpdater.iptablesRedirectCategory(target)
-        } catch(err) {
-          log.error("Failed to redirect default_c traffic", err)
+        if(policy.dnsmasq_entry){
+          await domainBlock.blockCategory(target,{
```
jasonlyc: Need to dynamically update domain lists in dnsmasq config as well. Looks like this function only fetches the domain list when the policy is created.

wzjjack: Yes, I am doing that.

control/DomainBlock.js
```diff
@@ -249,16 +256,93 @@ class DomainBlock {
...
+  async blockCategory(category, options) {
+    const domains = await this.getCategoryDomains(category);
+    await dnsmasq.addPolicyFilterEntry(domains, options).catch((err) => undefined);
+    sem.emitEvent({
+      type: 'ReloadDNSRule',
+      message: 'DNSMASQ filter rule is updated',
+      toProcess: 'FireMain',
+      suppressEventLogging: true
+    })
+    sem.on('UPDATE_CATEGORY_DYNAMIC_DOMAIN', async (event) => {
```
jasonlyc: I suggest to add functions in CategoryUpdater.js, which updates category domain list file dynamically. It can be hooked into CategoryUpdater.updateDomain, just as what is done to ip set update.
When dnsmasq-based category block is enabled, simply add a symbolic link in dnsmasq config directory to the corresponding category domain list. This can decouple the category domain list update and category domain block enforcement, the same way as ip set update and IPTables update in IPTables-based category block.

PR #1710, opened 2019-08-22
Bug: device name should be enriched for ssh guessing alarm

wzjjack:


extension/bro/BroNotice.js
```diff
@@ -48,6 +48,10 @@ class BroNotice {
       alarm["p.local.decision"] == "ignore";
       return;
     }
+    let deivcesName = [];
+    addresses.map(async(address) => {
+      deivcesName.push(await hostTool.getName(address));
+    })
```
MelvinTo: this map(async () => {}) like function is not recommended.

the async function in the map function is not guaranteed to return before processSSHScan, so the devicesName could be empty.

Should use `for(const address of addresses) ...`

Another issue is if the name of device does not exist, it should use ip address instead.

deivcesName .......... => typo...

wzjjack: deivcesName...not typo...

PR #1763, opened 2019-10-14
add safe threshold for system time

drahnieR:


scripts/fireupgrade.sh
```diff
@@ -142,6 +142,7 @@ if [[ ! -f /.dockerenv ]]; then
     ret=$?
     if [[ $ret -ne 0 ]]; then
         sudo systemctl stop ntp
+        sudo date -s "2019-10-14" # set minimal date here to prevent SSL failure on undergoing HTTPS calls
```
MelvinTo: also I saw once in production, the date info return from websites was very old.

so need to validate the date entry from website before using them.

PR #1776, opened 2019-10-18
Bug: Should detect error when redis is down on firekick diag UI

wzjjack:


extension/diag/app.js
```diff
@@ -201,8 +212,8 @@ class App {
 
       await exec(cmd);
       return imagePath;
-    } catch(err) {
-      return null;
+    } catch (err) {
+      throw new Error("Failed to get QRImage", err);
```
MelvinTo: throw error will cause more problem here?

MelvinTo: because it will skip all remaining code in the main function

PR #1847, opened 2019-12-25
Fix apply network ip when curren subnet change

canyara:


net2/ModeManager.js
```diff
@@ -116,6 +117,13 @@ async function _changeToAlternativeIpSubnet() {
   const altGateway = fConfig.alternativeInterface.gateway;
   const oldGateway = sysManager.myGateway();
   const oldIpSubnet = sysManager.mySubnet();
+  // check if is same subnet
+  const currIpSubnet = iptool.cidrSubnet(oldIpSubnet);
+  const altIp = iptool.cidrSubnet(altIpSubnet);
+  if (!currIpSubnet.contains(altIp.networkAddress) || !currIpSubnet.contains(altGateway)) {
```
MelvinTo: two cidr should exact match each other

PR #1859, opened 2020-01-09
fix:can not mute all security alarm

wzjjack:


alarm/AlarmManager2.js
```diff
@@ -1587,4 +1587,14 @@ module.exports = class {
     log.info("Exception object:", e);
     return e;
   }
+  isMuteSecurityAlarm(alarm) {
+    const securityAlarmTypes = [
+      "ALARM_SPOOFING_DEVICE",
```
MelvinTo: this is not a security alarm, i think it's used to detect if there is another spoofing device in the network that may be conflict with Firewalla simple mode ... @jasonlyc

PR #1866, opened 2020-01-14
Enhancement: Should generate alarm when VPN is disconnected unexpectedly #22

blackfirespace:


MelvinTo:
* need provide notification description in en.json for two types of alarms.

net2/HostManager.js
```diff
@@ -1408,6 +1428,27 @@ module.exports = class HostManager {
                   }
                 }
               }
+
```
jasonlyc: VPN Connect alarm should be generated only if the connection is successfully reestablished.

net2/HostManager.js
```diff
@@ -1411,6 +1431,31 @@ module.exports = class HostManager {
...
+                      'p.vpn.strictvpn': settings && settings.strictVPN || false
+                    }
+                  );
+                  await alarmManager2.enqueueAlarm(alarm);
+                }
+              }
+              if (updatedPolicy) {
+                updatedPolicy.waitresume = 0;
+                updatedPolicy.running = true;
+                await this.setPolicyAsync("vpnClient", updatedPolicy);
```
jasonlyc: do not call setPolicy in the synchronized part of this function. Return the updated part instead.

net2/HostManager.js
```diff
@@ -1446,7 +1480,43 @@ module.exports = class HostManager {
...
+                          'p.vpn.subtype': settings && settings.subtype,
+                          'p.vpn.devicecount': device_cout,
+                          'p.vpn.displayname': (settings && (settings.displayName || settings.serverBoxName)) || profileId,
+                          'p.vpn.time': broken_time,
+                          'p.vpn.strictvpn': settings && settings.strictVPN || false
+                        }
+                      );
+                      await alarmManager2.enqueueAlarm(alarm);
+                      updatedPolicy.waitresume = (settings.strictVPN) ? 1 : 0;
+                      await this.setPolicyAsync("vpnClient", updatedPolicy);
```
jasonlyc: It's okay to call setPolicy here since it is in the async part of this function.

net2/HostManager.js
```diff
@@ -1485,6 +1555,11 @@ module.exports = class HostManager {
                 await vpnClientEnforcer.unenforceInterfaceDNSRedirect(intfName, ovpnClient.getInterfaceName(), dnsServers);
               }
             }
+
+            const updatedPolicy = this.policy["vpnClient"];
+            if (!updatedPolicy) return;
+            updatedPolicy.waitresume = 0;
+            await this.setPolicyAsync("vpnClient", updatedPolicy);
```
jasonlyc: Same here. Return the updated part in the return variable instead

net2/HostManager.js
jasonlyc: do not call setPolicy in the synchronized part of this function. Return the updated part instead.

jasonlyc: Otherwise, it may trigger concurrent execution of this function

PR #1876, opened 2020-01-19
fix:write local domain conf file at the same time

wzjjack:


extension/dnsmasq/dnsmasq.js
```diff
@@ -1452,7 +1453,14 @@ module.exports = class DNSMASQ {
   //save data in redis
   //{mac:{ipv4Addr:ipv4Addr,name:name}}
   //host: { ipv4Addr: '192.168.218.160',mac: 'F8:A2:D6:F1:16:53',name: 'LAPTOP-Lenovo' }
-  async setupLocalDeviceDomain(restart, hosts, isInit) {
+  async setupLocalDeviceDomain(hosts, isInit) {
+    if (this.updatingLocalDomain) {
```
MelvinTo: If there is a scheduled update, no more scheduled update should be added, simply return.

for current code, if setupLocalDeviceDomain is called X times, the file will be updated X times still. If should be updated only once.

sensor/LocalDomainSensor.js
```diff
@@ -0,0 +1,62 @@
...
+const unlinkAsync = util.promisify(fs.unlink);
+class LocalDomainSensor extends Sensor {
+    constructor() {
+        super();
+    }
+    run() {
+        this.hookFeature(featureName);
+    }
+    async globalOn() {
+        const hosts = await hostManager.getHostsAsync();
```
MelvinTo: maybe only last-7day online devices should be considered to support local domain, this can reduce some file update effort

MelvinTo:
Generally file should be updated only:
* If userLocalDomain is specified, only update if userLocalDomain is changed
* If userLocalDomain is not specified, and preferredName is changed
* For other cases, ignore

wzjjack:
> * s specified, only update if userLocalDomain is changed
> * If userLocalDomain is not specified, and preferredName is ch

+ ip changed

wzjjack:
> Generally file should be updated only:
> 
> * If userLocalDomain is specified, only update if userLocalDomain is changed
> * If userLocalDomain is not specified, and preferredName is changed
> * For other cases, ignore

- add ip changed

sensor/LocalDomainSensor.js
wzjjack: `getHostsAsync ` is to return the online devices of the last 7 days.

Issue #18, opened 2016-09-19
take care of random drops in bone


jerrchen:
bone api some times will fail due to 

2016-09-19T18:31:08.038Z - info: FlowManager:FlowSummary not enough flows
Error while requesting  { [Error: connect ENETUNREACH 10.10.10.10:443]
  code: 'ENETUNREACH',
  errno: 'ENETUNREACH',
  syscall: 'connect',
  address: '10.10.10.10',
  port: 443 } Error
    at Request._callback (/home/pi/firewalla/lib/Bone.js:209:25)
    at self.callback (/home/pi/firewalla/node_modules/request/request.js:187:22)
    at emitOne (events.js:77:13)
    at Request.emit (events.js:169:7)
    at Request.onRequestError (/home/pi/firewalla/node_modules/request/request.js:813:8)
    at emitOne (events.js:77:13)
    at ClientRequest.emit (events.js:169:7)
    at TLSSocket.socketErrorListener (_http_client.js:269:9)
    at emitOne (events.js:77:13)
    at TLSSocket.emit (events.js:169:7)

PR #1879, opened 2020-01-19
Enhancement: Include VPN client connectivity information in checkin data #18

blackfirespace:


net2/Host.js
```diff
@@ -740,6 +740,9 @@ class Host {
       if (!results) return obj;
 
       obj.agents = results;
+      if (this.policy['vpnClient']) {
```
MelvinTo: no need to add if statement here, just move it up to the obj structure, like "monitored"

PR #1883, opened 2020-01-20
Enhancement: Should generate alarm when VPN is disconnected unexpectedly #22

blackfirespace:


alarm/Alarm.js
```diff
@@ -347,7 +347,7 @@ class VPNRestoreAlarm extends Alarm {
   }
 
   localizedNotificationContentArray() {
-    return [this["p.vpn.subtype"], this["p.vpn.devicecount"], this["p.vpn.displayname"], this["p.vpn.time"], this['p.vpn.subtypename'], this["p.vpn.strictvpn"]];
+    return [this["p.vpn.subtype"], this["p.vpn.devicecount"], this["p.vpn.displayname"], this['p.vpn.subtypename'], this["p.vpn.strictvpn"]];
```
MelvinTo: if we are still going to show time in notification, we should still add timestamp here in the array

Issue #35, opened 2016-10-23
systemd not detecting bro crash


jerrchen:
this might be a bug in systemd, even when bro crashed (the pid file is deleted).  Systemd still thinks bro is up.  which causes the system to pretty much dead.  

Need a small cron script to double check bro.  or look at brofish.service and see why it is not detecting the crash.

jerrchen:
this is fixed in bro-cron

PR #1900, opened 2020-01-31
Enhancement: add some flag in init data if no_auto_upgrade is added #35

blackfirespace:


net2/HostManager.js
```diff
@@ -802,6 +802,11 @@ module.exports = class HostManager {
     const versionUpdate = await sysManager.getVersionUpdate();
     if (versionUpdate)
       json.versionUpdate = versionUpdate;
+    if (require('fs').existsSync("/home/pi/.firewalla/config/.no_auto_upgrade")) {
```
MelvinTo: * should not use existsSync
* also no need to check every time when loading init data
* follow SysInfo.js to check the file periodically.

Issue #39, opened 2016-10-24
Need better way to set the pi password 


jerrchen:
This password should be dynamically set by firewalla owner once connected.  And can be reset by the app

jerrchen:
@MelvinTo  has closed this issue.  all new firewalla instances will reset password after the first successful marriage.   There after, every reboot Firewalla will get a new ssh password.

PR #1905, opened 2020-02-05
Enhancement: Scan individual devices to see if any ports are open on devices #39

blackfirespace:


sensor/DeviceServiceScanSensor.js
```diff
@@ -0,0 +1,148 @@
+/*    Copyright 2020 Firewalla LLC
```
wzjjack: Firewalla INC

PR #66, opened 2016-11-29
	refine direct connect mode

MelvinTo:

PR #1907, opened 2020-02-06
Enhancement: use cloud to pull doh server list instead of local config #66

blackfirespace:


extension/dnscrypt/dnscrypt.js
```diff
@@ -51,23 +51,37 @@ class DNSCrypt {
...
+      if (!serverName) {
+        return false;
+      }
+
+      return (allServers.filter((server) => (server && server.name && server.stamp && server.name === serverName)).length > 0) ? true : false;
+    });
+    content = content.replace("%DNSCRYPT_SERVER_LIST%", JSON.stringify(serverList));
+
+    if (reCheckConfig) {
+      if (fs.existsSync(runtimePath)) {
```
MelvinTo: generally, do not use any Sync functions if there is a way to do async

sensor/DNSCryptPlugin.js
```diff
@@ -166,7 +169,7 @@ class DNSCryptPlugin extends Sensor {
   // global on/off
   async globalOn() {
     this.adminSystemSwitch = true;
-    await dc.restart();
+    //await dc.restart();
     await this.applyAll();
```
MelvinTo: I think it looks okay to check the local file every time, even including globalOn

sensor/DNSCryptPlugin.js
```diff
@@ -100,8 +101,10 @@ class DNSCryptPlugin extends Sensor {
     }
   }
 
-  async applyAll() {
-    await dc.prepareConfig({});
+  async applyAll(reCheckConfig = false) {
+    log.info("DNSCryptPlugin.applyAll");
+    const result = await dc.prepareConfig({}, reCheckConfig);
+    if (!result) return;
```
MelvinTo: assume there is a possibility that the dnscrypt service crashed, but the file content is still there nicely.

The current code will not restart the service.

better to call `sudo systemctl start dnscrypt` to ensure it's started up

MelvinTo:
need resolve conflict

extension/dnscrypt/dnscrypt.js
```diff
@@ -51,23 +53,38 @@ class DNSCrypt {
...
 
     content = content.replace("%DNSCRYPT_ALL_SERVER_LIST%", this.allServersToToml(allServers));
 
+    let serverList = await this.getServers();
+    serverList = serverList.filter((serverName) => {
+      if (!serverName) {
+        return false;
+      }
+
+      return (allServers.filter((server) => (server && server.name && server.stamp && server.name === serverName)).length > 0) ? true : false;
```
MelvinTo: try to avoid code like this which is hard to understand the purpose.

what i wrote:

```javascript
    const allServers = await this.getAllServers();
    const allServerNames = await this.getAllServerNames();
    serverList = serverList.filter((n) => allServerNames.includes(n));
```

PR #1928, opened 2020-02-14
Enhancement: Generate alarm when devices are using VPN #52

blackfirespace:


alarm/Alarm.js
```diff
@@ -833,6 +833,17 @@ class PornAlarm extends OutboundAlarm {
   }
 }
 
+class VpnAlarm extends OutboundAlarm {
+  constructor(timestamp, device, pornID, info) {
```
drahnieR: pornID?

PR #1954, opened 2020-02-23
Tag flow

canyara:


alarm/PolicyManager2.js
```diff
@@ -1015,16 +1016,27 @@ class PolicyManager2 {
         }
         break;
 
+      // FIXME support intf
       case "mac":
-        await Block.addMacToSet([target], null, whitelist)
-        accounting.addBlockedDevice(target);
+        if (target.toLowerCase() == "tag") {
```
MelvinTo: why a target could be the string "tag"? What does it mean when it is "tag"?

PR #1963, opened 2020-02-24
Only restart socket between box and fireguard when config changes

wzjjack:


sensor/GuardianSensor.js
```diff
@@ -90,10 +90,10 @@ class GuardianSensor extends Sensor {
       if(!socketioServer) {
         throw new Error("invalid guardian relay server");
       }
-
+      const forceRestart = !this.socket || (await this.getRegion() != data.region) || (await this.getServer() != socketioServer)
```
MelvinTo: how to make sure the existing socket connection is not dead?

PR #1975, opened 2020-02-25
Interface policy support

canyara:


control/Block.js
```diff
@@ -379,14 +380,110 @@ async function setupTagRules(pid, tags, dstTag, dstType, allow = false, destroy
...
 }
 
+async function setupIntfsRules(pid, intfs, dstTag, dstType, allow = false, destroy = false, destroyDstCache = true) {
+  if (_.isEmpty(intfs)) {
+    return;
+  }
+
+  const NetworkProfileManager = require('../net2/NetworkProfileManager.js')
+  for (let index = 0; index < intfs.length; index++) {
+    if (!NetworkProfileManager.getNetworkProfile(intfs[index])) {
```
jasonlyc: You can use NetworkProfile.ensureCreateEnforcementEnv(uuid) to ensure the ip set is created even if the network profile does not exist at the moment. Since network may be restored from config history, the policy on the restored network should take effect at once.

control/Block.js
```diff
@@ -379,14 +380,108 @@ async function setupTagRules(pid, tags, dstTag, dstType, allow = false, destroy
...
 }
 
+async function setupIntfsRules(pid, intfs, dstTag, dstType, allow = false, destroy = false, destroyDstCache = true) {
+  if (_.isEmpty(intfs)) {
+    return;
+  }
+
+  const NetworkProfile = require('../net2/NetworkProfile.js');
+  for (let index = 0; index < intfs.length; index++) {
+    NetworkProfile.ensureCreateEnforcementEnv(intfs[index]);
```
jasonlyc: await

control/Block.js
```diff
@@ -379,14 +380,108 @@ async function setupTagRules(pid, tags, dstTag, dstType, allow = false, destroy
...
+        natOutRule6.mth(dstSet6, 'dst');
+      }
+
+      const inRule     = new Rule().chn(filterChain).jmp(filterDest).comment(comment)
+      const inRule6    = new Rule().chn(filterChain).jmp(filterDest).fam(6).comment(comment)
+      const natInRule  = new Rule('nat').chn(natChain).jmp(natDest).comment(comment)
+      const natInRule6 = new Rule('nat').chn(natChain).jmp(natDest).fam(6).comment(comment)
+
+      if (dstSet) {
+        inRule.mth(dstSet, 'dst');
```
jasonlyc: shoud be src

jasonlyc: Also need to fix this in setupTagRules

PR #2004, opened 2020-03-04
Bug: should not crash FireX process if log file is not accessible

wzjjack:


scripts/utils.sh
```diff
@@ -8,6 +8,7 @@ function setup_folders() {
     mkdir -p ~/.firewalla/run
     mkdir -p ~/.forever
     mkdir -p ~/logs
+    sudo chown -R pi ~/logs
```
MelvinTo: minor, better use `sudo chown -R pi ~/logs/`

the logs folder is a symlink, "-R" does follow symlink, but add / is for better reading.

PR #2015, opened 2020-03-07
Delete tag related info

canyara:


net2/TagManager.js
```diff
@@ -110,6 +110,22 @@ class TagManager {
         await rclient.delAsync(key);
         this.subscriber.publish("DiscoveryEvent", "Tags:Updated", null, this.tags[uid].o);
         await this.refreshTags();
+
```
jasonlyc: Better to move these logics into Tag.destroyEnv

canyara: The FireAPI process does not call Tag.destroyEnv

canyara: > Better to move these logics into Tag.destroyEnv

The FireAPI process does not call Tag.destroyEnv

jasonlyc: It can be done in the background

PR #2037, opened 2020-03-12
Fix: Check dhcp server status

wzjjack:
- make site to site limit on Blue is 1, and Gold is 10
- Fix: Check dhcp server status
- Bug: should restart firemain if timezone is changed in the runtime


platform/red/RedPlatform.js
```diff
@@ -78,6 +78,9 @@ class RedPlatform extends Platform {
   isFireRouterManaged() {
     return false;
   }
+  getAllowCustomizedProfiles(){
+    return 1;
```
MelvinTo: no profile on red

net2/SysManager.js
```diff
@@ -335,10 +335,15 @@ class SysManager {
     return tz;
   }
 
-  setTimezone(timezone, callback) {
+  async setTimezone(timezone, callback) {
```
MelvinTo: confused why async + callback together

net2/SysManager.js
```diff
@@ -335,34 +335,24 @@ class SysManager {
...
-          callback(null)
-        } catch (err) {
-          log.error("Failed to set timezone:", err);
-          callback(err)
-        }
-      })()
-    });
+      await exec(`sudo timedatectl set-timezone ${timezone}`);
+      await exec('sudo systemctl restart cron.service');
+      await exec('sudo systemctl restart firemain');
```
MelvinTo: no need to restart for the first time, when this.timezone was null

PR #2053, opened 2020-03-17
Enhancement: Api to customize the DNS settings in /etc/resolvconf/resolv.conf.d/head #108

blackfirespace:


controllers/netbot.js
```diff
@@ -3832,6 +3832,32 @@ class netBot extends ControllerBot {
         })
         break;
       }
+      case "getDNSProfile": {
```
MelvinTo: follow this to move get/set logic to the same sensor file

https://github.com/firewalla/firewalla/blob/9e21f28efa341e0ecaebff81d4302dacc03675e4/sensor/GuardianSensor.js#L50

sensor/SystemDNSSensor.js
```diff
@@ -0,0 +1,74 @@
...
+const extensionManager = require('./ExtensionManager.js')
+const systemDNSKey = 'sys:dns:custom';
+
+class SystemDNSSensor extends Sensor {
+  constructor() {
+    super();
+  }
+
+  async run() {
+    sem.on('SystemDNSUpdate', async (event) => {
```
MelvinTo: should auto set this when system restarts

PR #2080, opened 2020-03-24
Enhancement: provide api to check if a domain, ip, along with port is blocked by Firewalla #89

blackfirespace:
search exceptions and dnsmasq


alarm/ExceptionManager.js
```diff
@@ -470,4 +471,64 @@ module.exports = class {
     }
   }
 
+  searchExceptionValue(exception, val2) {
```
MelvinTo: why not reuse the existing match logic? create a fake alarm object, and check if the exception matches.

PR #2082, opened 2020-03-24
Enhancement: support checking tcp openvpn port in diag.firewalla.com

wzjjack:


sensor/VPNCheckPlugin.js
```diff
@@ -73,7 +74,8 @@ class VPNCheckPlugin extends Sensor {
       method: "POST",
       uri: api,
       json: {
-        tls_auth: taKey
+        tls_auth: taKey,
+        protocol: platform.getVPNServerDefaultProtocol()
```
MelvinTo: not the default protocol, should be the configured protocol. defined in policy:system -> vpn

if no protocol is defined in policy:system, use the default one

```
"{\"state\":true,\"serverNetwork\":\"10.176.49.0\",\"netmask\":\"255.255.255.0\",\"localPort\":\"1194\",\"externalPort\":\"1194\",\"portmapped\":false,\"protocol\":\"tcp\"}"
```

PR #2104, opened 2020-04-01
Enhancement: Detect if Firewalla is connected to a new router #80

blackfirespace:


sensor/RouterChangeSensor.js
```diff
@@ -0,0 +1,64 @@
...
+      await this.checkRouteMac();
+    });
+  }
+
+  async checkRouteMac() {
+    if (!Mode.isSpoofModeOn()) {
+      return;
+    }
+
+    const gatewayIp = sysManager.myGateway();
```
drahnieR, line 49: this has to be multi-interface/gold compatible

sensor/RouterChangeSensor.js
```diff
@@ -0,0 +1,64 @@
...
+    const gatewayMac = await hostTool.getMacByIP(gatewayIp);
+    const isMember = await rclient.sismemberAsync(routerMacKey, gatewayMac);
+    const hasAny = await rclient.scardAsync(routerMacKey);
+    if (!isMember && hasAny > 0) {
+      await modeManager.setNoneAndPublish();
```
drahnieR, line 54: turn off monitoring on a specific network might be a better solution
use policy targeting `network:<uuid>`, check https://github.com/firewalla/firewalla/blob/cc204133ef5f9c425e702b3077fd4f6886e381cd/controllers/netbot.js#L247

drahnieR: Also, we should send a notification to App as well. With this flag only, user won't get notified until he/she opens the app.

net2/HostManager.js
```diff
@@ -1008,6 +1008,9 @@ module.exports = class HostManager {
         if(firstBinding) {
           json.firstBinding = firstBinding
         }
+        if (json.mode === "spoof" || json.mode === "none") {
+          json.routerChangeFlag = await rclient.getAsync("sys:router:change");
```
drahnieR, line 1012: Client should be able to know what interfaces have problems so it can proceed on compatibility test.

PR #2109, opened 2020-04-01
Bug:devices name does not update

wzjjack:


hook/NewDeviceHook.js
```diff
@@ -104,15 +109,16 @@ class NewDeviceHook extends Hook {
         if(!name) {
           return // hostname is not provided by dhcp request, can't update name
         }
-
-        log.info("MAC Address", mac, " already exists, updating backup name");
-        sem.emitEvent({
-          type: "RefreshMacBackupName",
```
drahnieR, line 110: This event no longer used? I see reference in `hook/RefreshMacBackupNameHook.js`

wzjjack: no longer used.

PR #2129, opened 2020-04-09
https://github.com/firewalla/general_issues/issues/81

blackfirespace:


extension/dnsmasq/dnsmasq.conf
```diff
@@ -46,6 +46,13 @@ log-facility=/dev/null
 
 strict-order
 
+# If you don't want dnsmasq to read /etc/hosts, uncomment the
+# following line.
+#no-hosts
+# or if you want it to read another file, as well as /etc/hosts, use
+# this.
+addn-hosts=/home/pi/.firewalla/run/dnsmasq_addn_hosts
```
jasonlyc, line 54: It's better to add this into FILTER_DIR of dnsmasq.js instead of changing static config file for compatibility of all platforms

sensor/LocalDomainSensor.js
```diff
@@ -72,6 +74,7 @@ class LocalDomainSensor extends Sensor {
     async globalOff() {
         try {
             await unlinkAsync(LOCAL_DOMAIN_FILE);
+            await unlinkAsync(ADDN_HOSTS_FILE);
```
jasonlyc, line 77: You can just delete the add host config file from FILTER_DIR, leaving the real hosts file intact.

PR #2137, opened 2020-04-10
Fix Android app notification localization issue

wesley-firewalla:


bin/real
```diff
@@ -1 +0,0 @@
-real.armv7l
```
MelvinTo, line 1: why delete this?

controllers/netbot.js
```diff
@@ -497,9 +497,9 @@ class netBot extends ControllerBot {
         notifMsg["title-loc-args"] = newArray;
         notifMsg["loc-key"] = alarm.localizedNotificationContentKey();
         notifMsg["loc-args"] = alarm.localizedNotificationContentArray();
-        notifMsg["title_loc_key"] = alarm.localizedNotificationTitleKey();
+        notifMsg["title_loc_key"] = alarm.localizedNotificationTitleKey().replace(/[.:]/g, '_');
```
MelvinTo, line 500: is this absolutely necessary? if changing box side, the localization support for android will be delayed for at least three months.

bin/real
wesley-firewalla: Revert this changes for now. When got time, will test it on firewalla red.

controllers/netbot.js
wesley-firewalla: The `-` and `:` can not be used in resource name in android. So we can only fix it in box currently.

PR #2140, opened 2020-04-13
Enhancement: periodically cleanup rdns:* keys #87

blackfirespace:


sensor/OldDataCleanSensor.js
```diff
@@ -503,10 +534,15 @@ class OldDataCleanSensor extends Sensor {
 
     setTimeout(() => {
       this.scheduledJob();
+      this.dailyJob();
```
drahnieR, line 537: Don't run this at startup, or at least not immediately after startup

PR #2162, opened 2020-04-20
set localhost as dns server for resolving domain when doh enable

wzjjack:


control/DomainBlock.js
```diff
@@ -182,6 +183,14 @@ class DomainBlock {
   }
 
   async resolveDomain(domain) {
+    if (fc.isFeatureOn('doh')) {
+      resolver.setServers(['127.0.0.1:8853']);
```
MelvinTo, line 187: instead of 8853, use doh api to get the listening port on doh (dnscrypt) ==> 127.0.0.1:8854

do not hard code 8853/8854, use api to get the actual port.

MelvinTo: 8854 should be a configurable port from doh js file itself

PR #2165, opened 2020-04-21
Enhancement: a flag in checkin that it will temperarily shutdown the simple mode if the flag is there #82

blackfirespace:


sensor/BoneSensor.js
```diff
@@ -226,6 +233,32 @@ class BoneSensor extends Sensor {
     }
   }
 
+  async checkCloudSpoofOff(spoofOff) {
```
drahnieR, line 236: don't duplicate code, use `HostManager.spoof()` here.

scripts/bitbridge-ping.sh
```diff
@@ -33,7 +33,7 @@ is_bitbridge7_active () {
     fi
 }
 
-if is_spoof_mode && is_bitbridge7_active; then
+if [[ is_spoof_mode && ! -e /home/pi/firewalla/bin/dev && is_bitbridge7_active ]]; then
```
drahnieR, line 36: this is confusing, add check of `sys:bone:spoofOff` to is_spoof_mode() instead?

PR #2180, opened 2020-04-23
Enhancement: Need a way to detect if a firewalla repo is changed by user #147

blackfirespace:


sensor/BoneSensor.js
```diff
@@ -142,6 +142,7 @@ class BoneSensor extends Sensor {
   }
 
   async checkIn() {
+    this.getLocalRepoChangeFile();
```
drahnieR, line 145: why not await?

sensor/BoneSensor.js
```diff
@@ -285,6 +289,17 @@ class BoneSensor extends Sensor {
       }
     })
   }
+
+  getLocalRepoChangeFile() {
+    if (!this.localRepoChangeFiles) {
+      const cmd = "git status -uno --porcelain | awk '{print $1, $2}' | head -n 10";
```
drahnieR, line 295: change on firerouter should be included as well.

util/i18n.js
```diff
@@ -1,35 +1,53 @@
-'use strict'
-
-let i18n = require('i18n');
+'use strict';
 
+const log = require('../net2/logger.js')(__filename);
```
drahnieR, line 3: what are these changes about?

PR #2193, opened 2020-04-26
Enhancement: Add batch action api

wzjjack:
- delete alarms
- delete rules

### DIFF
--- a/controllers/netbot.js
+++ b/controllers/netbot.js
@@ -2726,6 +2726,22 @@ class netBot extends ControllerBot {
           this.simpleTxData(msg, null, err, callback)
         })
         break;
+      case "policy:batch":
+        (async () => {
+          /*
+            actions: {create: [policy instance], delete: [policy instance], update:[policyID]}
+          */
+          const actions = value.actions;
+          if (actions) {
+            const result = await pm2.batchPolicy(actions)
+            this.simpleTxData(msg, result, null, callback);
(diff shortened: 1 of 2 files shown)


controllers/netbot.js
```diff
@@ -2691,13 +2698,45 @@ class netBot extends ControllerBot {
...
+            await Promise.all(policyIDs.map(async(policyID)=>{
+              let policy = await pm2.getPolicy(policyID)
+              if (policy) {
+                await pm2.disableAndDeletePolicy(policyID)
+                policy.deleted = true;
+                results[policyID] = policy;
+              } else {
+                results[policyID] = "invalid policy";
+              }
+              this.simpleTxData(msg, results, null, callback);
```
MelvinTo, line 2713: this will be caused multiple times.

controllers/netbot.js
```diff
@@ -2590,7 +2590,14 @@ class netBot extends ControllerBot {
       case "alarm:delete":
         try {
           (async () => {
-            await am2.removeAlarmAsync(value.alarmID);
+            const alarmIDs = value.alarmIDs;
+            if (alarmIDs && _.isArray(alarmIDs)) {
+              await Promise.all(alarmIDs.map(async (alarmID) => {
```
MelvinTo, line 2595: prefer forloop as performance is not that important here, but stability is.

PR #2354, opened 2020-06-02
Enhancement: Batch operation on box api

wzjjack:


controllers/netbot.js
```diff
@@ -2733,29 +2806,13 @@ class netBot extends ControllerBot {
         break;
       case "policy:delete":
         (async () => {
-          const policyIDs = value.policyIDs;
```
drahnieR, line 2736: will this break existing clients?

alarm/ExceptionManager.js
```diff
@@ -487,4 +487,40 @@ module.exports = class {
...
+      'create': [],
+      'update': [],
+      'delete': []
+    };
+    for (const action in actions) {
+      const rawData = actions[action] || [];
+      switch (action) {
+        case 'create':
+        case 'update':
+          const emAction = action == 'create' ? 'createException' : 'updateException';
```
drahnieR, line 502: this could break some IDE functionalities, maybe
```
const emAction = action == 'create' ? this.createException : this.updateException
emAction.bind(this)(exception)
```

controllers/netbot.js
wzjjack: I added this a few weeks ago, not used yet.

alarm/PolicyManager2.js
```diff
@@ -2118,46 +2118,57 @@ class PolicyManager2 {
...
           for (const rawPolicy of rawData) {
             const { policy, alreadyExists } = await this.checkAndSaveAsync(new Policy(rawPolicy));
-            let result = policy;
-            if (alreadyExists == 'duplicated') {
-              result = 'duplicated'
-            }
-            results[action].push(result);
+            results[action].push({
+              policy: policy,
+              error: alreadyExists == 'duplicated'
```
drahnieR, line 2123: try ... catch on every operation?
maybe we should even return `alreadyExists` as a rejection

alarm/PolicyManager2.js
```diff
@@ -2118,46 +2118,57 @@ class PolicyManager2 {
...
+            try {
+              const oldPolicy = await this.getPolicy(pid);
+              await this.updatePolicyAsync(rawPolicy);
+              newPolicy = await this.getPolicy(pid);
+              this.tryPolicyEnforcement(newPolicy, 'reenforce', oldPolicy);
+            } catch (e) {
+              log.warn(`update policy ${rawPolicy} error`,e);
+            }
+            results[action].push({
+              policy: newPolicy,
```
drahnieR, line 2140: newPolicy will be `undefined` in case of error

drahnieR: also, do we need more error status than true/false?

alarm/PolicyManager2.js
```diff
@@ -2106,45 +2106,69 @@ class PolicyManager2 {
...
+            }
+            results[action].push({
+              policy: policy,
+              error: !policy
+            });
+          }
+          break;
+        case 'enable':
+        case 'disable':
+          const pmAction = action == 'enable' ? this.enablePolicy : this.disablePolicy;
```
drahnieR, line 2162: should we consider combine enable/disable into update? this feels like extra complexity

alarm/PolicyManager2.js
wzjjack: Yes, it will be undefined in case of error, so the error will be true. Keep `policy:batch` response data structure same would be helpful.

alarm/PolicyManager2.js
wzjjack: yes, may some policies update will succeed but some not. Need return failures and success

alarm/PolicyManager2.js
wzjjack: enable and disable may only pass policy ID

alarm/PolicyManager2.js
drahnieR: I mean try/catch on calls to `checkAndSaveAsync` and `updatePolicyAsync`. No guarantee it'll return without exception.

drahnieR: and is there a way for client to know if the create fails on duplication or other reason?

PR #2355, opened 2020-06-03
Bug: dns resolve set servers error

wzjjack:


control/DomainBlock.js
```diff
@@ -46,10 +46,14 @@ const BlockManager = require('../control/BlockManager.js');
 const blockManager = new BlockManager();
 
 const _ = require('lodash');
+let instance;
 class DomainBlock {
 
   constructor() {
-
+    if (instance == null) {
```
drahnieR, line 53: this is not necessary, `module.exports = new Class()` will be enough to ensure singleton

PR #2361, opened 2020-06-04
Enhancement: Add a new scan sensor to detect if any devices are using default username/passwords

blackfirespace:


sensor/InternalScanSensor.js
```diff
@@ -0,0 +1,247 @@
...
+      };
+    } catch (err) {
+     log.error("Failed to scan: " + err);
+    }
+    log.info('Scan end');
+  }
+
+  isSensorEnable() {
+    return fc.isFeatureOn("internal_scan");
+  }
```
wzjjack, line 108: when feature switch to enable from disabling, should run check.

sensor/InternalScanSensor.js
```diff
@@ -0,0 +1,247 @@
...
+    //this.supportPorts = ["tcp_23"];
+    sem.once("DeviceServiceScanComplete", async (event) => {
+      await this.checkAndRunOnce();
+    });
+  }
+
+  async checkAndRunOnce() {
+    await this.runOnce();
+
+    const interval = this.config.interval * 1000 || 7 * 24 * 60 * 60 * 1000; // one week
```
wzjjack, line 61: when feature switch to enable from disabling, should run check.

sensor/InternalScanSensor.js
```diff
@@ -0,0 +1,247 @@
...
+    }
+    //this.supportPorts = ["tcp_23"];
+    sem.once("DeviceServiceScanComplete", async (event) => {
+      await this.checkAndRunOnce();
+    });
+  }
+
+  async checkAndRunOnce() {
+    await this.runOnce();
+
```
wzjjack, line 60: when feature switch to enable from disabling, should run check.

PR #2382, opened 2020-06-09
Enhancement: Batch operation on box api

wzjjack:


controllers/netbot.js
```diff
@@ -4272,6 +4276,50 @@ class netBot extends ControllerBot {
...
+      const batchActionObjArr = rawmsg.message.obj.data.value;
+      const copyRawmsg = JSON.parse(JSON.stringify(rawmsg));
+      const results = [];
+      for (const obj of batchActionObjArr) {
+        obj.type = "jsonmsg"
+        obj.data.ignoreRate = true;
+        copyRawmsg.message.obj = obj;
+        let result, error;
+        try {
+          result = await this.msgHandlerAsync(gid, copyRawmsg);
```
drahnieR, line 4306: `msgHandler` calls `cmdHandler` essentially, and that latter calls `this.simpleTxData` on every command, this should not happen.

wzjjack: call `msgHandlerAsync` will pass a callback to simpleTxData, and it will trigger callback and return directly. I tested it and works fine.

wzjjack: `msgHandler` will call `this.simpleTxData ` without a callback and send data to the cloud, but `msgHandlerAsync ` will pass a callback, just resolve promise and return a result for every single action, like create policy or something else.

PR #2411, opened 2020-06-17
use full license

MelvinTo:


cli/heartbeat.js
```diff
@@ -127,7 +127,7 @@ async function getLatestCommitHash(cwd) {
 
 async function getLicenseInfo() {
   const licenseFile = "/home/pi/.firewalla/license";
-  const SUUID = (await getShellOutput(`awk '/SUUID/ {print $NF}' ${licenseFile}`)).replace(/[",]/g,'');
+  const SUUID = (await getShellOutput(`awk '/"UUID"/ {print $NF}' ${licenseFile}`)).replace(/[",]/g,'');
```
jimyang2008: variable name should be UUID as well

PR #2457, opened 2020-07-07
fix condition check

drahnieR:


scripts/check_fix_network.sh
```diff
@@ -146,10 +146,10 @@ function check_with_timeout() {
         restored=$RESTORED_AND_NEED_START_OVER
         break;
       else
-        skip=$([ ! -z "$reboot" ] && 'skipped')
+        [ 0 = "$reboot" ] && skip='skipped' || skip=''
```
jimyang2008, line 149: Although '=' works here, it is better to use '-eq' if we know for sure it is a test on non-empty integer.
```
[ 0 -eq $reboot ] && ...
```

scripts/check_fix_network.sh
```diff
@@ -146,10 +146,10 @@ function check_with_timeout() {
...
         echo "fail - reboot $skip"
         $LOGGER "FIREWALLA:FIX_NETWORK:failed to $message, even after restore, reboot $skip"
-        if [ -z "$reboot" ]; then reboot_if_needed; fi
+        if [ 1 = "$reboot" ]; then reboot_if_needed; fi
```
jimyang2008, line 152: Same here, better check with `-eq`
```
if [[ 1 -eq $reboot ]]; then ...
```

PR #2469, opened 2020-07-09
Filter out IPv4 broadcast address for any monitoring interface

jimyang2008:


hook/DeviceHook.js
```diff
@@ -63,10 +63,22 @@ class DeviceHook extends Hook {
...
+      log.warn("Invalid MAC address for process device update:", event);
       return;
     }
 
+    /*
+     * Filter out IPv4 broadcast address for any monitoring interface
+     */
+    if (ipv4Addr) {
+      let monInterfaces = sysManager.getMonitoringInterfaces();
+      let foundInterface = monInterfaces.find(e => e.bcast_address === ipv4Addr)
```
jasonlyc, line 75: const ip = require('ip');
......
monInterfaces.find(e => e.subnet && ip.cidrSubnet(e.subnet).broadcastAddress === ipv4Addr)

util/linux.js
```diff
@@ -87,6 +87,11 @@ exports.gateway_ip_for = function(nic_name) {
   return trim_exec_async("ip r | grep " + nic_name + " | grep default | cut -d ' ' -f 3 | sed -n '1p'");
 };
 
+exports.bcast_address_for = function(nic_name) {
+  var cmd = "ifconfig " + nic_name + " 2> /dev/null | fgrep 'Bcast:' | awk '{print $3}' |awk -F: '{print $2}'";
```
jasonlyc, line 91: No need to do this in case output formay of ifconfig may vary on different platforms

hook/DeviceHook.js
jimyang2008: fixed as suggested

util/linux.js
jimyang2008: reverted as suggested

PR #2496, opened 2020-07-21
support per device/group port scan

wzjjack:


sensor/DeviceServiceScanSensor.js
```diff
@@ -15,138 +15,250 @@
...
-      };
-    } catch(err) {
-      log.error('Failed to scan: ', err);
-    };
+    async applyPolicy(host, ip, policy) {
+        log.info("Applying device service scan policy:", ip, policy);
+        try {
+            if (ip === '0.0.0.0') {
+                if (policy === true) {
+                    this.systemSwitch = true;
```
drahnieR, line 61: looks like `systemSwitch` and `adminSystemSwitch` are the same thing.
policy:system all targets 0.0.0.0

sensor/DeviceServiceScanSensor.js
```diff
@@ -15,138 +15,250 @@
...
+                }
+            }
+        } catch (err) {
+            log.error("Got error when applying adblock policy", err);
+        }
+    }
 
-  async runOnce() {
-    log.info('Scan start...');
+    async applyScan() {
```
drahnieR, line 143: this is a little bit too complex form what i can see.

Can we have a global list of all devices, and mark them as if them should be scanned on each iteration, by system -> network -> group -> device policy? Device policy comes at last and have the highest priority.

And there should be only one interval timer to control the scan. which is controlled by the global switch.

sensor/DeviceServiceScanSensor.js
wzjjack: Nope, they aer different, adminSystemSwitch means feature on or off,  systemSwitch means target=>0.0.0.0 policy is true or  false.

sensor/DeviceServiceScanSensor.js
wzjjack: I thought about this way, but the policy may be async, we need to control them by different level.

wzjjack: e.g. 
if the device1 port scan is disabled, but tag1 port scan is enabled. will clear device1 scan timer and create a scan job for tag1, tag1 may include device1. when box restart, we may not confirm which policy will be applied first...

drahnieR: go through every policy on each iteration, port scan is a heavier job than reading redis. Is it even possible in the current design to exclude a device from scanning in a network?

sensor/DeviceServiceScanSensor.js
drahnieR: consider using `Config.isFeatureOn()` maybe? this is confusing

drahnieR: consider using `Config.isFeatureOn()` maybe? this is confusing

wzjjack: `Config.isFeatureOn()` is fine, but this sensor inherits `senor`, will call `hookFeature` which may listenning feature on and off, so we apply a new variable to handle this. Just keep the structure currently.

PR #2512, opened 2020-07-24
support per device/group port scan

wzjjack:


sensor/DeviceServiceScanSensor.js
```diff
@@ -55,56 +61,129 @@ class DeviceServiceScanSensor extends Sensor {
...
+          policy === null // exclude
+          policy === false | undefined // unset
+        */
+        if (!setting) return this.scanSettings['0.0.0.0'];
+        if (setting.devicePolicy === true) return true;
+        if (setting.devicePolicy !== null && setting.tagPolicy === true) return true;
+        if (setting.devicePolicy !== null && setting.networkPolicy === true) return true;
+        if (setting.devicePolicy !== null && setting.tagPolicy !== null &&
+          setting.networkPolicy !== null && this.scanSettings['0.0.0.0']) return true;
+        return false;
```
drahnieR, line 164:
```
settings.devicePolicy === true && return true
settings.devicePolicy === null && return false
settings.tagPolicy === true && return true
settings.tagPolicy === null && return false
settings.networkPolicy === true && return true
settings.networkPolicy === null && return false
this.scanSettings['0.0.0.0'] === true && return true
this.scanSettings['0.0.0.0'] === null && return false
```

sensor/DeviceServiceScanSensor.js
```diff
@@ -55,56 +61,129 @@ class DeviceServiceScanSensor extends Sensor {
...
+    try {
+      if (ip === '0.0.0.0') {
+        this.scanSettings[ip] = policy;
+      } else {
+        if (!host)
+          return;
+        switch (host.constructor.name) {
+          case "Tag": {
+            const tagUid = host.o && host.o.uid;
+            if (tagUid) {
```
drahnieR, line 84: we have a similar function to do this `HostManager.getTagMacs()`
implement something like `HostManager.getTagHosts()` instead of duplicate code

sensor/DeviceServiceScanSensor.js
```diff
@@ -55,56 +61,129 @@ class DeviceServiceScanSensor extends Sensor {
...
+                !this.scanSettings[h.o.mac] && (this.scanSettings[h.o.mac] = {})
+                this.scanSettings[h.o.mac].tagPolicy = policy;
+              })
+            }
+            break;
+          }
+          case "NetworkProfile": {
+            const uuid = host.o && host.o.uuid;
+            if (uuid) {
+              const hosts = await hostManager.getHostsAsync();
```
drahnieR, line 101: same above, check `HostManager.getIntfMacs()`

net2/HostManager.js
```diff
@@ -1710,17 +1710,9 @@ module.exports = class HostManager {
...
-          const tags = JSON.parse(host.tags);
-          return tags.includes(String(tag)) || tags.includes(Number(tag));
-        } catch (e) {
-          log.warn('Parse host tags error', host.mac)
-          return false;
-        }
-      })
+    tag = Number(tag);
+    let macs =  this.hosts.all.map(host => host.o)
+      .filter(host => !_.isEmpty(host.tags) && JSON.parse(host.tags).includes(tag))
```
drahnieR, line 1715: is `JSON.parse()` necessary here? `this.hosts.all` should all be `Host` objects

wzjjack: Always use `host.policy.tags` as host tags.

PR #2552, opened 2020-08-04
Enhancement: Dedup activity duration

wzjjack:


wzjjack:
before:
image
after:
image

PR #2655, opened 2020-08-30
Support system profiile tunning

jimyang2008:


scripts/main-start
```diff
@@ -339,12 +339,12 @@ if [[ ! -f /.dockerenv ]]; then
     sudo cp /home/pi/firewalla/etc/bitbridge6.service /etc/systemd/system/.
     sudo cp /home/pi/firewalla/extension/wifi/firewifi@.service /etc/systemd/system/.
     sudo systemctl daemon-reload
-    sudo systemctl enable firehb
+    ${NEED_FIREHB:=true} && sudo systemctl enable firehb
```
MelvinTo, line 342: need to stop firehb if NEED_FIREHB is false

PR #2659, opened 2020-08-31
fix brofish-ping for gold

drahnieR:


scripts/brofish-ping.sh
```diff
@@ -4,6 +4,8 @@
 # In case bro hangs, need to restart it.
 # -----------------------------------------
 
+shopt -s lastpipe
```
jimyang2008, line 7: Not critical.
This `shopt -s lastpipe` might not be necessary if it was ONLY intended for below change of piped commands, since the last command is a "while" block, which handles the variable all inside the same process

scripts/brofish-ping.sh
```diff
@@ -27,14 +29,15 @@ brofish_ping() {
...
-  RESULT=$(top -bn1 -p$(cat /blog/current/.pid) |grep $BRO_PROC_NAME|awk '{print $9}')
-
-  if [[ ${RESULT%%.*} -ge $CPU_THRESHOLD ]]; then
-    echo ${RESULT%%.*}
-    return 1
-  else
-    return 0
-  fi
+  top -bn1|grep $BRO_PROC_NAME|awk '{print $1 " " $9}'|
+  while read PID CPU; do
```
jimyang2008, line 33: This "while" block sets the variables(PID CPU) with values read from pipe and processes them all inside its code block. So above "shopt -s lastpipe" is not needed

jimyang2008 (approved):

scripts/brofish-ping.sh
```diff
@@ -55,11 +58,13 @@ done
...
 
-  /home/pi/firewalla/scripts/firelog -t cloud -m "brofish ping failed, cpu $cpu, restart brofish now"
+  if [[ $cpu == 'good' ]]; then exit; fi
+
+  /home/pi/firewalla/scripts/firelog -t cloud -m "brofish ping failed, restart brofish now, CPU $cpu"
 
   cd $FIREWALLA_HOME
   $FIREWALLA_HOME/bin/node scripts/diag_log.js \
-    --data '{ "msg": "brofish-ping failed", "broCPU": '"$cpu"' }'
+    --data '{ "msg": "brofish-ping failed", "broCPU": '${cpu%% *}' }'
```
jimyang2008, line 67: If we don't need first part, why do we have to echo it in `brofish_cpu()` line 35 ?

scripts/brofish-ping.sh
drahnieR: removing `shopt -s lastpipe` will stop this script from working, probably because of the `return` statement below?

scripts/brofish-ping.sh
drahnieR: for the logging here
https://github.com/firewalla/firewalla/pull/2659/files#diff-d63ae60bfb7eb491d5b288286b10e788R63

jimyang2008: but the logging part has `${cpu%% *}` in it, which is only the first part (value of CPU inside function), so the echo statement inside the function could be simply `echo $CPU` and then here we just need original line without change(use ${cpu} instead of ${cpu%% *}.

scripts/brofish-ping.sh
jimyang2008: This is strange and I am curious about the error, but it might take more time than necessary to fix it. Just go ahead with `shopt -s lastpipe` then.

jimyang2008 (approved):

PR #2661, opened 2020-09-01
CPU Profiling

wzjjack:


controllers/netbot.js
```diff
@@ -1222,6 +1223,21 @@ class netBot extends ControllerBot {
         });
         break;
       }
+      case "cpuProfile":{
+        (async () => {
+          const { applyProfileName, profiles} = value;
```
jimyang2008, line 1228: We might need some comments here to describe input data structure for future reference

jimyang2008 (approved):

PR #2671, opened 2020-09-02
add temperature monitoring feature to heartbeat.js

jimyang2008:


cli/heartbeat.js
```diff
@@ -211,10 +212,96 @@ async function update(status, extra) {
   return info;
 }
 
+async function getOverheatedThresholds() {
+  const uname = await getShellOutput("uname -m");
+  let temperatureThreshold = null;
+  let timeThreshold = null;
+  switch (uname) {
+    case "aarch64": {
+      const boardName = await getShellOutput("awk -F= '/BOARD=/ {print $2}' /etc/firewalla-release");
```
MelvinTo, line 221: cache the local threshold per process life time, no need to query per interval

jimyang2008: Correct me if I am wrong about `setInterval` call. 

The thresholds are calculated inside `getOverheatedThresholds()`, which is only invoked once inside `setInterval()` call. And later run of `monitorTemperature` will just use the same passed-in value, I suppose.

jimyang2008: Go with the way of global variable

cli/heartbeat.js
```diff
@@ -211,10 +218,97 @@ async function update(status, extra) {
...
+  }
+  return ;
+}
+
 const job = setInterval(() => {
   update("schedule");
 }, 24 * 3600 * 1000); // every day
 
+const jobTemperature = setInterval(async () => {
+  await monitorTemperature(false);
```
MelvinTo, line 309: remove false here

cli/heartbeat.js
```diff
@@ -211,10 +218,97 @@ async function update(status, extra) {
...
+}, 3 * 1000); // every 30 seconds
```
MelvinTo, line 310: 3 => 30

cli/heartbeat.js
```diff
@@ -52,6 +53,12 @@ const launchTime = Math.floor(new Date() / 1000);
 
 let uid = null;
 
+let sysStateCount = { "normal": 0, "overheated": 0 };
+let overheatedThresholds = null;
+(async function() {
```
MelvinTo, line 58: minor: run setInterval at the end of this function instead of main-loop

jimyang2008: Fixed as suggested

cli/heartbeat.js
jimyang2008: fixed

cli/heartbeat.js
jimyang2008: fixed

PR #2677, opened 2020-09-03
Enhancement: Box version updated time

wzjjack:
https://github.com/firewalla/firewalla.ios/issues/3004

### DIFF
--- a/net2/SysManager.js
+++ b/net2/SysManager.js
@@ -23,7 +23,7 @@ const fs = require('fs')
 const iptool = require('ip');
 var instance = null;
 const license = require('../util/license.js');
-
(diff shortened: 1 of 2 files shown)


net2/SysManager.js
```diff
@@ -988,6 +988,17 @@ class SysManager {
     }
     return false;
   }
+  async getBranchUpdateTime(branch) {
+    try {
+      await exec(`git config remote.origin.fetch "+refs/heads/${branch}:refs/remotes/origin/${branch}"`);
```
MelvinTo, line 993: no need to fetch change to local, it may take long time
just query the last update time remotely.

PR #2692, opened 2020-09-04
Bugfix: fix exception bug

wzjjack:


alarm/Exception.js
```diff
@@ -177,12 +187,12 @@ module.exports = class {
           }
         }
         let valArray = val, val2Array = val2;
-        this.isJsonString(val) && (valArray = JSON.parse(val));
-        this.isJsonString(val2) && (val2Array = JSON.parse(val2));
+        isJsonString(val) && (valArray = JSON.parse(val));
+        isJsonString(val2) && (val2Array = JSON.parse(val2));
         if (key && key.startsWith("p.tag.ids")) {
           valArray = valArray.map(Number);
```
drahnieR, line 193: this is no longer needed as it's already number at exception creation
We should make sure it's number in alarm as well.

or, actually, what's the point to convert it to number. maybe it's not necessary at all?

alarm/Exception.js
```diff
@@ -177,12 +187,12 @@ module.exports = class {
           }
         }
         let valArray = val, val2Array = val2;
-        this.isJsonString(val) && (valArray = JSON.parse(val));
-        this.isJsonString(val2) && (val2Array = JSON.parse(val2));
+        isJsonString(val) && (valArray = JSON.parse(val));
```
drahnieR, line 190: the parsing should be happening at object creation

drahnieR (changes requested):

drahnieR (approved):

PR #2742, opened 2020-09-15
enable DockerPlugin sensor

jimyang2008:


jimyang2008:
Tested at http://192.168.69.1:8834/v1/docs/#!/System/post_encipher_simple

PR #2822, opened 2020-10-21
Bug: fix bug on aggregated stats when 3-day network flows is introduced

wzjjack:


controllers/netbot.js
```diff
@@ -2257,13 +2257,25 @@ class netBot extends ControllerBot {
...
+        */
+      const requiredPromises = [
         this.hostManager.yesterdayStatsForInit(jsonobj, target),
         this.hostManager.last60MinStatsForInit(jsonobj, target),
         this.hostManager.last30daysStatsForInit(jsonobj, target),
         this.hostManager.newLast24StatsForInit(jsonobj, target),
         this.hostManager.last12MonthsStatsForInit(jsonobj, target)
-      ])
+      ];
+      if (['gold', 'navy'].includes(platform.getName())) {
```
MelvinTo, line 2275: 1. use platform way to handle platform-specific features. do not include platform names here as if-else statement. let platform code to decide what specific interval of stats should be included.
2. better to remove hard code time range, use generic key `this.hostManager.stats` with a map like this
{ "30d": {flows...}}
3. keep all existing 24/30/60 unchanged

drahnieR (approved):

PR #2918, opened 2020-11-19
use retry-timeout to handle big time shift caused by NTP sync failure

jimyang2008:


jimyang2008:
Adjust time to be off a bit(like 600 seconds) and let NTP fix it automatically, see if it can be sync'd without killing the box

PR #2920, opened 2020-11-19
Able to pause an active rule for a while and auto enable after that

wzjjack:


wzjjack:
https://github.com/firewalla/firecommit/issues/96

jasonlyc (approved):

PR #2951, opened 2020-11-27
add events support in firewalla

jimyang2008:


sensor/EventSensor.js
```diff
@@ -0,0 +1,106 @@
+/*    Copyright 2018 Firewalla LLC
```
MelvinTo, line 1: need to update copyrights for every file: `Copyright 2020 Firewalla INC.`

jimyang2008: updated

PR #2986, opened 2020-12-11
disable single sensor using config

drahnieR:


drahnieR:
include this in the next patch to ease feature control

PR #2993, opened 2020-12-14
Ability to collect the ethernet port throughput

jimyang2008:
https://github.com/firewalla/firecommit/issues/132

### DIFF
--- a/net2/HostManager.js
+++ b/net2/HostManager.js
@@ -729,6 +729,7 @@ module.exports = class HostManager {
       this.getCloudURL(json),
       this.networkConfig(json, true),
       this.networkProfilesForInit(json),
+      this.networkMetrics(json),
     ]
 
     await this.basicDataForInit(json, {});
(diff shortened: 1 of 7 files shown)


jasonlyc (approved):

net2/HostManager.js
```diff
@@ -950,6 +951,19 @@ module.exports = class HostManager {
     json.networkProfiles = await NetworkProfileManager.toJson();
   }
 
+  async networkMetrics(json) {
+    const nmkeys = await rclient.keysAsync("metric:throughput:stat:*");
```
MelvinTo, line 955: keysAsync will be expensive and it will block other redis calls.

jimyang2008: Fix this and all found occurrance of same logic(get physical ethernet interfaces) with following line

```
ls -l /sys/class/net | awk '/^l/ && !/virtual/ {print $9}'
```

jasonlyc (approved):

PR #3051, opened 2020-12-29
support hash dns

cli8910:


extension/dnsmasq/dnsmasq.js
```diff
@@ -762,14 +766,18 @@ module.exports = class DNSMASQ {
...
     this.workingInProgress = true;
-    domains = domains.map(d => formulateHostname(d)).filter(Boolean).filter(d => isDomainValid(d)).filter((v, i, a) => a.indexOf(v) === i).sort();
+    const hashDomains = domains.filter(d=>this.isHashDomain(d));
+    domains = domains.filter(d=>!this.isHashDomain(d)).map(d => formulateHostname(d)).filter(Boolean).filter(d => isDomainValid(d)).filter((v, i, a) => a.indexOf(v) === i).sort();
     for (const domain of domains) {
       blockEntries.push(`address=/${domain}/${BLACK_HOLE_IP}$${category}_block`);
       allowEntries.push(`server=/${domain}/#$${category}_allow`);
     }
+    for (const domain of hashDomains) {
+      blockEntries.push(`hash-address=/${domain}/${BLACK_HOLE_IP}$${category}_block`);
```
jasonlyc, line 776: should allowEntries also contain these hash entries?

jasonlyc, line 776: should "allowEntries" also include hash-address entries?

jasonlyc (changes requested):

extension/dnsmasq/dnsmasq.js
jasonlyc: dnsmasq currently does not support allow on hash-address. Since hash-address may not be used in allow rules currently, just keep it as it is and see if there is any need to support it in future.

jasonlyc (approved):

PR #3082, opened 2021-01-08
by default, only the ads list in "on"

cli8910:


sensor/AdblockPlugin.js
```diff
@@ -89,27 +91,33 @@ class AdblockPlugin extends Sensor {
...
-          const result = {};
-          log.info(`Load config list from bone: ${configlistKey}`);
-          const data = await bone.hashsetAsync(configlistKey);
-          const arr = JSON.parse(data);
-          if (Array.isArray(arr)) {
-            for (var i=0; i<arr.length; i++) {
-              result[arr[i]] = "on";
-            }
+        const result = {};
+        this.currentOnList.length = 0;
```
MelvinTo, line 96: is this to empty the array? if so, use `this.currentOnList = []`.

sensor/AdblockPlugin.js
```diff
@@ -209,6 +217,8 @@ class AdblockPlugin extends Sensor {
 
     async updateFilter() {
       const config = await this.getAdblockConfig();
+      const diffList = this.lastOnList.filter(x => !this.currentOnList.includes(x))
```
MelvinTo, line 220: it's not lastOnList vs. currentOnList. It's user config vs. cloud default config

1. if there is any user config, use user config
2. if any list in the user config is not available on the cloud any more, remove this list from effective list
3. if there is no user config, use cloud default config

sensor/AdblockPlugin.js
```diff
@@ -89,27 +91,40 @@ class AdblockPlugin extends Sensor {
...
-          const data = await bone.hashsetAsync(configlistKey);
-          const arr = JSON.parse(data);
-          if (Array.isArray(arr)) {
-            for (var i=0; i<arr.length; i++) {
-              result[arr[i]] = "on";
+        log.info(`Load config list from bone: ${configlistKey}`);
+        const data = await bone.hashsetAsync(configlistKey);
+        //const data = "{\"ads\": {\"default\": \"true\"}, \"ads-adv\":{}}";
+        const adlist = JSON.parse(data);
+        if (!platform.getName() == 'gold') {
```
MelvinTo, line 100: use platform feature check code instead of if statement, so that it can support future platforms.

Example:
https://github.com/firewalla/firewalla/blob/master/platform/gold/GoldPlatform.js#L216

sensor/AdblockPlugin.js
```diff
@@ -41,12 +41,14 @@ const rclient = require('../util/redis_manager.js').getRedisClient();
...
 const util = require('util');
+const platformLoader = require('../platform/PlatformLoader.js');
+const platform = platformLoader.getPlatform();
 
 const fc = require('../net2/config.js');
 
 const featureName = "adblock";
 const policyKeyName = "adblock";
-const configKey = "ext.adblock.config";
+const configKey = "ext.adblock.conf";
```
MelvinTo, line 51: you can use hash `sys:features:config` for feature configs

sensor/AdblockPlugin.js
```diff
@@ -89,27 +91,36 @@ class AdblockPlugin extends Sensor {
     }
 
     async getAdblockConfig() {
-      const json = await rclient.getAsync(configKey);
+      const result = {};
       try {
-        if (json == null) {
-          const result = {};
+        if (!platform.isAdblockCustomizedSupported()) {
+          result["ads"] = "off"
```
MelvinTo, line 97: should be on

cli8910: forget to change  it back after doing some test

sensor/AdblockPlugin.js
cli8910: has changed to sys:features:config and fix a bug about overriding the redis data, for example, when call set API the body is {"ads-adv":"on"}, and the redis data is {"ads":"on"}, the origin logic will cause the redis data is {"ads-adv":"on"} however the correct redis data should be {"ads":"on", "ads-adv":"on"}

jasonlyc (approved):

sensor/AdblockPlugin.js
```diff
@@ -78,7 +80,16 @@ class AdblockPlugin extends Sensor {
 
     async apiRun() {
       extensionManager.onSet("adblockConfig", async (msg, data) => {
-        await rclient.setAsync(configKey, JSON.stringify(data));
+        const originConfig = await rclient.hgetAsync(configKey, "adblock");
+        if (originConfig != null) {
+          const originConfigObj = JSON.parse(originConfig);
+          for (const field in data) {
```
MelvinTo, line 86: this means it will never be able to override any fields that are NOT passed by the app. Why not just simply overriding the redis key with data from app?

sensor/AdblockPlugin.js
```diff
@@ -89,27 +100,36 @@ class AdblockPlugin extends Sensor {
...
           log.info(`Load config list from bone: ${configlistKey}`);
           const data = await bone.hashsetAsync(configlistKey);
-          const arr = JSON.parse(data);
-          if (Array.isArray(arr)) {
-            for (var i=0; i<arr.length; i++) {
-              result[arr[i]] = "on";
+          // const data = "{\"ads\": {\"default\": true}, \"ads-adv\":{}}";
+          const adlist = JSON.parse(data);
+          // from redis
+          const configStr = await rclient.hgetAsync(configKey, "adblock");
```
MelvinTo, line 113: use the base function this.getFeatureConfig from Sensor.js

sensor/AdblockPlugin.js
```diff
@@ -78,7 +80,16 @@ class AdblockPlugin extends Sensor {
 
     async apiRun() {
       extensionManager.onSet("adblockConfig", async (msg, data) => {
```
MelvinTo, line 82: Need onGet function too for app to get current configuration. (default+redis configured)

MelvinTo: * what's configured by default in the config.json
* what's the runtime config that is configured by app in the redis

sensor/AdblockPlugin.js
```diff
@@ -269,17 +292,21 @@ class AdblockPlugin extends Sensor {
...
-        const file = `${dnsmasqConfigFolder}/${key}_adblock.conf`;
-        try {
-          if (fs.existsSync(file)) {
-            await fs.unlinkAsync(file);
+    _cleanUpFilter(config) {
+      try {
+        const result = []
+        if (typeof config == 'object') {
+          for (const key in config) {
+            if(config[key] == "on") result.push(key+"_adblock.conf")
```
MelvinTo, line 300: how to make sure any leftover configs are removed as well?

1. enabled list ad1 and ad2
2. ad1 is removed from cloud
3. when next cleanup filter, will ad1 list be removed from the dnsmasq config directory?

sensor/AdblockPlugin.js
MelvinTo: so that the list of adblock lists can be displayed in the app

sensor/AdblockPlugin.js
```diff
@@ -269,17 +292,21 @@ class AdblockPlugin extends Sensor {
...
           }
-        } catch (err) {
-          log.error(`Failed to delete file: '${file}'`, err);
         }
+        fs.readdirSync(dnsmasqConfigFolder).forEach(file => {
+          if (file.endsWith('_adblock.conf') && !result.includes(file)) {
```
jasonlyc, line 304: There are other group level adblock config file also ends with _adblock.conf. Need to use another way to pick the adblock domain list file. E.g., use a different suffix for domain list file.

jasonlyc (approved):

sensor/AdblockPlugin.js
jasonlyc, line 304: There are other group-level adblock config files ending with _adblock.conf. Need to use another way to pick the adblock domain list file. E.g., use a different suffix for domain list file.

jasonlyc (changes requested):

sensor/AdblockPlugin.js
MelvinTo: ignore the above comments for now

MelvinTo (changes requested):
x

MelvinTo (dismissed):

jasonlyc (dismissed):

MelvinTo (approved):

PR #3091, opened 2021-01-12
Use pipe for acl logs and buffers before write to redis

drahnieR:


scripts/audit-run
```diff
@@ -3,29 +3,26 @@
...
 # sudo ln -s /alog/acl-audit.log /var/log/acl-audit.log
 sudo mkdir -p /log/alog
-sudo ln -s /alog/acl-audit.log /log/alog/acl-audit.log
+sudo ln -sf /alog/acl-audit.log /log/alog/acl-audit.log
+
+
+sudo cp ${FIREWALLA_HOME}/etc/rsyslog.d/30-acl-audit.conf /etc/rsyslog.d/
+sudo systemctl restart rsyslog
+
+sudo iptables -C FW_DROP -j LOG --log-prefix "[FW_ACL_AUDIT]" &>/dev/null || sudo iptables -I FW_DROP -j LOG --log-prefix "[FW_ACL_AUDIT]"
```
MelvinTo, line 25: better to move this iptables commands to control/install_iptables_setup.sh

MelvinTo: also the processing of iptables drop logs may slow down red/blue, need a platform on/off func so that we can easily enable/disable in the future

MelvinTo (approved):

PR #3120, opened 2021-01-22
group global block set, more accurate security policy definition

drahnieR:


Treefogg (commented):
Not sure why they got changed with out my permition

PR #3128, opened 2021-01-25
adblock support redis query from dnsmasq

cli8910:


sensor/AdblockPlugin.js
```diff
@@ -342,7 +359,20 @@ class AdblockPlugin extends Sensor {
       this.reloadFilterImmediate = setImmediate(this._reloadFilter.bind(this));
     }
 
+    async getFastMode() {
+      let fastMode = true;
+      try {
+        const fastModeStr = await rclient.getAsync("adblock.fastmode");
```
MelvinTo, line 365: do not use a new redis key, using "policy:system" is preferred.

MelvinTo: so that it can be automatically manageable by apps

sensor/AdblockPlugin.js
```diff
@@ -275,14 +289,31 @@ class AdblockPlugin extends Sensor {
...
     }
 
+    preprocess(key, hashes) {
+      rclient.del(key);
+      const newHashes = [];
+      hashes.forEach((hash) => {
+        newHashes.push(hash.replace(/\//g, '.'));
+      });
+      const cmd = [key];
+      cmd.push.apply(cmd, newHashes);
```
jasonlyc, line 313: Since there may be 10K+ entries, use concat instead of Array.prototype.push.apply to avoid call stack size exceeded
https://github.com/nodejs/node/issues/27732#issuecomment-493149690

sensor/AdblockPlugin.js
```diff
@@ -89,6 +89,20 @@ class AdblockPlugin extends Sensor {
       extensionManager.onGet("adblockConfig", async (msg, data) => {
         return this.getAdblockConfig();
       });
+      extensionManager.onSet("adblockFastMode", async (msg, data) => {
+        await rclient.hsetAsync("policy:system", "adblock_fastmode", JSON.stringify(data));
```
jasonlyc, line 93: use a constant variable for "adblock_fastmode" to avoid duplicate literal string scattered around

jasonlyc (changes requested):

sensor/AdblockPlugin.js
```diff
@@ -89,6 +90,20 @@ class AdblockPlugin extends Sensor {
       extensionManager.onGet("adblockConfig", async (msg, data) => {
         return this.getAdblockConfig();
       });
+      extensionManager.onSet("adblockFastMode", async (msg, data) => {
```
MelvinTo, line 93: this api set is not necessary. policy:system already supports configuration load/change from app natively.

Refer to this example for api call to refresh adblock config, no need to send event ADBLOCK_CONFIG_REFRESH, directly call the apply change function
https://github.com/firewalla/firewalla/blob/master/sensor/FamilyProtectPlugin.js#L56

sensor/AdblockPlugin.js
```diff
@@ -254,7 +247,7 @@ class AdblockPlugin extends Sensor {
         }
         try {
           if (arr.length > 0) {
-            await this.writeToFile(arr, configFilePath + ".tmp");
+            await this.writeToFile(key, arr, configFilePath + ".tmp", this.fastMode);
```
jasonlyc, line 250: better to add a prefix for adblock in redis key

jasonlyc (changes requested):

jasonlyc (approved):

PR #3138, opened 2021-01-28
verify the domain is or is not hashed

cli8910:


util/util.js
```diff
@@ -160,6 +161,12 @@ function generateStrictDateTs(ts) {
   }
 }
 
+function isHashDomain(domain) {
```
jasonlyc, line 164: H3dF2wLoc6Br1d89gtM2vV6DK215zY.XUB8JoptqZZ8=
Will the hash pass the test if it contains "."?

jasonlyc (changes requested):

util/util.js
jasonlyc: Also make sure clear text domain will not match this

jasonlyc (changes requested):

PR #3148, opened 2021-02-01
support customized doh server

wzjjack:


extension/dnscrypt/dnscrypt.js
```diff
@@ -216,12 +182,24 @@ class DNSCrypt {
...
   async setAllServers(servers) {
-    if(servers === null) {
+    if (servers === null) {
       return rclient.delAsync(allServerKey);
     }
 
     return rclient.setAsync(allServerKey, JSON.stringify(servers));
   }
+  async getCustomizeServers() {
+    // string | object {name:'nextdns',stamp:'xyz'}
```
drahnieR, line 192: use separate key here and keep the structure consistent within single key

PR #3155, opened 2021-02-01
verify the domain is hashed or not 

cli8910:


util/util.js
```diff
@@ -17,6 +17,7 @@
 const _ = require('lodash')
 
 const validDomainRegex = /^[a-zA-Z0-9-_.]+$/
+const hashDomainRegex = /^([A-Za-z0-9+/.]{4})*([A-Za-z0-9+/.]{3}=)?$/;
```
jasonlyc, line 20: zconfig.alibabausercontent.com.w.cdngslb.com
This domain will also match hash domain regular expression

jasonlyc (changes requested):

util/util.js
jasonlyc: mapdownload.autonavi.com.danuoyi.tbcache.com

jasonlyc (approved):

PR #3161, opened 2021-02-03
support  new category

cli8910:


control/DomainBlock.js
```diff
@@ -283,7 +285,20 @@ class DomainBlock {
 
   // this function updates category domain mappings in dnsmasq configurations
   async updateCategoryBlock(category) {
-    const domains = await this.getCategoryDomains(category);
+    const CategoryUpdater = require('./CategoryUpdater.js');
```
jasonlyc, line 288: It's better to do these things in CategoryUpdateSensor.js in the "Policy:CategoryActivated" handler.

jasonlyc (changes requested):

PR #3170, opened 2021-02-04
support  new category

cli8910:


sensor/CategoryUpdateSensor.js
```diff
@@ -206,6 +213,14 @@ class CategoryUpdateSensor extends Sensor {
 
       sem.on('Policy:CategoryActivated', async (event) => {
         const category = event.category;
+        const domains = await categoryUpdater.getDefaultDomains(category);
```
jasonlyc, line 216:
```
if (!categoryUpdater._isCustomizedCategory(category)) {
  await this.updateCategory(category)
}
```
then you don't need to modify updateCategory function. It's better to rename the _isCustomizedCategory function to isCustomizedCategory since it is used outside the CategoryUpdater.js

```

jasonlyc (approved):

sensor/CategoryUpdateSensor.js
jasonlyc: Assume there are only customized category and categories from cloud

jasonlyc (changes requested):

jasonlyc (approved):

PR #3174, opened 2021-02-04
add dns total count

drahnieR:


sensor/ACLAuditLogPlugin.js
```diff
@@ -119,6 +124,29 @@ class ACLAuditLogPlugin extends Sensor {
     }
   }
 
+  async getIntfMac(localIP, record) {
+    const intf = new Address4(localIP).isValid() ?
+      sysManager.getInterfaceViaIP4(localIP) :
+      sysManager.getInterfaceViaIP6(localIP)
+
+    if (!intf) throw new Error('Interface not found for', localIP);
```
MelvinTo, line 132: is the error handled for code calling this function?

PR #3184, opened 2021-02-07
throw error when switch branch to null

wzjjack:


controllers/netbot.js
```diff
@@ -4113,7 +4113,9 @@ class netBot extends ControllerBot {
         targetBranch = prodBranch
         break
     }
-
+    if (!targetBranch) {
```
drahnieR, line 4116: put it in default case?

controllers/netbot.js
```diff
@@ -4113,7 +4113,9 @@ class netBot extends ControllerBot {
...
+      throw new Error("Can not switch to branch", targetBranch);
```
drahnieR, line 4117: log `target` instead of `targetBranch`, which is null in this case.

drahnieR (changes requested):

drahnieR (approved):

PR #3213, opened 2021-02-21
Box api for live stats

wzjjack:


api/routes/fastencipher2.js
```diff
@@ -30,39 +30,85 @@ const sc = require('../lib/SystemCheck.js');
...
-    encryption.decrypt,
-    sc.debugInfo,
 
-    (req, res, next) => {
-      const gid = req.params.gid;
+const msgHandler = (req, res, next) => {
+  const gid = req.params.gid;
+  const streaming = (req.body.message && req.body.message.obj && req.body.message.obj.streaming) || false;
+  (async () => {
+    if (streaming) {
```
MelvinTo, line 38: use a separate function to handle GET+event, do not change the existing POST function.

keep logic separated.

PR #3390, opened 2021-04-01
control enable/disable of feature from cloud

cli8910:


net2/config.js
```diff
@@ -301,7 +349,11 @@ setInterval(() => {
   syncDynamicFeaturesConfigs()
 }, 60 * 1000) // every minute
 
+syncCloudFeaturesConfigs()
 
+setInterval(() => {
+  syncCloudFeaturesConfigs()
+}, 5 * 60 * 1000)
```
MelvinTo, line 356: no need to run this very frequently, it's okay to run only when firemain/fireapi starts. because if it changes very fast at runtime, it may cause internal data out of sync.

net2/config.js
```diff
@@ -210,6 +226,33 @@ async function syncDynamicFeaturesConfigs() {
...
+  if(configServerUrl) {
+    const options = {
+      uri: configServerUrl,
+      family: 4,
+      method: 'GET',
+      maxAttempts: 5,
+      retryDelay: 1000,
+      json: true
+    };
+    request(options, (err, httpResponse, body) => {
```
MelvinTo, line 244: better to use async/await here to ensure data is loaded

net2/config.js
```diff
@@ -41,6 +42,28 @@ const readFileAsync = util.promisify(fs.readFile);
 
 const platform = require('../platform/PlatformLoader').getPlatform()
 
+const { rrWithErrHandling } = require('../util/requestWrapper.js')
+
+async function initCloudConfig() {
+  let configServerUrl = null;
+  if (f.isDevelopmentVersion()) configServerUrl = 'https://s3-us-west-2.amazonaws.com/fireapp/box_dev.json'
```
MelvinTo, line 49: use config.json to store the urls.

jasonlyc (approved):

PR #3391, opened 2021-04-01
store tech support file during reset

jimyang2008:


scripts/system-reset-all-overlayfs.sh
```diff
@@ -23,6 +23,9 @@ sudo rm -rf ${FIREWALLA_UPPER_DIR}.bak ${FIREWALLA_UPPER_WORK_DIR}.bak
 sudo mv ${FIREWALLA_UPPER_DIR}{,.bak}
 sudo mv ${FIREWALLA_UPPER_WORK_DIR}{,.bak}
 
+# store tech support file
+/home/pi/firewalla/scripts/store_support.sh
```
MelvinTo, line 27: make this an optional argument when calling reset function, also need to change the nodejs code to support that.

if option is enabled, run store support.sh
if not, do not run the script and cleanup any support files stored previously

jimyang2008: Change updated as requested

PR #3416, opened 2021-04-13
add event & log when socketio doesn't connect at all

drahnieR:


encipher/lib/encipherio.js
```diff
@@ -677,26 +678,26 @@ let legoEptCloud = class {
...
-          'mtype': mtype,
-          'fid': fid,
-          'mid': mid,
-        },
-        maxAttempts: 5,   // (default) try 5 times
-        retryDelay: 1000,  // (default) wait for 1s before trying again
-      };
-
       if (!this.disconnectCloud) {
+        const options = {
```
MelvinTo, line 682: socket io and this send function don't depend on each other.

MelvinTo: should continue sending requests even if socket io is down

drahnieR: https://github.com/firewalla/firewalla/pull/3416/files#diff-936f2142c83117982c48a92ed45004eb08d396ae47c5ab998f895e7a4a953006L734

It's saved to redis as a queue when disconnected

Treefogg (approved):

PR #3487, opened 2021-04-30
do not reflect internet block on DNS queries

jasonlyc:


jasonlyc:
Otherwise, blocking internet on a network will cause DNS self-check fail

jasonlyc:
https://github.com/firewalla/firecommit/issues/521

MelvinTo (approved):

PR #3604, opened 2021-06-02
fix legacy query, fix auditlog without port info

drahnieR:


drahnieR:
this fixes app aggregation not showing correctly now on master

MelvinTo (approved):

PR #3649, opened 2021-06-17
re-write metric:throughput data collection and processing in NetworkS…

jimyang2008:
…tatsSensor.js to avoid multiple long lasting bash processes


sensor/NetworkStatsSensor.js
```diff
@@ -48,28 +53,137 @@ class NetworkStatsSensor extends Sensor {
     this.processPingConfigure()
     this.pingResults = {}
     this.checkNetworkPings = {}
+    this.sampleJobs = {}
+    this.processJobs = {}
+  }
+
+  sleep(seconds) {
```
MelvinTo, line 60: there is always a delay util function, check it out

jimyang2008: fixed with "delay"

MelvinTo (approved):

PR #3679, opened 2021-06-25
always return hosts with port forward in init data

jasonlyc:


MelvinTo (approved):

jasonlyc:
https://github.com/firewalla/design/issues/160

PR #3887, opened 2021-08-17
disable active probe script for ethernet_state

jimyang2008:


jimyang2008:
This is disabled to avoid noisy events about ethernet disconnection, which would happen if an ethernet port is never connected upon every restart of Firewalla service.

MelvinTo (approved):

PR #3888, opened 2021-08-17
do not compare empty value in Policy.isEqualToPolicy

jasonlyc:


MelvinTo (approved):

Treefogg (commented):
Not sure if this is what I need

PR #3900, opened 2021-08-19
[Bug] monthly data usage may decrease for timezones behind GMT+0 

wzjjack:
https://github.com/firewalla/firecommit/issues/804

### DIFF
--- a/util/TimeSeries.js
+++ b/util/TimeSeries.js
@@ -53,13 +53,19 @@ module.exports = {
   getBoneAPITimeSeries: function () { return boneAPITimeSeries },
   getTimeSeriesWithTz: function () { return timeSeriesWithTz },
   supportTimeSeriesWithTz: async function () {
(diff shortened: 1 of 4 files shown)


util/TimeSeries.js
```diff
@@ -33,7 +33,33 @@ boneAPITimeSeries.granularities = {
...
+  }
+})()
+
 module.exports = {
-  getTimeSeries: function() {return timeSeries},
-  getBoneAPITimeSeries: function() {return boneAPITimeSeries}
+  getTimeSeries: function () { return timeSeries },
+  getBoneAPITimeSeries: function () { return boneAPITimeSeries },
+  getTimeSeriesWithTz: function () { return timeSeriesWithTz },
+  supportTimeSeriesWithTz: async function () {
```
MelvinTo, line 55: keys is very costly, this check if one time only for each fireapi process, should not check for every api call. check once and use sys:config flag to store the check result.

MelvinTo (approved):

PR #4079, opened 2021-11-04
always create/remove FWPRODUCTION under /tmp at the end of switch_bra…

jasonlyc:
…nch.sh


net2/Firewalla.js
```diff
@@ -147,12 +146,6 @@ function isProduction() {
   } else {
     return false
   }
-
```
jasonlyc, line 150: This will never be used after if ... else ... block above

MelvinTo (approved):

PR #4083, opened 2021-11-04
add wan block flows to compression

wzjjack:


sensor/FlowCompressionSensor.js
```diff
@@ -308,6 +358,51 @@ class FlowCompressionSensor extends Sensor {
...
+          completed = true
+        } else {
+          options.ts = flows[flows.length - 1].ts
+        }
+        allFlows = allFlows.concat(flows)
+        if (allFlows.length >= this.maxCount) {
+          // compress and dump flows to redis if it exceed count 
+          // big array of allFlows might cause oom
+          const ts = allFlows[allFlows.length - 1].ts;
+          await this.appendAndSave(ts, await this.compress(allFlows), 'wanBlock')
```
drahnieR, line 390: All data is kept in array and then sent to zlib, creates a big memory consumption. Using pipe should be a better approach. Same problem in build().

drahnieR (changes requested):

drahnieR (approved):

MelvinTo (approved):

PR #4110, opened 2021-11-22
fix heartbeat.zeek capability issue with zeek 3.0

drahnieR:


scripts/brofish-ping.sh
```diff
@@ -18,7 +18,7 @@ MMIN="15"
 
 brofish_hb() {
   # zeek logs get rotated every 3 mins, checking archive folder as well here
-  local RESULT=$(find /log/blog/*/heartbeat.* -mmin -${MMIN} 2>/dev/null)
+  local RESULT=$(find /log/blog -follow -name 'heartbeat.*' -mmin -${MMIN} 2>/dev/null)
```
jimyang2008, line 21: Maybe use `-L` instead of  `-follow`, which is Deprecated..
>        -follow
              Deprecated; use the -L option instead.  Dereference symbolic links.  Implies -noleaf.  The -follow option affects only those tests which appear after it on the command line.  Unless the -H or -L option has been specified, the position of the -follow
              option  changes the behaviour of the -newer predicate; any files listed as the argument of -newer will be dereferenced if they are symbolic links.  The same consideration applies to -newerXY, -anewer and -cnewer.  Similarly, the -type predicate will
              always match against the type of the file that a symbolic link points to rather than the link itself.  Using -follow causes the -lname and -ilname predicates always to return false.

jimyang2008 (approved):

scripts/brofish-ping.sh
drahnieR: manual doesn't match the binary on Gold

MelvinTo (approved):

PR #4296, opened 2022-01-13
Fix a bug that plugin callbacks may get the stale state using isFeatureOn()

yexia0:


net2/config.js
```diff
@@ -242,16 +242,21 @@ function reloadFeatures() {
   }
 
   for (const f in callbacks) {
-    if (!features)
+    if (!features) {
+      features = {}
```
drahnieR, line 246: `features = featuresNew` here might be better, `featuresNew` might not have every entry enabled

drahnieR (approved):

MelvinTo (approved):

PR #4362, opened 2022-01-28
[vpn] do not enable dns redirect on trojan and add a workaround for yaml lib bug

MelvinTo:


MelvinTo:
https://stackoverflow.com/a/22235064/832132

jasonlyc (approved):

PR #4710, opened 2022-05-16
use hgetall and associative array to speed up sanity check

drahnieR:


drahnieR:
on SGold:
Rules: before 1m1.357s, after 0m4.789s
Hosts: before 0m16.826s, after 0m13.216s

jimyang2008 (approved):

MelvinTo (approved):

PR #5270, opened 2023-02-20
use ROUTER_MANAGED for feature check

drahnieR:


scripts/sanity_check.sh
```diff
@@ -1,19 +1,24 @@
...
     PLATFORM='gold'
     ;;
   "aarch64")
     if [[ -e /etc/firewalla-release ]]; then
       PLATFORM=$( . /etc/firewalla-release 2>/dev/null && echo $BOARD || cat /etc/firewalla-release )
+      if [[ $PLATFORM == "blue" ]]; then
+        ROUTER_MANAGED='no'
+      fi
     else
       PLATFORM='unknown'
```
jimyang2008, line 16: for "unknown" platform, maybe need set ROUTER_MANAGED='no' as well ?

jimyang2008 (approved):
All look good with a minor comment for unexpected corner case

scripts/sanity_check.sh
drahnieR: probably all future platforms will have FireRouter, we write less code with this change

jimyang2008: In normal cases, all future platforms would have `/etc/firewalla-release` in place,  which would have the default value of 'yes' for ROUTER_MANAGED as expected.

However, if something unexpected happens to cause file `/etc/firewalla-release` lost, the "else" block would have PLATFORM set to "unknown". Then do we still want to leave **ROUTER_MANAGED** as "yes" in that case?

But it is an exceptional case, where many things would have been broken anyway. So feel free to leave it as is

drahnieR: yes, worst case is broken sanity check anyway. This is just trying to avoid writing code for future platforms.

jimyang2008 (approved):

MelvinTo (approved):

PR #5562, opened 2023-06-12
add bridge mode recognition, add LAN DNS to networks

drahnieR:


scripts/sanity_check.sh
```diff
@@ -642,37 +647,65 @@ check_network() {
     fi
 
     echo "---------------------- Network ------------------"
-    curl localhost:8837/v1/config/interfaces -o /tmp/scc_interfaces &>/dev/null
+    curl localhost:8837/v1/config/interfaces -s -o /tmp/scc_interfaces
```
jimyang2008, line 650: Maybe safer to use a unique file name like`/tmp/scc_interfaces-$$`,  just in case multiple runs of same script at the same time

jimyang2008: It might be safer to have a check on HTTP code of curl command , in case of any HTTP error, which might cause error output in the temporary file

```
http_code=$(curl -s -w '%{http_code}\n' 'localhost:8837/v1/config/interfaces' -o /tmp/scc_interfaces-$$)
if [[ "$http_code" != '200' ]]; then
   echo "ERROR:  HTTP error $http_code" >&2
fi
```

scripts/sanity_check.sh
```diff
@@ -642,37 +647,65 @@ check_network() {
...
+      # echo ${COL[0]}
+      # echo "ip${#IP6[@]} dns${#DNS[@]}"
+      # echo ${DNS_CONFIG["${COL[0]}"]}
+      # echo ${IP6[@]}
+      # echo ${DNS[@]}
+
+      LINE_COUNT=$(( "${#IP6[@]}" > "${#DNS[@]}" ? "${#IP6[@]}" : "${#DNS[@]}" ));
+      for (( IDX=0; IDX < $LINE_COUNT; IDX++ )); do
+        # echo $IDX
+        if [[ ${#IP6[@]} > $IDX ]]; then
```
jimyang2008, line 683: For numeric comparison, use `-gt` instead of `>` ?
```
pi@firewalla:~ (G) $ if [[ 3 > 12 ]]; then echo h; else echo l; fi
h
pi@firewalla:~ (G) $ if [[ 3 -gt 12 ]]; then echo h; else echo l; fi
l
```

scripts/sanity_check.sh
```diff
@@ -642,37 +647,65 @@ check_network() {
...
+          IP="${IP6[$IDX]}";
+        else
+          IP=; fi
+
+        if [[ z"${COL[7]}" == "z" ]]; then
+          if [[ $IDX -eq 0 ]]; then DN="${COL[7]}"; else DN=; fi
+        elif [[ ${#DNS[@]} > $IDX ]]; then
```
jimyang2008, line 690: same here, '-gt' rather than '>'

jimyang2008 (changes requested):

scripts/sanity_check.sh
drahnieR: this API should not return error, otherwise we should fix firerouter.
parallel runs are possible but rarely there's a config change between runs.

jimyang2008 (approved):

MelvinTo (approved):

PR #5589, opened 2023-06-16
cache user agent info in HttpFlow

jasonlyc:


MelvinTo (approved):

extension/flow/HttpFlow.js
```diff
@@ -84,96 +90,118 @@ class HttpFlow {
...
+      result = {};
+      try {
+        const detectResult = this.detector.detect(userAgent);
+        if (detectResult)
+          result.detect = detectResult;
+        const parseResult = useragent.parse(userAgent);
+        if (parseResult)
+          result.parse = parseResult;
+        if (Object.keys(result) > 0)
+          uaInfoCache.set(userAgent, result);
```
drahnieR, line 105: saving result is not necessary here. If match on LRU, just reset expire on redis key should be fine.

PR #5816, opened 2023-07-31
fix a firehb issue that cause wrong output of gateway mac prefix

drahnieR:


jimyang2008 (approved):
should be fine after adding a special case handling for multiple WANs

jimyang2008 (approved):

MelvinTo (approved):

PR #6023, opened 2023-09-28
modularize iptables scripts, create individual ones for fine tuned policy reset

drahnieR:


control/reset_iptables_allow_block.sh
```diff
@@ -0,0 +1,26 @@
...
+
+reset_ip_rules
+
+create_filter_table
+
+sudo iptables-restore "$iptables_file"
+sudo ip6tables-restore "$ip6tables_file"
+
+
+# most policy created, block/allow related set starts with c_b._
```
jasonlyc, line 16: ipset naming convention is same in QoS and PBR rules. This may flush ipset that are used in PBR and QoS to match IP/cidr/domain/category.

control/reset_iptables_allow_block.sh
```diff
@@ -0,0 +1,26 @@
+#!/bin/bash
+
+# shellcheck source=iptables_common.sh
+source "$(dirname "$0")/iptables_common.sh"
+
+reset_ipset
+
+reset_ip_rules
```
jasonlyc, line 8: no need to reset ip rules while resetting allow/block rules

jasonlyc (changes requested):

control/reset_iptables_allow_block.sh
drahnieR: On reseting allow/block, all other rules will be re-enforced. As there're too many things overlapping each other.

jasonlyc (approved):

PR #6730, opened 2024-04-27
Remove unused dependencies from root package.json

parkerbxyz:
This pull request primarily removes unused dependencies from the root package.json file. I used [`depcheck`](https://github.com/depcheck/depcheck) to find the unused dependencies. This should reduce/resolve Dependabot alerts for vulnerable dependencies that are not in use.


MelvinTo (approved):

package.json
```diff
@@ -8,25 +8,17 @@
...
-    "grunt-mocha": "^0.4.13",
-    "grunt-mocha-test": "^0.12.7",
     "mocha": "^2.2.5",
     "node-pty": "^0.9.0"
   },
   "dependencies": {
     "async": "^2.1.1",
     "bee-queue": "^1.2.2",
     "bluebird": "^3.5.5",
-    "body-parser": "1.9.x",
```
drahnieR, line 24: in use

package.json
```diff
@@ -8,25 +8,17 @@
...
     "botbuilder": "^0.6.1",
     "camel-case": "^3.0.0",
     "child-process-promise": "*",
     "commander": "^2.9.0",
-    "cookie-parser": "~1.3.5",
```
drahnieR, line 29: in use

package.json
```diff
@@ -38,24 +30,18 @@
...
     "jsonfile": "^2.4.0",
     "jsonwebtoken": "^8.3.0",
-    "later": "^1.2.0",
     "lodash": "^4.17.15",
     "minimatch": "3.0.4",
     "minimist": "*",
     "mkdirp": "^0.5.1",
-    "mobile-detect": "https://registry.npmjs.org/mobile-detect/-/mobile-detect-1.3.3.tgz",
     "moment": "^2.15.0",
-    "morgan": "^1.9.1",
```
drahnieR, line 48: in use

package.json
```diff
@@ -38,24 +30,18 @@
...
     "mustache-express": "^1.2.5",
     "network": "git+https://git@github.com/melvinto/network.git",
     "node-df": "^0.1.4",
     "node-forge": "^0.7.1",
     "node-persist": "0.0.8",
     "node-ssdp": "^4.0.0",
-    "passport": "*",
```
drahnieR, line 55: in use

package.json
```diff
@@ -64,9 +50,7 @@
     "requestretry": "^4.0.2",
     "socket.io-client": "^2.2.0",
     "stats-lite": "https://registry.npmjs.org/stats-lite/-/stats-lite-2.0.3.tgz",
-    "swagger-node-express": "^2.1.3",
```
drahnieR, line 67: in use

drahnieR (changes requested):
quite a few dependencies are actually in use, i've marked those inline

parkerbxyz:
> quite a few dependencies are actually in use, i've marked those inline

I saw that some of these are in use in subdirectories that have their own `package.json`. Are these in use by scripts that use `package.json` in the root directory?

package.json
parkerbxyz: This doesn't appear to be in use outside the `api` directory, which has its own `package.json` file:

https://github.com/firewalla/firewalla/blob/4550b0756ce72d253290d437c5f7d54a8d213005/api/package.json#L9

package.json
parkerbxyz: This doesn't appear to be in use outside the `api` directory, which has its own `package.json` file:

https://github.com/firewalla/firewalla/blob/4550b0756ce72d253290d437c5f7d54a8d213005/api/package.json#L10

package.json
parkerbxyz: This doesn't appear to be in use outside the `api` directory, which has its own `package.json` file:

https://github.com/firewalla/firewalla/blob/4550b0756ce72d253290d437c5f7d54a8d213005/api/package.json#L15

package.json
parkerbxyz: This doesn't appear to be in use outside the `api` directory, which has its own `package.json` file:

https://github.com/firewalla/firewalla/blob/4550b0756ce72d253290d437c5f7d54a8d213005/api/package.json#L16

package.json
parkerbxyz: This doesn't appear to be in use outside the `api` directory, which has its own `package.json` file:

https://github.com/firewalla/firewalla/blob/4550b0756ce72d253290d437c5f7d54a8d213005/api/package.json#L19

package.json
parkerbxyz: I restored this with the version being used in `api/package.json` since it was newer.

parkerbxyz:
> That's true, but we are not really using `api/package.json` in practice. package.json under root should be including every module across whole project. I think you could remove `api/package.json` in this PR

Got it. I removed unused `package.json` files and restored dependencies that are still in use to the root `package.json` file.

MelvinTo (approved):

drahnieR (approved):

PR #7595, opened 2025-01-09
Parse wlan vendor and enrich host

caixingcheng:


sensor/APCMsgSensor.js
```diff
@@ -340,12 +399,105 @@ class APCMsgSensor extends Sensor {
...
+      const vendorBuff = Buffer.from(firstHexVendor); // used for best match
+      const cmd = `grep -e "^${baseVendor}" ${OUI_FILE_PATH}`;
+      log.info(`Looking up vendor info for ${vendor} with command ${cmd}`);
+      exec(cmd, (err, stdout, stderr) => {
+        if (err) {
+          log.error(`Failed to lookup vendor info for ${vendor}`, err.message);
+          return null;
+        }
+        if (stdout) {
+          log.info("cmd output: " + stdout);
```
MelvinTo, line 435: this will flood log file

sensor/APCMsgSensor.js
```diff
@@ -340,12 +399,105 @@ class APCMsgSensor extends Sensor {
...
+      return null;
+    const hexVendorInfo = vendor.split(' ');
+
+    if (hexVendorInfo.length >= 1) {
+      const firstHexVendor = hexVendorInfo[0].toUpperCase().substring(2); // the first vendor info is the most significant, omit the 0x prefix
+      const baseVendor = firstHexVendor.substring(0, 6);
+      const vendorBuff = Buffer.from(firstHexVendor); // used for best match
+      const cmd = `grep -e "^${baseVendor}" ${OUI_FILE_PATH}`;
+      log.info(`Looking up vendor info for ${vendor} with command ${cmd}`);
+      exec(cmd, (err, stdout, stderr) => {
```
MelvinTo, line 429: we have execAsync to deal with external command better.

sensor/APCMsgSensor.js
caixingcheng: ok, I'll remove it

sensor/APCMsgSensor.js
```diff
@@ -340,12 +399,104 @@ class APCMsgSensor extends Sensor {
...
+            let maxMatch = baseVendor.length;
+            let vendorName = fistMatchLine.substring(index + 1).trim();
+            for (let line of lines) {
+              log.info(`Checking vendor info ${line}`);
+              line = line.trim();
+              if (line.length == 0) {
+                log.info('Skipping empty line');
+                continue;
+              }
+              index = line.indexOf(' ');
```
MelvinTo, line 456: we really don't need to write so many code just to implement this requirement

I use bash as an example
```bash
#!/bin/bash

MAC="$1"

START=6
STRLEN=${#MAC}

while [[ $START -le $STRLEN ]]; do
    PREFIX=${MAC:0:$START}
    RESULT=$(awk "\$1 == \"$PREFIX\" {print \$0}" /usr/share/nmap/nmap-mac-prefixes | head -n 1 | cut -d' ' -f2-)
    if [[ "$RESULT" != "IEEE Registration Authority" ]]; then
        echo $RESULT
        exit 0
    fi
    START=$(( START + 1 ))
done
```

caixingcheng: because I want a more efficient implementation, your example might search the 50000+ lines multi times

jasonlyc (approved):

drahnieR (changes requested):

PR #7608, opened 2025-01-14
fix null reference in DeviceHook

drahnieR:


caixingcheng (approved):
seems good, no comments

jasonlyc (approved):

PR #8686, opened 2026-01-07
create control modules for iptables, ipset, tlsset

drahnieR:


sensor/QuicLogPlugin.js
```diff
@@ -163,9 +162,8 @@ class QuicLogPlugin extends Sensor {
     rule.pro('udp');
     rule.dport(443);
     rule.jmp('LOG --log-level 7');
```
caixingcheng, line 164: missing option '-D'

PR #8688, opened 2026-01-09
Reduce the frequency of Redis bgsave operations

chgchi:


chgchi:
simple fix for issue https://github.com/firewalla/firecommit/issues/6599

PR #8921, opened 2026-04-13
Orange 1.982 Prod Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The change is tightly scoped and easy to understand: it adds a small platform hook to set `gbe_min_ipg_11B` during startup on `orange`.
- The script is defensive about the sysfs node being absent:
  - it searches dynamically,
  - exits early when nothing is found.
- Using a `before_main_run` hook is the right integration point for a one-time platform tuning like this.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** The script uses:
  ```bash
  [[ -n "$gbe_file" ]] || return 0
  ```
  `return` is only valid when the file is sourced or executed inside a shell function. This hook is a standalone script with `#!/bin/bash`, so if the sysfs node is missing it can emit:
  - `return: can only 'return' from a function or sourced script`

  and exit non-cleanly. This should be `exit 0`, not `return 0`.
- **Low:** The script blindly writes `1` without checking whether the current value is already `1`. Not a correctness bug, but it does unnecessary privileged work every run and makes the hook less self-documenting.
- **Low:** There is no logging on success or failure. For a hardware/platform workaround, a small log line would help future debugging if the expected tuning is missing.

### 💡 Suggestions
- Replace:
  ```bash
  return 0
  ```
  with:
  ```bash
  exit 0
  ```
- Consider making the write idempotent by checking the current value before overwriting.
- Add a lightweight log message when:
  - the sysfs node is not found,
  - the value is set,
  - the write fails.

  That would make this kind of platform-specific knob much easier to verify in the field.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8921  
Head SHA: `3705d9f12f30b256882577c965b5407317e43bef`  
Checked at: 2026-04-13 12:15:46 CST

PR #8922, opened 2026-04-13
fix iptables-restore error handling during init

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The refactor has a clear goal: separate the generated “restore state” files from the setup-script snapshots by introducing:
  - `ipset` vs `ipset.script`
  - `iptables` / `ip6tables` vs `iptables.script` / `ip6tables.script`

  That is a good direction and should reduce accidental cross-use of runtime vs setup artifacts.
- The initialization fallback in `IptablesControl` is sensible in principle:
  - if restore fails during initialization,
  - flush the relevant family,
  - restore from the setup snapshot,
  - then continue with per-rule execution.

  That is a pragmatic recovery path.
- The `flush_iptables.sh` update improves reuse by allowing:
  - IPv4 only
  - IPv6 only
  - both families

  instead of always flushing both.
- The `IpsetControl` cleanup around `_swp` handling is a real maintainability improvement. Avoiding duplicate flush/destroy operations for swap sets makes the operation queue clearer.
- The `getIpsetRestoreFile(script = false)` naming is better than the previous private helper once there are now two file variants.

- No linked issue found.

### ⚠️ Issues found
- **High:** `Rule.exec()` now logs full command text plus stack trace at `error` level for every failed rule execution:
  ```js
  log.error(cmd, err.message, new Error().stack)
  ```
  That is a noisy observability change and may leak more command detail than necessary into error logs. Earlier behavior was already chatty; this makes it even louder and elevates all individual failures to hard errors, even in fallback paths where some failures may be expected during reconciliation.
- **Medium:** The initialization fallback in `IptablesControl` swallows setup-restore failure and still proceeds to execute queued commands individually:
  ```js
  } catch (setupErr) {
    log.error(`Error restoring setup file for v${family}: ${setupErr.message}`);
  }
  log.info(`Executing queued commands individually for v${family}...`);
  ```
  If both bulk restore and setup restore fail, continuing may leave the firewall in a half-reset state with weaker guarantees than before. That may be acceptable as best effort, but it is risky and should be treated more explicitly.
- **Medium:** `flush_iptables.sh` now preserves DHCP/VPN NAT rules only when family `4` is selected, which is correct for iptables, but the refactor subtly changes execution structure a lot. There is no visible validation here that the `FW_*` chain recreation remains equivalent for both:
  - `MANAGED_BY_FIREROUTER=yes`
  - non-managed mode

  after the loop conversion. This is the kind of shell refactor that is easy to get slightly wrong without regression coverage.
- **Medium:** `readSetupScriptResult()` / `readSetupScriptResults()` now read `.script` files only. That matches the new naming, but it also means any environment that still generates the old file names will silently stop loading setup state. The migration assumption may be valid, but it is not guarded or backward-compatible.
- **Low:** `getIpsetRestoreFile(script = false)` is public now while `_getIptablesRestoreFile(...)` remains private. The API naming is slightly inconsistent between the two controls.
- **Low:** `log.error(\`Error restoring iptables v${family}: ${err.stderr}\`)` assumes `err.stderr` is the useful payload. In many exec failures that field is empty or truncated while `err.message` is more informative. The follow-up `log.info(err.message)` helps, but the split is awkward.

### 💡 Suggestions
- Consider keeping individual-rule execution failures at `warn` unless they truly indicate unrecoverable state. Logging full command + stack at `error` for every failed rule can make production logs much harder to use.
- In the initialization fallback path, distinguish:
  - bulk restore failed
  - setup restore succeeded, continue
  - setup restore failed, fallback is now degraded

  and consider surfacing a stronger failure signal if both restore layers fail.
- If environments may still have old non-`.script` setup files around, consider a temporary fallback read path or at least a migration note.
- Add regression coverage for `flush_iptables.sh` with:
  - family `4`
  - family `6`
  - both families
  - `MANAGED_BY_FIREROUTER=yes/no`
  - DHCP mode preserving MASQUERADE
- Consider aligning helper naming across ipset/iptables controls for consistency.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8922  
Head SHA: `d3725d013ae77c35ae2dcd163d1fee5e1177a2c9`  
Checked at: 2026-04-13 15:01:04 CST

jasonlyc (approved):

PR #8923, opened 2026-04-13
record local switch flow statistics into upload/download sumflows

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The new message types are introduced cleanly and wired through both producer/consumer sides:
  - `MSG_FWAPC_SWITCH_ACL_ACCOUNTING`
  - `MSG_FLOW_SWITCH_ACCOUNTING`
- The overall data path makes sense:
  - AP/switch accounting message arrives in `APCMsgSensor`
  - records are normalized into per-endpoint flow events
  - `FlowAggregationSensor` aggregates them into the existing traffic cache model
- Emitting two directional views for each record is the right shape for downstream accounting:
  - `mac1 -> mac2`
  - `mac2 -> mac1`
- The aggregation code also tries to preserve existing rollup dimensions:
  - device
  - interface
  - tag
  - global

  which is the right integration point rather than inventing a parallel accounting store.
- There is basic malformed-JSON protection on the subscription path, which is good for a pub/sub-fed sensor.

- No linked issue found.

### ⚠️ Issues found
- **High:** `processSwitchACLAccountingMessage()` does not validate `ts` before converting it:
  ```js
  const tsInSeconds = ts / 1000;
  ```
  If `ts` is missing, malformed, or not numeric, this becomes `NaN`, and the downstream flow is still emitted. `processSwitchAccountingFlow()` only checks `!ts`, which does not reject `NaN`, so bad timestamps can poison tick bucketing.
- **High:** The new flow key:
  ```js
  const key = `${mac}:${dstMac}:switch`;
  ```
  does not include direction, interface, or tag context. Within a tick, multiple switch-accounting events for the same MAC pair but different interface/tag situations will be merged into one entry. That can silently collapse flows that should stay distinct for attribution purposes.
- **Medium:** `APCMsgSensor` does sequential async lookups inside a loop:
  - `hostTool.getTags(host1, intf1)`
  - `hostTool.getTags(host2, intf2)`

  for every record. If a switch sends a large accounting batch, this becomes unnecessarily slow and can back up message handling.
- **Medium:** There is no deduplication / monotonicity protection. If the switch accounting payload is cumulative and the same sample is delivered twice, the code will emit and aggregate both copies, double-counting traffic.
- **Medium:** `processSwitchAccountingFlow()` infers intra-traffic using:
  ```js
  if (uidTickKey.startsWith('intf:') && intf && intf === dIntf)
  ```
  and similar tag/global logic, but the `uidTickKey` has already been rewritten into:
  ```js
  ${key}:local@${tick}
  ```
  so `startsWith('intf:')`, `startsWith('tag:')`, and `startsWith('global')` will no longer match as intended once `@tick` is appended. That makes the `intra` marking logic wrong.
- **Medium:** `uidTickKey.split(':')[1]` is brittle for tag extraction after the later key rewriting. For a key like `tag:123:local@...`, this happens to return `123`, but it is relying on string layout after multiple transformations and is easy to break.
- **Low:** The sensor fetches:
  - `switch_mac`

  from the payload comment/spec, but the actual implementation ignores it completely. If the same MAC pair can appear on different switches, there is no switch-level attribution or sanity check.
- **Low:** Error logging includes full flow payloads:
  ```js
  log.error(`Failed to process switch accounting flow`, event.flow, err.message);
  ```
  Depending on volume, this can get noisy.

### 💡 Suggestions
- Validate `ts` explicitly before emitting any switch-accounting flow, and reject non-finite timestamps.
- Revisit the aggregation key so it does not accidentally merge distinct accounting contexts.
- Batch or parallelize per-record host/tag lookups, or cache MAC→host/tag resolution within a message batch.
- Confirm whether switch ACL accounting records are deltas or cumulative counters; if cumulative, add duplicate/replay protection.
- Fix the `intra` detection logic so it uses the original scope key before `:local@tick` rewriting, or keep parsed metadata alongside the cache key instead of re-parsing strings.
- If `switch_mac` matters operationally, either include it in attribution or validate/document why it is intentionally ignored.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8923  
Head SHA: `9b53dce915d9b4f82cfaf2742283a4bd974e09d9`  
Checked at: 2026-04-14 00:15:57 CST

MelvinTo (approved):

PR #8924, opened 2026-04-14
support embeddedOps in init API for batching embedded get requests

yijin-png:
support embeddedOps in init API for batching embedded get requests

### DIFF
--- a/controllers/netbot.js
+++ b/controllers/netbot.js
@@ -4042,6 +4042,7 @@ class netBot extends ControllerBot {
               await sysManager.updateAsync()
               const fwapcOps = data.fwapcOps || [];
               const dapOps = data.dapOps || [];
+              const embeddedOps = data.embeddedOps || [];
               try {
                 const json = {};
                 const tasks = fwapcOps.map(async (op) => {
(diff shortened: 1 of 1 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The feature is useful: allowing a batched request to include `embeddedOps` can reduce round trips and make the caller fetch related data in one response.
- The implementation keeps `embeddedOps` isolated from the existing:
  - `fwapcOps`
  - `dapOps`

  flow, which is good for readability.
- Supporting a custom response key is also practical:
  - defaults to `item`
  - allows `key` override when the caller wants a stable response shape
- Per-op failure handling is also better than failing the whole batch on one bad embedded request.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** Errors from `embeddedOps` are silently downgraded to warnings and omitted from the response:
  ```js
  log.warn(`Failed to execute embeddedOps item ${item}:`, err.message);
  ```
  That means the caller cannot distinguish:
  - “field legitimately absent”
  - “embedded op failed internally”

  In a batched API, silent partial failure is hard to debug and easy for clients to mis-handle.
- **Medium:** `json[key || item] = await this.getHandler(...)` can overwrite existing response fields. Since this batch also merges:
  - `fwapcOps`
  - `dapOps`
  - `hostManager.toJson(...)`

  an `embeddedOps` key can collide with another top-level property and silently replace it.
- **Medium:** `target` is passed straight through from the request into `getHandler`:
  ```js
  target: target,
  ```
  but this diff adds no validation of which targets/items are allowed in `embeddedOps`. If `getHandler` exposes broader internal read paths, this widens the effective API surface through a generic batching escape hatch.
- **Low:** The comment says:
  ```js
  // defaults to item as json key, but allows custom key override
  ```
  but there is no normalization or namespacing, so two ops with the same `key` or `item` will race to the same slot and the last completion wins.
- **Low:** The implementation inherits `apiVer` from the outer request, which is probably intended, but there is no per-op override/compatibility check. That can be fine now, but it is worth documenting because batched embedded requests may mix items with different version expectations.

### 💡 Suggestions
- Return explicit per-op error information in the JSON instead of only logging warnings, so clients can detect partial failure.
- Protect against top-level key collisions, either by:
  - reserving an `embedded` namespace, or
  - rejecting duplicate/conflicting keys.
- Validate allowed `item`/`target` combinations for `embeddedOps` before dispatching to `getHandler`.
- Consider detecting duplicate `key || item` values up front and rejecting or warning deterministically.
- If `embeddedOps` is intended as a public contract, document its supported schema and failure behavior clearly.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8924  
Head SHA: `c5defb59f6507c80d2c9f6cb8d04d212cffdd59f`  
Checked at: 2026-04-14 11:46:21 CST

jasonlyc (approved):

PR #8925, opened 2026-04-14
fix dnsmasq conf change detection when config files move between directories

yijin-png:
Include file paths in md5 checksum calculation in checkConfsChange(), so that moving config files between directories is correctly detected as a change and triggers a dnsmasq restart.

### DIFF
--- a/extension/dnsmasq/dnsmasq.js
+++ b/extension/dnsmasq/dnsmasq.js
@@ -2316,7 +2316,7 @@ module.exports = class DNSMASQ {
     try {
       let md5sumNow = '';
       for (const confs of paths) {
-        const stdout = await execAsync(`find ${confs} -type f | (while read FILE; do (cat "\${FILE}"; echo); done;) | sort | md5sum | awk '{print $1}'`).then(r => r.stdout).catch((err) => null);
+        const stdout = await execAsync(`find ${confs} -type f | sort | (while read FILE; do echo "\${FILE}"; cat "\${FILE}"; echo; done;) | md5sum | awk '{print $1}'`).then(r => r.stdout).catch((err) => null);
         md5sumNow = md5sumNow + (stdout ? stdout.split('\n').join('') : '');
       }
       const md5sumBefore = await rclient.getAsync(dnsmasqConfKey);


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The intent of the change is sound: including file paths in the digest makes the config fingerprint sensitive not only to file content, but also to which file owns that content. That avoids collisions when the same concatenated contents can be produced from different file layouts.
- Sorting the `find` output before hashing is a good improvement for determinism across filesystem traversal order.
- The change is small, easy to reason about, and localized to the checksum path.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** The shell command still interpolates `${confs}` directly into `find ${confs} -type f ...` without quoting or escaping. If any configured path ever contains whitespace or shell metacharacters, the checksum command can fail or behave unexpectedly. Even if current callers only pass trusted paths, this remains brittle shell construction.
- **Medium:** `while read FILE; do ...` still uses bare `read` without `-r`. File names containing backslashes will be mangled before `cat` runs, which can produce incorrect hashes or spurious failures.
- **Medium:** `find ... | sort | while read ...` is line-delimited, so file names containing newlines are still not handled safely. That may be rare in practice, but the pipeline is not path-safe in the general case.
- **Low:** This changes checksum semantics for all existing configs. If the stored md5 in Redis was produced by the old algorithm, the first run after deployment will always look like a config change even if content is identical. That may be acceptable, but it is worth calling out if the checksum drives reload behavior.
- **Low:** The implementation still shells out to a fairly complex pipeline for something that is effectively “hash file path + content.” It works, but it remains harder to maintain and audit than a JS-native implementation.

### 💡 Suggestions
- Quote or otherwise safely escape `${confs}` before embedding it in the shell command.
- Use a null-delimited pipeline if you want this to be robust for arbitrary file names, e.g. `find -print0` plus a null-safe reader/sorter.
- At minimum, switch to `read -r` to avoid backslash mangling.
- If a one-time checksum invalidation is expected because of the algorithm change, document it in the PR or rollout notes.
- Longer term, consider moving this hashing logic into Node instead of relying on a shell pipeline.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8925  
Head SHA: `fea98aacaa9ca39730f774ba75a0c7e187f0b6ab`  
Checked at: 2026-04-14 16:00:59 CST

jasonlyc (approved):

PR #8926, opened 2026-04-14
device detection improvments

drahnieR:
add expiry to bonjour source
store name based detection and ua detection source info
name based detection now triggered by DeviceUpdate event

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -125,3 +125,7 @@ api/db/token.json
 /extension/frp/frpc.ini
 /extension/frp/frpc.support.ini
 /extension/frp/frpc.support.service
+
+.claude/
+CLAUDE.md
+AGENTS.md
--- a/hook/DeviceHook.js
+++ b/hook/DeviceHook.js
@@ -68,7 +68,7 @@ class DeviceHook extends Hook {
     let ipv6Addr = host.ipv6Addr
 
     if (!mac) { // ignore if no mac
-      log.warn("Invalid MAC address for process device update:", event);
+      log.warn("Invalid MAC address for process device update:", event.message);
       return;
     }
 
@@ -76,7 +76,7 @@ class DeviceHook extends Hook {
      * Filter out IPv4 broadcast address for any monitoring interface
      */
     if (ipv4Addr && sysManager.isMulticastIP4(ipv4Addr)) {
-      log.warn(`Ignore IP address ${ipv4Addr} as broadcast/multicast address`, event);
+      log.warn(`Ignore IP address ${ipv4Addr} as broadcast/multicast address`, event.message);
       return
     }
 
--- a/net2/Monitorable.js
+++ b/net2/Monitorable.js
@@ -1,4 +1,4 @@
-/*    Copyright 2021-2025 Firewalla Inc.
+/*    Copyright 2021-2026 Firewalla Inc.
  *
  *    This program is free software: you can redistribute it and/or  modify
  *    it under the terms of the GNU Affero General Public License, version 3,
@@ -235,7 +235,7 @@ class Monitorable {
       obj = _.pick(obj, fields)
     }
 
-    log.debug('Saving', this.getMetaKey(), fields, obj)
+    log.debug('Saving', this.getMetaKey(), obj)
     if (Object.keys(obj).length)
       await rclient.hmsetAsync(this.getMetaKey(), obj)
   }
--- a/net2/config.json
+++ b/net2/config.json
@@ -179,7 +179,8 @@
         "_raop", "_sleep-proxy", "_remotepairing", "_remotepairing-tunnel", "_apple-mobdev2",
         "_asquic", "_dacp", "_spotify-connect", "_spotify-social-listening", "_trel"
       ],
-      "ignoreNames": []
+      "ignoreNames": [],
+      "expire": 2592000
     },
     "AdvancedNmapSensor": {},
     "NmapSensor": {},
@@ -489,7 +490,10 @@
     "DeviceNameUpdateSensor": {},
     "DeviceIdentificationSensor": {
       "intervalListCheck": 86400,
-      "interval": 3600
+      "interval": 3600,
+      "expire": {
+        "bonjour": 2592000
+      }
     },
     "IPv6DiscoverySensor": {},
     "NetworkMonitorSensor": {
--- a/sensor/BonjourSensor.js
+++ b/sensor/BonjourSensor.js
@@ -310,12 +310,14 @@ class BonjourSensor extends Sensor {
 
     if (Object.keys(detect).length) {
       log.verbose('Bonjour', mac, detect)
+      const source = _.pick(service, ['type', 'name', 'ipv4Addr'])
+      source.expire = Date.now() / 1000 + this.config.expire || 30 * 24 * 3600
       sem.emitLocalEvent({
         type: 'DetectUpdate',
         from: 'bonjour',
         mac,
         detect,
-        source: _.pick(service, ['type', 'name', 'ipv4Addr']),
+        source,
         suppressEventLogging: true,
       })
     }
--- a/sensor/BonjourSensor.js
+++ b/sensor/BonjourSensor.js
@@ -311,7 +311,7 @@ class BonjourSensor extends Sensor {
     if (Object.keys(detect).length) {
       log.verbose('Bonjour', mac, detect)
       const source = _.pick(service, ['type', 'name', 'ipv4Addr'])
-      source.expire = Date.now() / 1000 + this.config.expire || 30 * 24 * 3600
+      source.expire = Date.now() / 1000 + (this.config.expire || 30 * 24 * 3600)
       sem.emitLocalEvent({
         type: 'DetectUpdate',
         from: 'bonjour',
--- a/sensor/DeviceIdentificationSensor.js
+++ b/sensor/DeviceIdentificationSensor.js
@@ -1,4 +1,4 @@
-/*    Copyright 2023-2024 Firewalla Inc.
+/*    Copyright 2023-2026 Firewalla Inc.
  *
  *    This program is free software: you can redistribute it and/or  modify
  *    it under the terms of the GNU Affero General Public License, version 3,
@@ -45,48 +45,24 @@ class DeviceIdentificationSensor extends Sensor {
     this.expire = config.get('bro.userAgent.expires')
 
     for (const host of hosts) try {
-      // keep user feedback and other detection sources
-      const keepsake = _.pick(host.o.detect, ['feedback', 'bonjour', 'cloud'])
-
-      host.o.detect = await this.detect(host)
-
-      Object.assign(host.o.detect, keepsake)
+      if (!host.o.detect) host.o.detect = {}
+      if (host instanceof Host && sm.isFirewallaMac(host.o.mac)) {
+        log.debug('Found Firewalla device', host.o.mac)
+        host.o.detect.nameBased = { name: 'Firewalla' }
+        continue
+      }
+      host.o.detect.ua = await this.userAgentDetect(host)
       await this.mergeAndSave(host)
     } catch(err) {
       log.error('Error identifying device', host.o.mac, err)
     }
   }
 
-  async detect(host) {
-    if (host instanceof Host && sm.isFirewallaMac(host.o.mac)) {
-      log.debug('Found Firewalla device', host.o.mac)
-      return { name: 'Firewalla' }
-    }
-
-    const name = getPreferredName(host.o)
-    if (name) {
-      const type = await nameToType(name)
-      if (type) {
-        log.debug('Type from name', host.o.mac, name, type)
-        return { type }
-      }
-    }
-
+  async userAgentDetect(host) {
     const key = `host:user_agent2:${host.o.mac}`
     const results = await rclient.zrevrangebyscoreAsync(key, this.now, this.now - this.expire).map(JSON.parse)
 
-    // parse legacy results to give detection an instant bootup
-    if (httpFlow.detector) try {
-      const oldKey = `host:user_agent:${host.o.mac}`
-      const oldResults = await rclient.smembersAsync(oldKey)
-      for (const result of oldResults) {
-        const ua = JSON.parse(result).ua
-        // should be fine not to dedup here
-        results.push(httpFlow.detector.detect(ua))
-      }
-    } catch(err) {
-      log.error('Error getting legacy user agents for', host.o.mac, err)
-    }
+    // lagacy key host:user_agent:<mac> should no longer available
 
     const deviceType = {};
     const deviceBrand = {};
@@ -143,22 +119,50 @@ class DeviceIdentificationSensor extends Sensor {
       counter[key] = 1
   }
 
+  async applyNameBasedType(mac) {
+    const host = hostManager.getHostFastByMAC(mac)
+    if (!host) return
+    const name = getPreferredName(host.o)
+    if (!name) return
+    const type = await nameToType(name)
+    if (!type) return
+    log.debug('Type from name', mac, name, type)
+    sem.emitLocalEvent({
+      type: 'DetectUpdate',
+      from: 'nameBased',
+      mac,
+      detect: { type },
+      suppressEventLogging: true,
+    })
+  }
+
   async mergeAndSave(host) {
     const detect = host.o.detect
     if (!Object.keys(detect)) return
 
-    Object.assign(detect, detect.bonjour, detect.cloud)
-    for (const key of Object.keys(detect)) {
+    // key deletion won't affect source data
+    const bonjour = detect && detect.bonjour && JSON.parse(JSON.stringify(detect.bonjour)) || {}
+    const now = Date.now() / 1000
+    for (const key in bonjour) {
       if (key.endsWith('.source'))
-        delete detect[key]
+        if (!bonjour[key].expire)
+          // add default expire time if not set
+          detect.bonjour[key].expire = now + (this.config.expire.bonjour || 30 * 24 * 3600)
+        else if (bonjour[key].expire > now)
+          delete bonjour[key.slice(0, -7)];
+        delete bonjour[key]
     }
+
+    const cloud = host.o._identifyExpiration && host.o._identifyExpiration > now ? detect.cloud : {}
+
+    Object.assign(detect, detect.ua, detect.nameBased, bonjour, cloud)
     log.debug('Saving', host.o.mac, detect)
     await host.save('detect')
 
     const type = detect.feedback && detect.feedback.type || detect.type
     if (type) {
       log.verbose(`Applying ${type} tag to ${host.getUniqueId()}`)
-      let tag = await TagManager.getTagByName(type, Constants.TAG_TYPE_DEVICE);
+      let tag = TagManager.getTagByName(type, Constants.TAG_TYPE_DEVICE);
       if (!tag)
         tag = await TagManager.createTag(type, { type: Constants.TAG_TYPE_DEVICE })
       // policy should be loaded at this point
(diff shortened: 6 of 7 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The overall direction makes sense: this PR is trying to make device identification more source-aware and less destructive by:
  - separating `ua`, `bonjour`, `cloud`, and `nameBased`
  - attaching expiration metadata to transient detection sources
  - avoiding wholesale overwrite of `host.o.detect`
- Adding expiration support for Bonjour and identity sources is a good idea in principle. Detection data that comes from ephemeral observations should not live forever.
- The new `nameBased` event path is also directionally useful:
  - it decouples name-based inference from the hourly UA pass
  - it lets device naming updates trigger type inference sooner
- The logging cleanup in `DeviceHook` is reasonable. Logging `event.message` instead of the whole event object reduces noise and likely avoids dumping large payloads.

- No linked issue found.

### ⚠️ Issues found
- **High:** The Bonjour expiration logic is inverted in `mergeAndSave()`. This block:
  - adds a default expire if missing
  - **deletes the field when `expire > now`**

  means unexpired Bonjour-derived values are removed, while expired ones are kept. That is the opposite of the intended behavior and will suppress valid Bonjour detection data almost immediately.
- **High:** `source.expire = Date.now() / 1000 + this.config.expire || 30 * 24 * 3600` in `BonjourSensor` has precedence issues. If `this.config.expire` is undefined, the expression becomes `NaN || 2592000`, which yields `2592000` as an absolute Unix timestamp near 1970, not “now + 30 days.” The fallback is therefore wrong.
- **High:** `applyNameBasedType()` emits:
  ```js
  detect: { type }
  ```
  under `from: 'nameBased'`, but `mergeAndSave()` merges `detect.nameBased` directly into the top-level object. That produces:
  - `detect.type` from name-based source, yes
  - but also persists the nested source object format inconsistently with the Firewalla special-case path, which writes `host.o.detect.nameBased = { name: 'Firewalla' }`

  The source schema is inconsistent and will be hard to reason about.
- **Medium:** The Firewalla special-case path now does:
  ```js
  host.o.detect.nameBased = { name: 'Firewalla' }
  ```
  and then `continue`, without calling `mergeAndSave(host)`. That means the classification never gets persisted in this code path unless something else later saves the host.
- **Medium:** `mergeAndSave()` mutates `detect.bonjour[key].expire` while iterating over a cloned `bonjour` object. The comment says “key deletion won't affect source data,” but the code still writes back into `detect.bonjour`. That mixed clone/original behavior is hard to follow and easy to break.
- **Medium:** `let tag = TagManager.getTagByName(type, Constants.TAG_TYPE_DEVICE);` dropped the previous `await`. If `getTagByName` is async, this is now a bug. If it was intentionally changed to sync, the surrounding code gives no clue and this deserves confirmation.
- **Medium:** The PR removes legacy user-agent parsing entirely:
  ```js
  // lagacy key host:user_agent:<mac> should no longer available
  ```
  If older installations still have useful UA evidence only in the legacy key, this creates a cold-start regression in identification accuracy after upgrade.
- **Low:** `log.debug('Saving', this.getMetaKey(), obj)` in `Monitorable.save()` drops the `fields` argument from logs. That may reduce observability when debugging partial writes.
- **Low:** `.gitignore` changes for `.claude/`, `CLAUDE.md`, and `AGENTS.md` are unrelated to the device-identification behavior and should ideally be separated.

### 💡 Suggestions
- Fix the Bonjour expiration check first. The merge should keep entries whose `expire > now` and drop ones that are expired.
- Fix the fallback expire calculation in `BonjourSensor` by parenthesizing it so the default is added to `now`, not used as an absolute timestamp.
- Normalize the `nameBased` source shape so all producers store the same schema.
- Persist the Firewalla name-based detection path by calling `mergeAndSave(host)` before `continue`.
- Confirm whether `TagManager.getTagByName` is synchronous. If not, restore the `await`.
- If legacy UA data still matters for upgrades, consider keeping a one-release migration/fallback path instead of removing it abruptly.
- Split the unrelated `.gitignore` cleanup from the behavioral changes.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8926  
Head SHA: `9e9a84e6e15c90e964a21017363af7596606828f`  
Checked at: 2026-04-14 16:16:03 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Adding expiration metadata to Bonjour-derived detection sources is directionally good. Device-identification signals should not live forever, and moving toward bounded freshness is the right systems behavior.
- Splitting UA-based detection from name-based detection is also a reasonable design direction. It makes the detection pipeline easier to reason about and opens the door for source-specific refresh logic.
- The new buffering around `NewDeviceFound` / `RegularDeviceInfoUpdate` in `DeviceIdentificationSensor.run()` is a practical way to avoid thrashing on bursts of updates while still waiting for the host object to settle.

- No linked issue found.

### ⚠️ Issues found
- **High:** `sensor/DeviceIdentificationSensor.js` breaks persistence/schema compatibility for Firewalla devices:
  - previously `detect()` returned `{ name: 'Firewalla' }`
  - now the code writes `host.o.detect.nameBased = { name: 'Firewalla' }` and `mergeAndSave()` only flattens `detect.nameBased` at save time
  - but that Firewalla fast-path does `continue` before `mergeAndSave(host)` is called
  - so Firewalla devices are left with nested `detect.nameBased` only, unlike the flattened `detect.type` / `detect.name` shape the rest of the code expects
- **High:** `sensor/DeviceIdentificationSensor.js` likely introduces an async bug here:
  - `let tag = TagManager.getTagByName(type, Constants.TAG_TYPE_DEVICE);`
  - If `TagManager.getTagByName` is async, this now stores a Promise in `tag`, making `if (!tag)` always false and skipping `createTag()`. That would break automatic device-tag application entirely.
- **High:** The Bonjour expiration cleanup logic is wrong in `mergeAndSave()`:
  - `else if (bonjour[key].expire < now) delete bonjour[key.slice(0, -7)];`
  - this deletes the value when the source metadata is expired, but keeps the stale `*.source` metadata key until the subsequent `delete bonjour[key]`
  - more importantly, if multiple Bonjour-derived fields share the same source object semantics, this cleanup is piecemeal and can leave partially inconsistent flattened state depending on key order
- **Medium:** Removing legacy UA parsing in `userAgentDetect()` can regress cold-start identification:
  - the old code explicitly parsed `host:user_agent:<mac>` to bootstrap detection immediately after restart
  - the new code drops that path entirely and just comments “lagacy key ... should no longer available”
  - unless there is a guaranteed migration elsewhere, devices that only have legacy UA data will now lose detection quality until enough new `host:user_agent2:*` data accumulates
- **Medium:** `sensor/BonjourSensor.js` adds `source.expire`, but the expression must be parenthesized exactly as now to be correct. This PR fixed that, but it also means all existing previously stored Bonjour source entries without `expire` are handled lazily in `mergeAndSave()`. That is fine as a migration path, but it should be called out because the system will behave differently for old vs new records until each host is reprocessed.
- **Medium:** `sensor/DeviceIdentificationSensor.js` now merges `detect.ua`, `detect.nameBased`, `bonjour`, and `cloud` into the root object, but it no longer preserves arbitrary prior subtrees like `feedback` via a keepsake pass. `feedback` survives only because it was already on `detect` and is not overwritten here. That is subtle and easy to break later; the old logic made the preservation contract more explicit.
- **Low:** `hook/DeviceHook.js` changes logging from `event` to `event.message`. That reduces log context for malformed-device debugging. If `event.message` is missing or much smaller than the full event payload, troubleshooting becomes harder.
- **Low:** The `.gitignore` additions for `.claude/`, `CLAUDE.md`, and `AGENTS.md` are unrelated to the device-identification logic and should ideally be split out. They add noise to an already behavior-heavy PR.
- **Low:** `net2/Monitorable.js` changes debug logging from `fields, obj` to just `obj`, which removes information useful when diagnosing partial-field saves.

### 💡 Suggestions
- Fix the Firewalla-device fast path so it still goes through the same persisted shape as all other devices. Either:
  - write to the flattened root keys before `continue`, or
  - assign `nameBased` and still call `mergeAndSave(host)`.
- Verify `TagManager.getTagByName`’s contract and restore `await` if it returns a Promise.
- Tighten the Bonjour expiration cleanup so expired data removal is deterministic and explicit:
  - derive the base key once
  - delete both the flattened key and its `.source` metadata together
  - avoid mutating mixed copies (`detect.bonjour` vs cloned `bonjour`) in different places.
- Reconsider dropping legacy UA parsing unless you have a proven migration window. If the old key truly cannot exist anymore, add a comment explaining why that invariant is safe.
- Add focused tests for:
  - Firewalla MAC detection persistence shape
  - expired Bonjour source removal
  - tag creation/application path
  - behavior when only legacy UA data exists
- Split unrelated `.gitignore` / logging cleanups into a separate PR to keep this one easier to validate.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8926  
Head SHA: `03a70f72b259cf51fc8486cce5784cb5d07ade63`  
Checked at: 2026-04-14 17:32:05 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Adding expiration metadata to Bonjour-derived detection data is directionally useful. Persisting source freshness is better than keeping Bonjour hints forever with no aging policy.
- The refactor in `DeviceIdentificationSensor` separates user-agent detection from merge/persist logic, which is a reasonable cleanup.
- Using `getHostFastByMAC()` in the event path is a sensible performance improvement for high-frequency `DetectUpdate` handling.
- The small log cleanups in `DeviceHook.js` and `DeviceOfflineSensor.js` reduce noisy object dumps and make log lines more focused.
- Ignoring local Claude/Codex workspace artifacts in `.gitignore` is harmless and practical.

- No linked issue found.

### ⚠️ Issues found
- **High:** The Firewalla-device fast path in `DeviceIdentificationSensor.job()` no longer persists the detection result.
  - For Firewalla MACs, the code now does:
    - `host.o.detect.nameBased = { name: 'Firewalla' }`
    - `continue`
  - That skips `mergeAndSave(host)`, so the detection never gets normalized/saved/tagged through the regular path.
  - This is a regression from the previous behavior, where Firewalla-device detection was part of the returned detect result and then flowed through persistence.
- **High:** `TagManager.getTagByName(type, Constants.TAG_TYPE_DEVICE)` lost its `await` in `_mergeAndSave()`.
  - If `getTagByName` is async, `tag` will be a promise object, the `if (!tag)` branch will never behave correctly, and `tag.o` / `tag.getTagUid()` access becomes unsafe.
  - This was already a key concern in earlier review and still looks wrong in the current diff.
- **Medium:** Bonjour expiration handling is still too fragile.
  - The code mutates `detect.bonjour` source records in-place, then deep-copies only `detect.bonjour` for pruning.
  - Expiration is attached generically to every Bonjour key using the same `source` object shape, but the merge logic assumes all non-`.source` keys have a corresponding source entry and removes fields solely based on `expire`.
  - That can silently delete still-useful Bonjour-derived fields if source bookkeeping becomes incomplete or partially migrated.
- **Medium:** `globalOn()` no longer starts the periodic identification job.
  - Previously it scheduled `this.job()` via `this.intervalTask`.
  - Now it only starts `checkList()` and relies on events for merges.
  - That means stale hosts that do not emit `NewDeviceFound` / `RegularDeviceInfoUpdate` / `DetectUpdate` may never be re-identified, even though the config still has `interval`.
- **Medium:** `mergeAndSave()` is now debounced with `setTimeout(..., 2000)` per MAC, but there is no cleanup on feature disable/globalOff beyond clearing interval timers.
  - Pending merge jobs can still fire after feature state changes or shutdown transitions, which makes lifecycle behavior less deterministic.
- **Low:** `lagacy key host:user_agent:<mac> should no longer available` removes the legacy instant-bootstrapping path but does not include a migration/fallback plan. If older persisted UA data is still present in deployed systems, detection quality may regress immediately after upgrade.
- **Low:** `Monitorable.js` changed the debug logging signature from logging `fields` + `obj` to logging only `obj`. That loses a little observability when partial-field saves are being debugged.

### 💡 Suggestions
- Restore persistence for Firewalla-device detection:
  - either call `await this.mergeAndSave(host)` before `continue`, or
  - route the fast path through the same normalized save pipeline as other detections.
- Re-check `TagManager.getTagByName(...)` and make it explicitly awaited if it is async. This is the highest-confidence correctness issue in the patch.
- Keep the periodic identification interval unless there is strong evidence the event-only model is sufficient. If you want to reduce cost, lower the cadence rather than removing it entirely.
- Make Bonjour expiration pruning more defensive:
  - treat missing `.source` as “unknown”, not automatically safe/unsafe
  - avoid mutating source state and derived merged state in the same object graph if you can help it
- Clear pending `mergeJobs` timeouts in `globalOff()` and/or when the feature is turned off.
- If the legacy `host:user_agent:<mac>` key is intentionally retired, document the expected detection-quality transition and consider a one-time migration path.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8926  
Head SHA: `18ef8ef6b45a400ac4772a93eebf7101eceed89f`  
Checked at: 2026-04-15 11:32:04 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Adding expiration metadata to Bonjour-derived detection data is a useful improvement. It gives the system a way to age out stale Bonjour hints instead of keeping them forever.
- The refactor in `DeviceIdentificationSensor` separates user-agent detection from merge/persist behavior, which makes the code easier to follow.
- Switching event-driven host lookup to `getHostFastByMAC()` is a sensible performance optimization for frequent `DetectUpdate` handling.
- The Firewalla-device fast path now still flows through `mergeAndSave(host)`, which is better than bypassing persistence entirely.
- The smaller log cleanups in `DeviceHook.js` and `DeviceOfflineSensor.js` reduce noisy object dumps.

- No linked issue found.

### ⚠️ Issues found
- **High:** `TagManager.getTagByName(type, Constants.TAG_TYPE_DEVICE)` still lost its `await` in `_mergeAndSave()`.
  - If `getTagByName` is async, `tag` becomes a promise, `if (!tag)` will not behave as intended, and later access like `tag.o` / `tag.getTagUid()` is unsafe.
  - This is the most concrete correctness issue in the diff and can break device-tag application.
- **High:** `globalOn()` no longer schedules the periodic identification job.
  - Previously the sensor ran `this.job()` immediately and then on an interval.
  - Now it only starts `checkList()` and relies on event-driven merges.
  - That means existing quiet hosts with no fresh `DetectUpdate`, `NewDeviceFound`, or `RegularDeviceInfoUpdate` may never get refreshed UA/name-based detection again, even though `interval` is still configured in `net2/config.json`.
- **Medium:** The Bonjour expiration merge path is still brittle.
  - `_mergeAndSave()` mutates `detect.bonjour` by adding synthetic `.source` entries and expire timestamps, then deep-copies it and deletes fields based on source expiry.
  - That makes the persisted structure depend on internal bookkeeping being complete and consistent across versions.
  - If a Bonjour key exists without a matching `.source`, behavior becomes implicit rather than clearly defined.
- **Medium:** `mergeAndSave()` is now debounced with `setTimeout(..., 2000)` per MAC, but pending timers are not cleared in `globalOff()`.
  - The feature can be turned off while already-scheduled merge jobs still fire later.
  - That creates lifecycle inconsistency and makes shutdown/disable behavior less deterministic.
- **Medium:** The legacy fallback from `host:user_agent:<mac>` was removed entirely.
  - If deployed systems still have useful legacy UA data but have not yet populated `host:user_agent2:<mac>`, detection quality can regress immediately after upgrade.
  - The comment says the legacy key “should no longer available”, but the code does not enforce or migrate that assumption.
- **Low:** `Monitorable.js` now logs only `obj` and no longer logs `fields` during partial saves. That reduces observability when debugging selective metadata writes.
- **Low:** `.gitignore` now ignores `AGENTS.md`, which may be intentional for local tooling, but in repos that already use `AGENTS.md` as human-readable instructions this can also hide legitimate tracked changes if contributors are not careful.

### 💡 Suggestions
- Restore `await` for `TagManager.getTagByName(...)` if the method is async. If it was intentionally changed to sync, confirm that contract explicitly and update surrounding call sites consistently.
- Keep the periodic `job()` schedule in `globalOn()`, even if event-driven merges remain. The periodic pass is still valuable for stale hosts and recovery from missed events.
- Clear all pending `mergeJobs` timers in `globalOff()` and ideally guard the delayed callback with a feature-enabled check before running `_mergeAndSave`.
- Make Bonjour expiration handling more defensive:
  - define behavior for missing `.source`
  - avoid mutating and deriving from the same nested object when possible
  - consider a clearer schema for per-source metadata
- If the legacy UA key is truly obsolete, consider a migration path or at least a transitional fallback window rather than removing it abruptly.
- Add tests for:
  - async/sync tag lookup behavior
  - pending merge timer cleanup on feature disable
  - host refresh when no new events arrive
  - Bonjour data with missing or expired `.source` entries

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8926  
Head SHA: `501b5ae08d3a4274524ae8c5558e70af5b94f1a2`  
Checked at: 2026-04-15 12:01:07 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Adding expiry metadata to Bonjour-derived detection data is a useful idea. It gives the pipeline a way to age out stale Bonjour hints instead of preserving them indefinitely.
- Using `getHostFastByMAC()` in the event path is a sensible optimization for frequent `DetectUpdate` handling.
- The small log cleanups in `DeviceHook.js` and `DeviceOfflineSensor.js` reduce noisy object dumps.
- The Firewalla-device path now still reaches `mergeAndSave(host)`, which is better than skipping persistence entirely.

- No linked issue found.

### ⚠️ Issues found
- **High:** `TagManager.getTagByName(type, Constants.TAG_TYPE_DEVICE)` still appears to be called without `await` in two places inside `DeviceIdentificationSensor.js`.
  - In `_mergeAndSave()`
  - In the preload tag creation loop
  If `getTagByName` is async, this is a real correctness bug: `tag` becomes a promise, `if (!tag)` won’t behave as intended, and later property/method access can break.
- **High:** `globalOn()` no longer schedules the periodic identification job.
  - Previously the sensor ran `this.job()` immediately and on an interval.
  - Now it only starts `checkList()` and relies on event-driven merges.
  - That means quiet existing hosts with no fresh events may never be re-identified, even though `interval` is still configured.
- **Medium:** The legacy UA fallback was removed entirely:
  - old `host:user_agent:<mac>` parsing is gone
  - comment says the key “should no longer available”
  That assumption may not hold across upgrades. If older deployments still have useful legacy UA data but have not yet populated `host:user_agent2:<mac>`, device identification quality can regress.
- **Medium:** The Bonjour expiry merge path is still fragile.
  - `_mergeAndSave()` mutates `detect.bonjour` to synthesize `*.source` entries with expiry, then deep-copies and prunes based on those generated fields.
  - This mixes bookkeeping and user-facing state in the same object graph, which makes partial/migrated states harder to reason about.
- **Medium:** `mergeAndSave()` now debounces work with `setTimeout(..., 2000)`, but `globalOff()` does not clear pending `mergeJobs`.
  - Jobs can still fire after disable/shutdown transitions, making lifecycle behavior less deterministic.
- **Low:** `Monitorable.js` now logs only `obj` and no longer logs `fields` for partial saves. That reduces observability when debugging selective persistence.
- **Low:** `.gitignore` now ignores `AGENTS.md` and `CLAUDE.md`. That may be intentional for local tooling, but it can also hide legitimate tracked docs changes if this repo ever adopts those files intentionally.

### 💡 Suggestions
- Re-check `TagManager.getTagByName(...)` and restore `await` if the method is async. This is the highest-confidence bug in the patch.
- Keep the periodic `job()` schedule in `globalOn()`, even if event-driven merging remains. The periodic pass is still useful for stale hosts and missed events.
- Clear pending `mergeJobs` timers in `globalOff()` and/or guard delayed callbacks with a feature-enabled check.
- Consider a transition strategy for the removed legacy UA key instead of assuming it is already gone everywhere.
- Make Bonjour expiry handling more explicit, ideally separating source metadata from the merged detection object more cleanly.
- If keeping the `.gitignore` additions, make sure they do not conflict with repo-level documentation conventions.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8926  
Head SHA: `06431851c8a8210c968efc0bc45c9f2c6fde84c0`  
Checked at: 2026-04-15 17:45:52 CST

PR #8927, opened 2026-04-14
add ut cases for testing dap learned flows count

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Adding a focused integration-style DAP test file is valuable. The coverage target is meaningful:
  - learned count behavior
  - local learned count behavior
  - state transitions
  - CLI-driven analyze flows
  - audit/APC-fed blocked flow scenarios
- The helper structure is generally good. Breaking setup into reusable builders/helpers makes the scenarios easier to extend:
  - host creation
  - conn/local flow creation
  - APC/audit fixtures
  - CLI wrappers
  - cleanup helpers
- `parseDapPolicyFromStdout()` is a pragmatic way to validate the human-facing CLI without needing a dedicated machine-readable mode.
- The tests are careful to isolate data with dedicated MAC/IP fixtures, which is the right instinct for Redis-backed integration tests.

- No linked issue found.

### ⚠️ Issues found
- **High:** Nearly all of the actual behavioral tests are marked `it.skip(...)`. As submitted, this adds a large test file but almost no effective regression protection. For a PR whose value is “check learnedCount/localLearnedCount behavior,” skipped tests do not validate the feature in CI.
- **High:** `createDapApcBlockedFlows()` and the APC-based path use APC message fields (`dst`, `sport`, `dport`, `proto`) that differ from the iptables/audit shape used elsewhere, but the local-learning optimizing scenario is left commented out and unresolved. That means one of the core claimed scenarios is not actually tested.
- **Medium:** `postDapDeviceState()` talks to `127.0.0.1:8843` directly. That makes the tests strongly environment-coupled and brittle in CI or developer machines where the service is not running, not listening there, or not loaded with compatible state. There is no skip/guard around service availability.
- **Medium:** `getDapBinary()` depends on `/data/dapenv/dap` via `process.env.DAP_BIN`, and `isDapBinaryExecutable()` is defined but never used to gate tests. So the file has no actual runtime protection against missing binaries even though the helper exists.
- **Medium:** The cleanup coverage is incomplete for a file this integration-heavy. The tests create/modify:
  - host keys
  - flow keys
  - intel keys
  - policy/internal keys
  - APC/audit buffers
  - local neighbor history

  but some intermediate side effects can still escape. For example, plugin instances and DB3 state are partly reused/shared, and test isolation depends heavily on all helpers behaving perfectly.
- **Medium:** `parseDapPolicyFromStdout()` is fragile:
  - depends on emoji markers
  - strips `…": false`
  - repairs trailing commas heuristically

  This makes the tests sensitive to presentation-only CLI changes rather than actual policy schema changes.
- **Low:** The file contains several comments like:
  - `investigate how to set device as isolation`
  - `lagacy`
  - references to unresolved scenarios

  which signals the test suite is still exploratory rather than production-ready.
- **Low:** There are some dead/unused pieces in the file, for example helpers or imports that are only partially exercised because the corresponding tests are skipped/commented. That adds maintenance cost.

### 💡 Suggestions
- Convert the key `it.skip(...)` cases into real runnable tests, or reduce the PR to the subset that can actually run reliably in CI.
- Use `isDapBinaryExecutable()` and a service-availability probe to skip gracefully when the environment is missing required runtime dependencies.
- If possible, add a machine-readable CLI output mode for DAP instead of parsing decorated stdout.
- Tighten isolation further:
  - ensure all Redis DB3 keys are cleaned
  - avoid shared long-lived plugin/client state across tests where possible
  - explicitly flush APC/audit plugin buffers in teardown
- Split exploratory scaffolding from committed regression tests; commented-out scenario drafts should not stay in the final suite.
- If local-learning optimizing behavior is the goal, finish one end-to-end APC/audit scenario and assert it concretely instead of leaving it commented.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8927  
Head SHA: `8d2be034f914761ab310a8b341cec5bb10731064`  
Checked at: 2026-04-14 17:16:35 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The test file is thoughtfully structured around reusable helpers instead of copy-pasting fixture setup:
  - host creation / cleanup
  - conn/local flow builders
  - APC / audit-log blocked-flow builders
  - CLI / HTTP wrappers
  - policy parsing helper
- The scenarios being targeted are meaningful for DAP validation:
  - `learnedCount` progression
  - `localLearnedCount` behavior
  - state transitions between learning / optimizing / not_applicable
  - audit/APC-fed blocked-flow effects
- The Redis cleanup coverage is broader than usual for integration-style tests, which is important for a stateful subsystem like DAP.

- No linked issue found.

### ⚠️ Issues found
- **High:** Every new test case is marked `it.skip(...)`. As submitted, this PR adds no executable coverage and will not protect the behavior it is trying to validate. That makes the change mostly dormant test scaffolding rather than an actual regression-prevention improvement.
- **Medium:** `parseDapPolicyFromStdout()` is brittle because it parses human-formatted CLI output using marker strings and cleanup regexes:
  - it depends on `📋 DAP Policy:` / `📊 DAP Internal:` remaining unchanged
  - it strips `,\s*…":\s*false`, which is effectively parsing a presentation format rather than a stable machine interface
  - any CLI wording/formatting change can break tests without any DAP behavior change
- **Medium:** The test mutates real shared runtime state and relies on best-effort cleanup, but there is no `finally`-style protection inside each scenario. If a setup step or assertion throws before `afterEach` finishes cleanly, leftover Redis / host / audit state can contaminate later runs and make failures non-deterministic.
- **Medium:** `before(() => { ... apcSensor = new APCMsgSensor({}); })` and the similar plugin setup use suite-scoped instances. For stateful sensors/plugins with internal buffers, reusing one instance across multiple tests is risky and can leak buffered data or mocked state between cases once the skipped tests are enabled.
- **Low:** There are several hard-coded sleeps like `await new Promise(resolve => setTimeout(resolve, 1000));`. Those usually make integration tests flaky:
  - too short on busy CI
  - unnecessarily slow on healthy systems
  A poll-until-condition pattern would be more reliable.
- **Low:** The describe titles have typos / unclear wording such as `cleck learnedCount value...`, which hurts maintainability a bit in test reports.

### 💡 Suggestions
- Convert at least the core happy-path cases from `it.skip` to active tests before merging, even if some harder APC/local scenarios remain skipped.
- If the DAP CLI can support a machine-readable mode, prefer that over scraping decorated stdout. A JSON output flag would make these tests much more stable.
- Instantiate `ACLAuditLogPlugin` / `APCMsgSensor` in `beforeEach` and tear them down per test once these scenarios are enabled.
- Replace fixed `setTimeout(1000)` waits with polling for the expected Redis/policy state, with a bounded timeout.
- Consider wrapping complex setup flows in helper functions that ensure cleanup on partial failure, especially for:
  - host creation
  - Redis keys in DB 0 / DB 3
  - audit buffers
- Clean up the test naming so failures are easier to read in CI output.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8927  
Head SHA: `c06af5987703d31ac3c26cc69c37739abc7bcea2`  
Checked at: 2026-04-14 17:31:14 CST

jasonlyc (approved):

PR #8928, opened 2026-04-14
record time series for switch ACL accounting traffic

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Extracting the repeated local-traffic accounting logic into `recordLocalTraffic()` is a good refactor. It removes duplicated bookkeeping from `BroDetect` and makes the intended “local MAC / interface / tag” accounting path easier to follow.
- Updating `APCMsgSensor` to call the same helper is directionally correct. It should reduce drift between local-flow accounting from Bro and switch-accounting-originated local traffic.
- The new helper preserves the prior distinction between:
  - intra traffic when both sides share the same interface/tag
  - directional upload/download accounting otherwise
  That’s the right model for local traffic aggregation.

- No linked issue found.

### ⚠️ Issues found
- **High:** `recordLocalTraffic()` still only records traffic from one endpoint’s perspective (`mac`, `intf`, `tags`) and never updates the destination MAC/interface/tag counters unless the caller invokes it twice. That is true for `APCMsgSensor`, but in `BroDetect` local-flow handling it is only called once:
  - `mac: localMac`
  - `intf: intfInfo && intfInfo.uuid`
  - `dIntf: dstIntfInfo && dstIntfInfo.uuid`
  - `tags, dstTags`

  Before this refactor, `BroDetect`’s inline code also only accounted for the local endpoint’s own `lo:<mac>`, `lo:intf:*`, and `lo:tag:*` buckets. If the intent of this PR is to unify Bro local flows with APC local-accounting semantics, it still leaves asymmetry between the two ingestion paths: APC now records both perspectives, but Bro still records only one. That can keep aggregate local counters inconsistent across sources.
- **Medium:** The helper changes `lo:global` semantics for local flows to always use `tupleIntra`:
  - `this.recordTraffic(tupleIntra, 'lo:global')`

  while the old inline code also used `tupleIntra` for `lo:global`, so that part is preserved. But the refactor makes it much easier for callers to assume `upload/download` are globally represented for local flows when they are not. The helper name is generic enough that future callers may misuse it unless the “global is intra-only” behavior is documented.
- **Medium:** `recordLocalTraffic()` assumes `tags` is iterable by key:
  - `for (const key in tags)`

  In current call sites that is probably true, but the helper is less defensive than the surrounding code paths. If any caller passes `undefined`/`null` tags, this will throw. The extracted helper should guard its inputs because it is now shared across Bro and APC paths.
- **Medium:** For interface-level accounting, `recordLocalTraffic()` only records the source-side interface bucket (`lo:intf:${intf}` or `lo:intra:intf:${intf}`), never the destination-side interface bucket. Again, APC works around this by calling the helper twice, but Bro does not. That means same-interface intra traffic and cross-interface local traffic can still be undercounted on the destination-side interface aggregation depending on ingestion path.
- **Low:** The helper uses `dstTags && dstTags[key] && dstTags[key].includes(tag)` but does not normalize array shape or dedupe repeated tags. If `hostTool.getTags()` can return duplicate tag IDs, local tag traffic can still be overcounted.

### 💡 Suggestions
- Decide explicitly whether local traffic accounting should be symmetric per endpoint.
  - If yes, `BroDetect` should call `recordLocalTraffic()` for both source and destination perspectives, similar to the new `APCMsgSensor` logic.
  - If no, document that the helper is intentionally source-perspective-only and that APC must invoke it twice.
- Make `recordLocalTraffic()` defensive:
  - default `tags = {}`
  - default `dstTags = {}`
  - guard `mac` / `intf` values explicitly
- Consider renaming or documenting the helper to clarify semantics, e.g. that:
  - `lo:global` records intra bytes only
  - `lo:<mac>` / `lo:intf:*` / `lo:tag:*` are source-perspective buckets
- Add tests comparing:
  - Bro local-flow accounting
  - APC local-flow accounting
  for the same logical local conversation, especially for:
  - same-interface intra traffic
  - cross-interface local traffic
  - shared tags vs distinct tags
- If duplicate tags are possible from `getTags()`, normalize/dedupe tag arrays before recording.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8928  
Head SHA: `e2d4495619f72d4db2fb5640a56b87a2dc0ec65e`  
Checked at: 2026-04-14 18:31:00 CST

MelvinTo (approved):

PR #8930, opened 2026-04-15
add bypass chain for category block/disturb rule

caixingcheng:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The overall approach makes sense: introducing a dedicated per-policy bypass chain for category `block` / `disturb` rules is a clean way to separate matching from marking and avoid duplicating the terminal action across multiple parent chains.
- The implementation is applied consistently across the main rule-installation paths in `control/Block.js`:
  - global
  - generic identities
  - devices
  - tags
  - interfaces
  - rule groups
- The destroy path is directionally correct: parent jumps are removed first, then the per-policy chain is flushed and deleted in `alarm/PolicyManager2.js`. That ordering avoids deleting a still-referenced chain.
- Reusing `prepareOutboundOptions(options)` to preserve the rule comment on the raw bypass-chain rule is a nice touch for operability/debugging.

- No linked issue found.

### ⚠️ Issues found
- **High:** The bypass chain creation in `alarm/PolicyManager2.js` is not idempotent.
  - On install, it unconditionally does `-N FW_${policy.pid}_BYPASS` for both families and only then `-F`.
  - If the policy is re-applied, partially rolled back, or recovered after a previous failed install, `-N` will fail when the chain already exists, and the subsequent `-F` will never serve as the intended “ensure empty before use” step.
  - Given how often policy rules can be re-enforced in firewall codepaths, this is a real correctness risk.
- **High:** The new bypass-chain rule in `control/Block.js` is inserted without any de-duplication/ensure semantics.
  - Each setup function builds a raw rule with `-A` on create and pushes it via `iptc.addRule(rawRule)`.
  - If rule application is retried without a clean destroy, the per-policy bypass chain can accumulate duplicate terminal mark rules, which changes behavior for counters and complicates teardown because `-D` removes only one matching rule at a time.
- **Medium:** Chain lifecycle is only handled in the install/uninstall flow in `alarm/PolicyManager2.js`, but the bypass behavior is injected deep inside the generic `setup*Rules()` functions in `control/Block.js`.
  - That coupling is brittle: any future caller that passes `byPassChain` without going through the same chain-create / chain-delete lifecycle will generate jumps into a chain that may not exist.
  - The abstraction boundary is leaky right now.
- **Medium:** Error handling is weak around chain creation and deletion.
  - If chain creation succeeds but later rule programming fails, the code can leave orphaned `FW_<pid>_BYPASS` chains behind.
  - Likewise, uninstall blindly attempts `-F` and `-X` without guarding for partial state.
  - For firewall orchestration code, best-effort cleanup usually needs explicit existence checks or tolerant error handling.
- **Medium:** `setupTagsRules()` programs two parent jumps (`FW_FIREWALL_DEV_G_BLOCK` and `FW_FIREWALL_NET_G_BLOCK`, similarly for disturb) into the same bypass chain, but only one shared terminal rule is added inside that chain.
  - That may be intended, but it means device-group and network-group traffic become indistinguishable once inside `FW_<pid>_BYPASS`.
  - If later logic depends on per-parent counters/debug traces, this design collapses that visibility.
- **Low:** The family argument is inconsistent between create and delete paths:
  - create uses `fam('4')` / `fam('6')`
  - destroy uses `fam(4)` / `fam(6)`
  - This may work if `Rule.fam()` normalizes both, but the inconsistency is unnecessary in firewall code and can become a subtle bug if the helper is stricter than expected.
- **Low:** There are a few style/maintainability regressions in `control/Block.js`, for example `const rawRules= [];` and several extra blank lines around the new bypass-chain blocks. Small, but this file is already complex and benefits from strict consistency.

### 💡 Suggestions
- Make bypass-chain creation idempotent. For example:
  - check for existence first, or
  - tolerate `-N` failure and always follow with `-F`, or
  - add an `ensureChain()` helper in the iptables control layer and use that everywhere.
- Do the same for deletion: make flush/delete resilient to partial state so uninstall can safely clean up after interrupted installs.
- Consider moving the bypass-chain lifecycle closer to where `byPassChain` is consumed, or formalize it as a first-class helper/API so callers cannot accidentally use `byPassChain` without provisioning the backing chain.
- Add de-duplication protection for the raw terminal rule inside the bypass chain, especially if policy application can be retried.
- Standardize `Rule.fam(...)` usage to one representation across create and destroy paths.
- Add tests for:
  - re-applying the same category block/disturb policy twice
  - partial install failure followed by retry
  - uninstall after partial state
  - create/destroy for both IPv4 and IPv6

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8930  
Head SHA: `7ae32fbca58a185d7bfcf81c95912141a700844e`  
Checked at: 2026-04-15 10:01:13 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a substantial step forward from the previous extra-time approach. Moving from “pause the blocking rule and attach a shadow policy” to an explicit `bypass` action is directionally much cleaner.
- The main design choices are solid:
  - preserve the original blocking/disturb rules
  - create a separate temporary bypass rule with `affectedPids`
  - let app-time quota drive whether that bypass is enforced or unenforced
- `AccessRequestManager` is better structured now:
  - `getAppsFromPolicy()` centralizes app extraction for single-target and multi-target rules
  - `findMatchingTimeLimitRules()` returning `{ bypassPolicies, blockOrDisturbPolicies }` is more expressive than returning one mixed array
  - approval flow now handles both time-limit rules and non-time-limit block/disturb rules in one model
- `AppTimeUsageManager` support for `onQuotaReached === "unenforce"` is a good extension. That makes the bypass lifecycle match the intended “temporary allowance until quota is consumed” behavior.
- Adding `AccessRequestSensor` hooks on `Policy:Activated` / `Policy:Deactivated` is the right idea so bypass rules can track evolving underlying policies.

- No linked issue found.

### ⚠️ Issues found
- **High:** `alarm/Policy.js` has a typo in `compareFields`:
  - `"affactedPids"`

  while the rest of the PR consistently uses `affectedPids`. That means policy comparison/update logic will not actually compare this field correctly, which is especially important because bypass policies are updated in place as affected rules change.
- **High:** `sensor/AccessRequestSensor.js` iterates tags incorrectly:
  - `const tags = Object.values(this.tagManager.tags);`
  - `for (let uid in tags) { ... }`

  Here `uid` is the array index, not the tag UID. That means `AccessRequestManager.getUserRelatedTags(uid)` is being called with the wrong identifier, so bypass-policy updates on activation/deactivation are very likely broken.
- **High:** `control/Bypass.js` references `categoryUpdater` here:
  - `categoryUpdater.getBfCategoryName(target)`

  but there is no import or definition for `categoryUpdater` in this file. That is a runtime failure path when `policy.useBf` is involved.
- **Medium:** `PolicyManager2.unenforce()` now contains auto-delete timer logic specifically for bypass policies, but it uses `pid` without defining it in that scope:
  - `await this.deletePolicy(pid);`
  - `delete this.enabledTimers[pid];`

  Unless `pid` is already destructured earlier in the omitted part of the method, this is fragile and worth double-checking. In this diff, it reads like a likely scoping mistake.
- **Medium:** `matchApp()` no longer checks action type directly and no longer checks schedule with `isInSchedule()`:
  - action filtering moved downstream
  - schedule filtering commented out

  Some of that is intentional, but it also broadens what counts as a “match” and pushes more responsibility to callers. That makes the contract less obvious and easier to misuse elsewhere.
- **Medium:** In `api/routes/time_limits.js`, this line is suspicious:
  - `const { block } = require('../../control/Block.js');`

  but `block` is not used anywhere in the diff. That is minor by itself, but it suggests this route changed in a piecemeal way and may still have stale assumptions.
- **Low:** `AccessRequestManager.getRequest()` now silently deletes `note`/`mac` when their stored value is the string `"null"`. That is practical, but it is also papering over bad serialization upstream instead of fixing the source.
- **Low:** The bypass feature adds quite a bit of iptables chain plumbing (`FW_<pid>_BYPASS`) across many rule setup paths. The general direction is fine, but this expands the surface area significantly and really needs regression coverage to avoid rule-order surprises.

### 💡 Suggestions
- Fix the `affactedPids` typo immediately; that one is easy and important.
- In `AccessRequestSensor`, iterate actual tag objects/UIDs correctly. Right now the user/tag matching logic looks wrong because it uses array indexes.
- Add the missing `categoryUpdater` import in `control/Bypass.js`, or remove that branch until it is wired correctly.
- Recheck `PolicyManager2.unenforce()` for `pid` scoping in the new bypass auto-delete timer block.
- Consider restoring clearer filtering boundaries in `matchApp()` or documenting that it is now only a partial predicate and action/schedule filtering must happen elsewhere.
- Add regression tests for:
  - single-app block rule + temporary bypass
  - multi-target block rule + shared bypass
  - bypass policy quota reached → unenforce
  - underlying blocked policy activated/deactivated while bypass exists
  - BF-set category bypass path
  - policy equality/update behavior with `affectedPids`

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8930  
Head SHA: `1dde2189da0da687cf1ed024f368a44c6655bad1`  
Checked at: 2026-04-17 20:16:00 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision fixes several of the earlier correctness problems:
  - `affectedPids` is now spelled consistently in `Policy.js`
  - `AccessRequestSensor` now iterates `this.tagManager.tags` directly instead of iterating array indexes
  - `Bypass.js` now imports and constructs `CategoryUpdater`
  - the bypass auto-delete timer in `PolicyManager2.unenforce()` now uses `policy.pid` instead of an undefined `pid`
- The overall design still looks directionally solid:
  - keep original block/disturb rules active
  - add a temporary `bypass` policy tied to `affectedPids`
  - let `AppTimeUsageManager` enforce/unenforce the bypass based on quota
- Adding `active_bypass_policy` to Redis migration/cleanup paths is a good follow-through. That reduces the chance of stale bypass policy indices lingering after upgrades or cleanup.

- No linked issue found.

### ⚠️ Issues found
- **High:** `AccessRequestManager.approveRequest()` still has a broken multi-app path for `req.app` in `policy:<id>` format. Inside the `affectedPids.size > 0` branch it does:
  - `apps = getAppsFromPolicy(policy);`

  but `apps` is not declared there, and later the code builds a different `const apps = Array.from(appset).sort();`. In practice this means the `policy:<id>` branch never populates `appset`, so `apps.length` stays `0` and bypass creation/update for multi-target block rules will fail or behave unpredictably.
- **High:** `api/routes/time_limits.js` still contains this line:
  - `const apps = fs.accessRequestManager.getAppsFromRule(policy);`

  That looks like a runtime bug. `fs` is Node’s filesystem module, and this PR adds `getAppsFromPolicy()` to `AccessRequestManager`, not `getAppsFromRule()` to `fs.accessRequestManager`. Unless there is unseen local monkey-patching, this path will throw when building notification text for `policy:<id>` requests.
- **Medium:** `matchApp()` still has schedule checking commented out:
  - `// if (!isInSchedule(policy, includeNonTimeLimitRules)) return false;`

  That broadens matching semantics significantly. It may be intentional for access-request discovery, but as written the helper name suggests a full match predicate while it now ignores one of the original rule-eligibility dimensions.
- **Medium:** In `PolicyManager2`, bypass-chain creation/destruction is spread across many rule setup paths in `Block.js` plus separate bypass enforcement in `Bypass.js`. The design can work, but the rule lifecycle is now fairly complex:
  - create bypass chain on enforce
  - insert RETURN rules by tag/interface
  - flush/delete chains on unenforce
  - also bypass dnsmasq categories

  That increased coupling really needs regression coverage because ordering bugs here will be subtle and traffic-specific.
- **Low:** `AccessRequestManager.getRequest()` still strips string `"null"` values for `note` and `mac`. Practical, but it is masking upstream serialization quality issues rather than fixing the source.

### 💡 Suggestions
- Fix the `policy:<id>` app extraction path in `approveRequest()` first. The simplest repair is to declare/fill one shared `apps` array explicitly instead of mixing an undeclared assignment with a later `const apps = Array.from(appset)`.
- In `api/routes/time_limits.js`, replace the bad `fs.accessRequestManager.getAppsFromRule(policy)` call with the real helper you added (`getAppsFromPolicy(policy)`), ideally imported directly.
- Either restore schedule filtering in `matchApp()` or rename/document the helper so callers understand it is now a partial predicate.
- Add focused tests for:
  - single-target bypass creation
  - `policy:<id>` multi-target bypass creation
  - bypass notification text for multi-app policies
  - `affectedPids` updates when underlying policies activate/deactivate
  - bypass chain creation/cleanup symmetry for block vs disturb rules

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8930  
Head SHA: `78619695698914e420ce116d099ef7004f9328d1`  
Checked at: 2026-04-17 22:01:04 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is still directionally better than the earlier “disable original rule + shadow policy” approach.
- The core model is sensible:
  - keep the original block/disturb policies intact
  - create a temporary `bypass` policy with `affectedPids`
  - let `AppTimeUsageManager` enforce/unenforce the bypass based on quota
- A few previously obvious problems are fixed in this revision:
  - the bad `fs.accessRequestManager.getAppsFromRule(...)` reference is gone
  - `affectedPids` is now consistently included in `Policy` normalization/comparison
  - bypass policies are added to cleanup/migration paths via `active_bypass_policy`
  - `AccessRequestSensor` now iterates actual tag IDs instead of array indexes

- No linked issue found.

### ⚠️ Issues found
- **High:** `api/routes/time_limits.js` still calls `accessRequestManager.getAppsFromPolicy(policy)`, but `getAppsFromPolicy` is added as a module-level export, not as an instance method on `class AccessRequestManager`. With the diff shown, that path will still fail at runtime for `policy:<id>` request formatting unless there is unseen prototype wiring elsewhere.
- **High:** `sensor/AccessRequestSensor.js` has the same API-shape problem. It uses:
  - `AccessRequestManager.getUserRelatedTags(uid)`
  - `AccessRequestManager.getAppsFromPolicy(policy)`
  - `AccessRequestManager.getInstance()`

  That only works if every caller treats the import as a module namespace object. The interface is inconsistent enough that this is fragile and easy to break.
- **High:** multi-target policies are still excluded from the normal matching/listing flow. `matchTarget()` returns `false` whenever `policy.targets.length > 1`, so `findMatchingTimeLimitRules(userId, "all", ...)` will not discover those rules during the active-policy scan. That means the general `/time_limits` listing can still miss the exact multi-target rules this PR is trying to support.
- **Medium:** schedule filtering remains commented out in `matchApp()`:
  - `// if (!isInSchedule(policy, includeNonTimeLimitRules)) return false;`

  That materially changes the meaning of “matching”. If intentional, it should be split from “currently effective” logic; otherwise this risks surfacing inactive rules and attaching bypass behavior to policies outside their active schedule.
- **Medium:** the bypass lifecycle is now spread across `PolicyManager2`, `Block.js`, `Bypass.js`, dnsmasq handling, and `AccessRequestSensor`. The design can work, but the blast radius is now large enough that subtle ordering/cleanup bugs are likely without targeted regression coverage.
- **Low:** `getRequest()` silently strips `"null"` strings from `note` and `mac`. That is pragmatic cleanup, but it hides an upstream serialization problem rather than fixing the source.

### 💡 Suggestions
- Make the helper surface consistent:
  - either add `getAppsFromPolicy()` / `getUserRelatedTags()` as real class or instance methods
  - or import the standalone helpers directly where they are used
- Rework the `matchTarget()` / `findMatchingTimeLimitRules()` split so multi-target policies are still included in the `"all"` listing path, while specific `policy:<id>` requests continue to work.
- Restore schedule filtering, or explicitly separate:
  - “policy is related to this app/user”
  - “policy is currently effective”
- Add focused tests for:
  - `/time_limits` GET with multi-target rules
  - `policy:<id>` request creation and notification formatting
  - bypass creation/update for multi-app policies
  - `affectedPids` maintenance after `Policy:Activated` / `Policy:Deactivated`
  - iptables/dnsmasq bypass creation and cleanup symmetry

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8930  
Head SHA: `54c3511edf47985e6fd0b4c0da405d51fb39b0ee`  
Checked at: 2026-04-17 23:01:04 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision fixes the earlier helper API inconsistency:
  - `api/routes/time_limits.js` now imports `getAppsFromPolicy` directly
  - `sensor/AccessRequestSensor.js` now imports `getAccessRequestManager`, `getUserRelatedTags`, and `getAppsFromPolicy` explicitly instead of mixing module/object/class access styles
- The overall approach is still a meaningful design improvement over the old shadow-policy flow:
  - original block/disturb rules stay intact
  - temporary access is modeled as a `bypass` policy with `affectedPids`
  - quota exhaustion flips that bypass back off through `AppTimeUsageManager`
- The supporting plumbing is reasonably complete:
  - `affectedPids` is normalized in `Policy`
  - bypass policies are tracked in Redis cleanup/migration
  - `HostManager` excludes bypass rules from normal policy rule handling
  - `AccessRequestSensor` now listens for policy activation/deactivation and updates related bypass policies

- No linked issue found.

### ⚠️ Issues found
- **High:** multi-target rules are still excluded from the general matching path. `matchTarget()` returns `false` whenever `policy.targets.length > 1`, so `findMatchingTimeLimitRules(userId, "all", ...)` will not return those rules during the active-policy scan. That means `/api/routes/time_limits.js` can still fail to list multi-app rules unless they are addressed through the special `policy:<id>` path.
- **High:** the quota lookup for non-time-limit multi-target rules in `api/routes/time_limits.js` is still inconsistent with how bypass keys are stored. `approveRequest()` stores bypass policies under:
  - `bypass:extraTime:${userId}:${apps.join(',')}`

  for the full app set, but the GET path looks them up with:
  - `getBypassPolicyKey(userId, [appQuotaInfo.app])`

  For a multi-target rule represented as `policy:<pid>`, that lookup uses a single synthetic app token instead of the full app list, so the bypass quota/time-used state for multi-app rules will not be found correctly.
- **Medium:** schedule filtering is still commented out in `matchApp()`:
  - `// if (!isInSchedule(policy, includeNonTimeLimitRules)) return false;`

  That materially changes the semantics of “matching” and risks returning inactive rules as candidates for access-request handling or UI display.
- **Medium:** the bypass lifecycle is now spread across `PolicyManager2`, `Block.js`, `Bypass.js`, `dnsmasq`, and `AccessRequestSensor`. The architecture can work, but it is now complex enough that ordering and cleanup bugs are likely without targeted regression coverage, especially around rule reenforcement and expire/delete timing.
- **Low:** `getRequest()` still silently strips `"null"` strings from `note` and `mac`. That is harmless in practice, but it hides an upstream serialization issue instead of fixing the source.

### 💡 Suggestions
- Fix the multi-target discovery path first. The current `matchTarget()` behavior makes the new multi-app support incomplete because the general listing code still cannot reliably see those rules.
- In `api/routes/time_limits.js`, use the full app set from the rule when looking up bypass state for non-time-limit multi-target policies, not `[appQuotaInfo.app]`.
- Either restore schedule filtering in `matchApp()` or split the logic into separate helpers:
  - “policy is related to this app/user”
  - “policy is currently effective”
- Add regression tests for:
  - `/time_limits` GET with multi-target rules
  - bypass quota lookup for multi-app non-time-limit rules
  - activation/deactivation updates to `affectedPids`
  - bypass expire/delete behavior and cleanup of iptables + dnsmasq state

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8930  
Head SHA: `c2ecc309302ebbbdd7fcbc2e8924898d22dc83c4`  
Checked at: 2026-04-18 07:16:05 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision fixes the earlier bypass-key lookup bug in `api/routes/time_limits.js` by using the full sorted app set:
  - `accessRequestManager.getBypassPolicyKey(userId, appQuotaInfo.apps)`
- The helper import surface is now consistent across the touched files:
  - `getAppsFromPolicy` is imported directly where needed
  - `AccessRequestSensor` no longer mixes module-namespace and instance-style access
- The general direction still makes sense:
  - preserve original block/disturb rules
  - represent temporary allowance as a `bypass` policy tied to `affectedPids`
  - let `AppTimeUsageManager` flip the bypass off when quota is exhausted
- Supporting cleanup/plumbing is reasonably thorough:
  - `affectedPids` participates in policy normalization/equality
  - `active_bypass_policy` is included in migration and cleanup
  - `HostManager` excludes bypass policies from normal policy-rule buckets

- No linked issue found.

### ⚠️ Issues found
- **High:** multi-target rules are still effectively invisible to the normal discovery path. `matchTarget()` still returns `false` whenever `policy.targets.length > 1`, so `findMatchingTimeLimitRules(userId, "all", ...)` will not surface those rules in the general active-policy scan. That means the `/time_limits` list can still miss multi-app rules even though later code tries to display and resolve them.
- **Medium:** schedule filtering remains disabled in `matchApp()`:
  - `// if (!isInSchedule(policy, includeNonTimeLimitRules)) return false;`

  That changes the meaning of “matching” from “currently effective rule” to “related rule regardless of schedule”. If intentional, it should be split into a differently named predicate; otherwise inactive rules may leak into request handling and UI state.
- **Medium:** the bypass lifecycle remains spread across several components:
  - `AccessRequestManager`
  - `AppTimeUsageManager`
  - `PolicyManager2`
  - `Block.js`
  - `Bypass.js`
  - `dnsmasq`
  - `AccessRequestSensor`

  The architecture can work, but the state machine is now complicated enough that ordering and cleanup regressions are a real risk without targeted tests.
- **Low:** `AccessRequestSensor.js` still reintroduces a lot of duplicated bootstrap/import logic in one patch:
  - lock setup
  - redis client
  - policy manager
  - constants
  - event listeners

  It is workable, but it makes the sensor harder to reason about and raises maintenance cost for what is essentially synchronization glue.
- **Low:** `getRequest()` still strips `"null"` strings from `note` and `mac`. Practical, but it hides an upstream serialization bug instead of fixing the producer.

### 💡 Suggestions
- Fix `matchTarget()` / `findMatchingTimeLimitRules()` so multi-target rules can participate in the `"all"` listing path, not only in the explicit `policy:<id>` path.
- Either restore schedule checks in `matchApp()` or split the logic into two helpers:
  - relation/membership check
  - active/in-schedule check
- Add focused regression tests for:
  - `/time_limits` GET with multi-target rules
  - bypass quota lookup for multi-app policies
  - `Policy:Activated` / `Policy:Deactivated` updates to `affectedPids`
  - bypass expire/delete cleanup across iptables and dnsmasq
- Consider moving bypass-policy synchronization logic into a more central policy service instead of leaving `AccessRequestSensor` responsible for cross-component consistency.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8930  
Head SHA: `2cfade2d53420afd3527834782373da13d32485b`  
Checked at: 2026-04-18 07:30:49 CST

PR #8931, opened 2026-04-15
suppress event logging on flow:switch_accounting event

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The change is minimal and low-risk in terms of functional behavior.
- Adding `suppressEventLogging: true` to both `MSG_FLOW_SWITCH_ACCOUNTING` emissions is a sensible way to reduce noisy event logs for high-frequency accounting messages.
- The patch is applied symmetrically to both traffic directions, so behavior stays consistent for `mac1 -> mac2` and `mac2 -> mac1`.

- No linked issue found.

### ⚠️ Issues found
- **Low:** This PR only suppresses logging at the event emission site. If any downstream consumer depends on these events being visible in standard logs for debugging/accounting verification, that visibility will now disappear without an alternative trace path.
- **Low:** There is no accompanying guard, comment, or test explaining why `MSG_FLOW_SWITCH_ACCOUNTING` is safe to hide while other emitted message types are not. That makes the intent a bit opaque for future maintainers.

### 💡 Suggestions
- Add a short comment near the emit path explaining that switch accounting events are high-volume and intentionally excluded from routine event logs.
- If these events are sometimes useful during troubleshooting, consider ensuring there is still a debug-level or opt-in tracing path elsewhere rather than removing observability entirely from normal logs.
- A tiny regression test or log-behavior assertion would help prevent accidental re-enabling later if log noise is the issue being fixed.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8931  
Head SHA: `f5f150b581a3f8d51d2e29f4b0e7cfcda0179130`  
Checked at: 2026-04-15 11:31:27 CST

MelvinTo (approved):

PR #8935, opened 2026-04-15
set isoGID as string in processApcBlockFlowMessage

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Converting `isoGID` to `String(msg.gid)` in `sensor/APCMsgSensor.js` is a sensible normalization step. It avoids mixed numeric/string storage in Redis records and makes downstream comparisons more predictable.
- The new test covers the intended regression well: it exercises a real APC block flow path, flushes the audit buffer, reads the stored Redis payload, and asserts that `isoGID` is present and serialized as a string.
- The test cleanup is thoughtful:
  - removing the created host fixture
  - deleting the Redis audit keys
  - wrapping cleanup in `finally`
- Stubbing `ruleStatsPlugin` in `flushApcAuditBuffer()` is a practical way to isolate the behavior under test.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** The production change normalizes only non-null/defined `gid`, but it does not validate shape. Values like objects, arrays, or booleans would now also be coerced into strings and persisted. If `gid` is expected to be numeric or numeric-like, this should be constrained more tightly.
- **Medium:** The new test uses polling with fixed sleeps in `flushApcAuditBuffer()` (`setTimeout(..., 50)` loop up to 5 seconds). That is workable, but still makes the test timing-sensitive and slower than necessary. If plugin initialization changes, this can become flaky.
- **Low:** The test only verifies the happy path for `gid: 99`. It does not cover edge cases that motivated the normalization:
  - `gid: 0`
  - `gid: '99'`
  - `gid: null`
  - missing `gid`
  Without those, it is still easy to regress falsy or already-string values.
- **Low:** `APC_TEST_PEER_MAC` is described as “not a real host,” but the test writes reverse-direction audit state for it and cleans it up manually. That is fine, but it hints that the fixture behavior is a little broader than the test title suggests.

### 💡 Suggestions
- If `gid` is supposed to be numeric, consider validating it before stringifying, rather than accepting any defined value.
- Add a couple of focused tests for edge cases:
  - `gid: 0` should be preserved as `'0'`
  - missing/null `gid` should not create `isoGID`
  - string `gid` should remain stable
- If possible, replace the polling-based buffer flush with a more deterministic hook or explicit plugin initialization in the test setup.
- A brief comment in `APCMsgSensor.js` explaining why `isoGID` is normalized to string would help future maintainers.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8935  
Head SHA: `b60da8288f8b45380d184b9d68ba2cec6bd75abd`  
Checked at: 2026-04-15 15:00:58 CST

jasonlyc (approved):

PR #8937, opened 2026-04-15
Improve category blocking coverage for supported apps

yijin-png:
Extend category blocking to also cover supported apps that require richer app-level blocking behavior.

### DIFF
--- a/alarm/PolicyManager2.js
+++ b/alarm/PolicyManager2.js
@@ -1780,6 +1780,53 @@ class PolicyManager2 {
             });
           }
         }
+
+        // --- Derive app-level block rules from category ---
+        if (isBlockOrdisturb) {
+          const allDerivedAppTargets = [];
+          for (const catTarget of targets) {
+            const derived = await this._getDerivedAppTargetsForCategory(catTarget);
+            allDerivedAppTargets.push(...derived);
+          }
+
+          if (!_.isEmpty(allDerivedAppTargets)) {
+            log.info(`Policy ${pid}: derive app targets from category:`, allDerivedAppTargets);
+
+            // DNS blocking (append to existing dnsmasq entries)
+            if (direction !== "inbound" && !localPort && !remotePort) {
+              await domainBlock.blockCategory({
+                pid, scope, categories: allDerivedAppTargets, intfs, guids,
+                action, tags, parentRgId, seq, wanUUID, routeType,
+                append: true
+              });
+            }
+
+            // TLS host sets
+            if (platform.isTLSBlockSupport() || platform.isUdpTLSBlockSupport()) {
+              for (const dt of allDerivedAppTargets)
+                tlsHostSets.push(categoryUpdater.getHostSetName(dt));
+            }
+
+            // Activate + ipset (dnsmasq_only=false, no bloom filter)
+            for (const dt of allDerivedAppTargets) {
+              await categoryUpdater.activateCategory(dt);
+              await categoryUpdater.updateCategoryState(dt, devOpts, pid, false, isBlockOrdisturb, true);
+              remoteSets.push({
+                remoteSet4: categoryUpdater.getAggrIPSetName(dt),
+                remoteSet6: categoryUpdater.getAggrIPSetNameForIPV6(dt)
+              });
+              // connSet for signature-detected connections (e.g. roblox-sig blockType=connection)
+              connSets.push({
+                connSet4: categoryUpdater.getConnectionIPSetName(dt),
+                connSet6: categoryUpdater.getConnectionIPSetNameForIPV6(dt)
+              });
+            }
+
+            // Push into targets for post-switch TLS activation
+            targets.push(...allDerivedAppTargets);
+          }
+        }
+
         remoteTupleCount = 2;
         break;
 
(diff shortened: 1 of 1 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The intent makes sense: when a category policy is blocked, deriving app-level targets with ACL support can close coverage gaps between category rules and app-specific enforcement paths.
- The change is implemented symmetrically in both enforce and unenforce flows, which is better than adding derived targets on apply without corresponding cleanup on removal.
- Reusing existing helpers (`domainBlock`, `categoryUpdater`, TLS host sets, connection ipsets) keeps the implementation aligned with the current policy pipeline instead of inventing a parallel mechanism.
- `_getDerivedAppTargetsForCategory()` keeps the cloud-config lookup encapsulated, which is cleaner than inlining that filter logic in both call sites.

- No linked issue found.

### ⚠️ Issues found
- **High:** The derived targets are appended directly into `targets` with `targets.push(...allDerivedAppTargets)` in both enforce and unenforce paths, but there is no deduplication.
  - If multiple categories map to the same app, or if the original `targets` already contain one of those `TLX-fw-*` values, the code can emit duplicate DNS/TLS/ipset work and potentially duplicate downstream rule handling.
  - For policy orchestration code, idempotency matters a lot.
- **High:** `_getDerivedAppTargetsForCategory()` fetches and parses Redis cloud config on every category in the loop.
  - In the current structure, a single policy with multiple category targets does repeated `getAsync + JSON.parse` calls for the same config inside one enforcement pass.
  - That is unnecessary overhead in a hot path and makes behavior more brittle when Redis/config is transient.
- **Medium:** The new logic mutates `targets` during processing, after already iterating the original categories.
  - Even if it currently happens late enough not to recurse immediately, this makes the control flow harder to reason about and increases the risk of future regressions if later code assumes `targets` still represents only the original policy targets.
- **Medium:** `_getDerivedAppTargetsForCategory()` assumes every derived app target should be materialized as `TLX-fw-${app}`.
  - That naming convention is hardcoded here, with no validation that the corresponding category updater / host set / domain block layers actually recognize every emitted target.
  - If cloud config contains stale or malformed app keys, this path can generate invalid derived targets silently.
- **Medium:** The added log lines dump full derived target arrays:
  - `derive app targets from category`
  - `unenforce derived app targets`
  This is not as severe as logging full alarm bodies, but it can still get noisy and expose too much internal targeting detail in routine policy logs.
- **Low:** The new helper only checks `features.acl === true` and `category === <category>`, but it does not guard against empty/missing app names before building `TLX-fw-${app}`.
- **Low:** Because derived app targets are folded into the same TLS/remote/conn-set collections as first-class targets, debugging later policy behavior may get harder unless there is some marker distinguishing original targets from derived ones.

### 💡 Suggestions
- Deduplicate `allDerivedAppTargets` before using them, and avoid mutating `targets` in place unless a later stage strictly requires it.
- Load and parse the cloud config once per enforcement/unenforcement pass, then derive app targets from that cached object instead of re-reading Redis for every category.
- Consider keeping derived targets in a separate collection all the way through the pipeline rather than merging them back into `targets`.
- Validate derived app names before constructing `TLX-fw-*` targets, or at least skip obviously invalid entries.
- Tone down the info logs or log only counts/examples rather than the complete target list.
- Add tests for:
  - two categories deriving the same app target
  - original targets already containing a derived app target
  - malformed / missing `appConfs`
  - enforce → unenforce symmetry for derived targets

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8937  
Head SHA: `65d69c76bece1e24b3fb6330f42257f82fd3ce09`  
Checked at: 2026-04-15 19:30:59 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The overall intent makes sense: when a category policy is enforced for `block` / `disturb`, deriving matching `TLX-fw-*` app targets can help keep category-level behavior aligned with app-level ACL handling.
- The PR correctly avoids applying Bloom-filter handling to derived small extended target lists:
  - `bfTargets = targets.filter(t => !categoryUpdater.isSmallExtendedTargetList(t))`
  - guarded `activateCategory(categoryUpdater.getBfCategoryName(target))`

  That is the right separation.
- Adding the same derivation logic to both enforce and unenforce paths is important. At least directionally, the cleanup path is not forgotten.

- No linked issue found.

### ⚠️ Issues found
- **High:** `_getDerivedAppTargetsForCategory()` still does a Redis fetch + `JSON.parse` every time it is called, and this PR now calls it inside loops in both enforce and unenforce paths. For multi-category policies, that means repeated remote reads/parses in hot policy application code. This was already a key concern and still stands.
- **High:** `targets.push(...derivedAppTargets)` mutates the working `targets` array in place after it may already contain original category targets. There is still no deduplication. If:
  - the same derived app is returned for multiple categories, or
  - enforce / unenforce logic is re-entered with overlapping targets,

  the same target can be appended multiple times and processed repeatedly.
- **Medium:** The `TLX-fw-${app}` construction is still hardcoded in `_getDerivedAppTargetsForCategory()`. That keeps policy behavior tightly coupled to one target encoding convention. If the extended-target format changes, this logic silently breaks.
- **Medium:** The derivation logic is duplicated in two places:
  - enforce path
  - unenforce path

  Same loop, same append behavior, same logging shape. That duplication increases drift risk and makes future fixes easy to miss in one half.
- **Medium:** Because derived targets are appended back into the same `targets` array, downstream code no longer has a clean distinction between:
  - original category targets
  - derived app targets

  That makes reasoning about later operations harder and increases the risk of category-specific code accidentally touching app-derived targets in places not guarded by `isSmallExtendedTargetList`.
- **Low:** The new `log.info` calls may be noisy for routine policy operations, especially if a category maps to many apps.

### 💡 Suggestions
- Cache the parsed `appConfs` instead of fetching/parsing Redis for each category inside each enforce/unenforce pass.
- Deduplicate derived targets before appending, or better, keep `originalTargets` and `derivedTargets` as separate arrays and combine only where necessary.
- Extract the category→derived-app-target resolution into one shared helper used by both enforce and unenforce paths.
- Avoid hardcoding `TLX-fw-${app}` here if there is already a canonical helper for extended target generation.
- Consider preserving separate variables for:
  - original category targets
  - derived app targets
  - BF-eligible category targets

  That will make the later control flow much easier to audit.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8937  
Head SHA: `a67931961f4ff70d4e1d8482d1e9de42368939d1`  
Checked at: 2026-04-16 16:31:44 CST

PR #8938, opened 2026-04-15
Updated Guardian.js (200ms delay + per-cycle cap of 20)

hhirokazu:
Added 200ms delay and const MAX_OPS_PER_CYCLE = 20 on the while loop.

### DIFF
--- a/sensor/Guardian.js
+++ b/sensor/Guardian.js
@@ -758,7 +758,8 @@ module.exports = class {
         return;
       }
       const encryptMessageAsync = util.promisify(cw.getCloud().encryptMessage).bind(cw.getCloud());
-      while (this.isSocketConnected()) {
+      const MAX_OPS_PER_CYCLE = 20;
+      while (this.isSocketConnected() && count < MAX_OPS_PER_CYCLE) {
         const op = await this.dequeueOp();
         if (!op)
           break;
@@ -782,6 +783,7 @@ module.exports = class {
             mspId: mspId
           });
           count++;
+          await delay(200);
         } catch (err) {
           log.error(`Failed to sync op to msp`, err);
           // push back to the queue


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Adding a per-cycle cap in `sensor/Guardian.js` is a reasonable direction. Bounding the number of dequeued ops prevents one sync loop from monopolizing the socket/worker indefinitely when backlog is large.
- The small `await delay(200)` between sends is a pragmatic rate-smoothing measure. If the MSP side or socket path is sensitive to bursts, this should reduce spike behavior.
- Keeping the limit local (`MAX_OPS_PER_CYCLE = 20`) makes the intent easy to understand.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** The new throttle changes throughput significantly but does not show how the remaining queue gets resumed promptly after the capped cycle exits.
  - Before, the loop drained until the queue was empty or the socket disconnected.
  - Now it stops after 20 ops even if the socket is still healthy.
  - If there is no reliable immediate re-trigger after exit, queued ops may sit longer than expected.
- **Medium:** `await delay(200)` adds a hardcoded ~4 second worst-case delay for a full 20-op batch.
  - That may be acceptable, but it is a meaningful behavior change for sync latency.
  - Without making the delay configurable or documenting the intended backpressure target, it is hard to tune or validate.
- **Low:** `MAX_OPS_PER_CYCLE` is defined inside the function. If this value is expected to evolve with production load, a config-driven limit would be easier to operate and test.
- **Low:** The diff only increments `count` on successful sync. If failures are re-queued and the loop continues, one bad op pattern could still cause repeated work inside the same cycle until another condition breaks it. That may be intentional, but it is worth checking.

### 💡 Suggestions
- Confirm there is a dependable re-entry path after the capped loop exits; otherwise the new cap may trade burst reduction for stalled backlog processing.
- Consider making both the batch size and inter-op delay configurable instead of hardcoding `20` and `200ms`.
- Add a debug log when the loop exits because `MAX_OPS_PER_CYCLE` was reached, so it is easier to diagnose queue backpressure in production.
- Add/extend tests around:
  - backlog > 20 ops
  - repeated reconnect / re-trigger behavior
  - failure + requeue behavior under the new capped loop

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8938  
Head SHA: `7dc59dcbfbe8f20f0c6498fa4d168c887178a28d`  
Checked at: 2026-04-16 04:45:50 CST

jasonlyc (approved):

PR #8939, opened 2026-04-16
fix a racing condition on category block

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Changing `scripts/main-start` to check for the existing ICMPv6 redirect rule before inserting it is a good idempotency fix. That avoids piling up duplicate `ip6tables` entries across restarts.
- The new `Rule.icmp6()` helper in `net2/Iptables.js` is a useful abstraction. It makes common ICMPv6 rule construction less error-prone than passing raw protocol/type flags around manually.
- Renaming the init path in `BlockControl` to `callInitScript()` makes the intent a bit clearer: that method now really just runs the setup script and does not claim to fully manage the initialization lifecycle.
- In `IpsetControl`, checking whether the swap set actually exists before issuing `swap` / `destroy` is directionally safer than blindly assuming the `_swp` set was created.

- No linked issue found.

### ⚠️ Issues found
- **High:** Starting `BlockControl` in `'initializing'` state from the constructor changes lifecycle semantics globally:
  - previously it entered idle immediately
  - now rules queued before `callInitScript()` are preserved

  That may be intentional, but it also means any caller that assumes a fresh `BlockControl` is idle can now silently accumulate queued work before initialization has actually started. For firewall orchestration code, hidden state-machine changes like this are risky.
- **High:** `callInitScript()` no longer records `initTS`, but `finishInitialization()` still reports elapsed time from `sysManager.startTS` only. That means the code lost the ability to measure actual initialization duration and now logs a broader service-start metric instead. More importantly, it suggests the initialization-state refactor is incomplete rather than fully thought through.
- **Medium:** The `IpsetControl` fallback `this.existingSets = fromInitialization ? new Set() : previousSets || this.existingSets;` is still brittle.
  - During initialization, seeding from an empty set changes behavior significantly.
  - If `previousSets` lookup fails or is stale, the control flow can still diverge between expected and actual kernel state.
  - This is safer than crashing, but not enough for something that swaps/destroys firewall sets.
- **Medium:** In the swap logic, when `swapSetName` is missing, the code logs an error and then falls through into the cleanup branch for `c_*` sets:
  - that can still flush/destroy the original set even though the intended swap did not happen
  - so the missing-swap-set case remains risky
- **Low:** `Rule.icmp6()` mixes field mutation and module push in a compact style:
  - `this.proto = [...]`
  - then `this.modules.push(...)`
  - then `return this`

  It works, but the formatting and inline `return` make it harder to read and maintain than the rest of the builder methods.
- **Low:** `PolicyManager.js` switched from `.pro('icmpv6').opt('--icmpv6-type', 'redirect')` to `.icmp6('redirect')`, which is cleaner, but only for this one rule. If the helper is now the preferred path, mixed styles across the codebase may become inconsistent.

### 💡 Suggestions
- Re-check the `BlockControl` state machine end-to-end. If constructor-time `'initializing'` is required, document the invariant clearly and make sure all callers still transition explicitly and safely.
- If preserving pre-init rules is the goal, consider keeping a separate pre-init queue rather than changing the initial global state of `BlockControl`.
- In `IpsetControl`, make the missing-swap-set path fail safer: avoid flushing/destroying the original set when the replacement set was never created.
- Add tests for:
  - rules queued before `callInitScript()`
  - repeated init/finish sequences
  - missing `_swp` set during swap cleanup
  - idempotent ICMPv6 redirect insertion
- Consider using `icmp6()` consistently anywhere similar ICMPv6 rules are built.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8939  
Head SHA: `9a53a485996d30e5c734ff395a83a24e935f3057`  
Checked at: 2026-04-16 11:45:55 CST

jasonlyc (approved):

PR #8940, opened 2026-04-16
support multiple targets in app disturb rules

yijin-png:
Allow app disturb rules to specify multiple targets, matching the existing multi-target support in block and time-limit rules.

### DIFF
--- a/alarm/PolicyDisturbManager.js
+++ b/alarm/PolicyDisturbManager.js
@@ -91,21 +91,56 @@ class PolicyDisturbManager {
     });
   }
 
+  _getAppNameFromTarget(target) {
+    if (target && target.startsWith('TLX-fw-'))
+      return target.substring('TLX-fw-'.length);
+    return target || "";
+  }
+
+  _getAppNames(policy) {
+    const appNames = [];
+    if (!_.isEmpty(policy.targets)) {
+      for (const t of policy.targets) {
+        const name = this._getAppNameFromTarget(t);
+        if (name && !appNames.includes(name))
+          appNames.push(name);
+      }
+    }
+    const primary = policy.app_name || this._getAppNameFromTarget(policy.target);
+    if (primary && !appNames.includes(primary))
+      appNames.push(primary);
+    return appNames.length > 0 ? appNames : [""];
+  }
+
   async _registerPolicy(policy) {
     const pid = String(policy.pid);
     log.info(`Registering policy ${pid} ...`);
 
-    const appName = policy.app_name || "";
+    const appNames = this._getAppNames(policy);
     const disturbLevel = policy.disturbLevel || "";
     //set default default values
     let defaultDisturbVal = { "rateLimit": 10240, "dropPacketRate": 0, "increaseLatency": 0 };
     let disableQuic = false;
 
-    if (this._appConfValue && this._appConfValue.hasOwnProperty(appName)){
-      disableQuic = this._appConfValue[appName].disableQuic || false;
-      if (this._appConfValue[appName].hasOwnProperty(disturbLevel))
-        defaultDisturbVal = Object.assign(defaultDisturbVal, this._appConfValue[appName][disturbLevel]);
-    } else if (this._generalConfValue && this._generalConfValue.hasOwnProperty(disturbLevel)) {
+    let hasAppConf = false;
+    for (const appName of appNames) {
+      if (this._appConfValue && this._appConfValue.hasOwnProperty(appName)) {
+        hasAppConf = true;
+        // Use the strictest effective behavior when app-specific defaults differ.
+        disableQuic = disableQuic || (this._appConfValue[appName].disableQuic || false);
+        if (this._appConfValue[appName].hasOwnProperty(disturbLevel)) {
+          const appConf = this._appConfValue[appName][disturbLevel];
+          if (appConf.rateLimit != null)
+            defaultDisturbVal.rateLimit = Math.min(defaultDisturbVal.rateLimit, appConf.rateLimit);
+          if (appConf.dropPacketRate != null)
+            defaultDisturbVal.dropPacketRate = Math.max(defaultDisturbVal.dropPacketRate, appConf.dropPacketRate);
+          if (appConf.increaseLatency != null)
+            defaultDisturbVal.increaseLatency = Math.max(defaultDisturbVal.increaseLatency, appConf.increaseLatency);
+        }
+      }
+    }
+
+    if (!hasAppConf && this._generalConfValue && this._generalConfValue.hasOwnProperty(disturbLevel)) {
       defaultDisturbVal = Object.assign(defaultDisturbVal, this._generalConfValue[disturbLevel]);
     }
 
--- a/alarm/PolicyDisturbManager.js
+++ b/alarm/PolicyDisturbManager.js
@@ -18,6 +18,7 @@
 const log = require('../net2/logger.js')(__filename);
 const Policy = require('./Policy.js');
 const _ = require('lodash');
+const crypto = require('crypto');
 const sem = require('../sensor/SensorEventManager.js').getInstance();
 const rclient = require('../util/redis_manager.js').getRedisClient();
 const Message = require('../net2/Message.js');
@@ -25,6 +26,10 @@ const AsyncLock = require('../vendor_lib/async-lock');
 const Constants = require('../net2/Constants.js');
 const lock = new AsyncLock();
 const LOCK_RW = "lock_rw";
+
+function _shortHash(s) {
+  return crypto.createHash('md5').update(String(s)).digest('hex').slice(0, 8);
+}
 
 class PolicyDisturbManager {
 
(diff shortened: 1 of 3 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Extending `PolicyDisturbManager` to consider `policy.targets` for app disturb is the right direction. It aligns the disturb path with the existing multi-target policy model instead of relying only on a single `app_name` / `target`.
- `_getAppNameFromTarget()` and `_getAppNames()` make the target normalization logic more explicit and reusable than scattering `TLX-fw-` parsing inline.
- The intent behind merging multiple app configs conservatively is sensible. Using stricter effective behavior for a combined disturb policy is a safer default than accidentally weakening enforcement.
- The `PolicyManager2` comment update is small but helpful: it documents that `targets` now affect app disturb too, not just multi-category block / app time limit.

- No linked issue found.

### ⚠️ Issues found
- **High:** The fallback behavior in `_getAppNames()` returns `[""]` when no app names are found. That means `_registerPolicy()` will iterate an empty-string app name and may treat `""` as a real lookup key:
  - `_appConfValue.hasOwnProperty("")`
  - later logging / merging logic also includes it

  This is brittle and can accidentally activate config keyed by the empty string, or at minimum make the code path harder to reason about. It should return an empty array, not a sentinel empty-string app.
- **High:** The new merge strategy for multiple app configs is opinionated but not obviously correct:
  - `rateLimit` → `Math.min(...)`
  - `dropPacketRate` → `Math.max(...)`
  - `increaseLatency` → `Math.max(...)`
  - `disableQuic` → OR

  That may be “strictest”, but it is also a semantic change in how disturb profiles are composed. If two apps are combined accidentally or transiently, the resulting disturb behavior can become much harsher than either individual policy author expected. This needs explicit product-rule backing.
- **Medium:** `_getAppNameFromTarget()` still hardcodes the `TLX-fw-` prefix assumption. That was already a known concern in similar policy-target code: if target encoding changes or other derived app target formats exist, this parser will silently misclassify them instead of failing loudly.
- **Medium:** `_getAppNames(policy)` mixes names from:
  - `policy.targets`
  - `policy.app_name`
  - parsed `policy.target`

  without any precedence rule beyond de-duplication. If those fields disagree, the code now merges all of them together. That can mask malformed policy state rather than surfacing it.
- **Medium:** The new logging is fairly verbose:
  - full `appNames`
  - full `defaultDisturbVal`

  For frequently enforced policies, this may create noisy logs and make it harder to spot actual anomalies.
- **Low:** `hasOwnProperty` is used directly on config objects. In Node this usually works, but `Object.prototype.hasOwnProperty.call(...)` is safer if these config blobs are externally loaded or may have unusual prototypes.

### 💡 Suggestions
- Change `_getAppNames()` to return `[]` when nothing is found, and handle the empty case explicitly in `_registerPolicy()`.
- Define and document the merge semantics for multi-app disturb policies. If “strictest wins” is intended, it should be an explicit product rule, not just an implementation guess.
- Consider validating for conflicting policy fields:
  - `app_name`
  - `target`
  - `targets`

  If they resolve to different apps unexpectedly, log a warning or reject instead of silently merging all of them.
- Centralize app-target parsing rather than hardcoding `TLX-fw-` here again. This looks like the same class of target-format coupling that has already been risky elsewhere.
- Trim log verbosity or lower some of the new `info` logs to `debug`.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8940  
Head SHA: `d3a547fdcd4208c15ebbab233f29bf3900d695a7`  
Checked at: 2026-04-16 16:31:05 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Returning `appNames` as an empty array instead of `[""]` fixes the most obvious edge-case bug from the earlier revision. That removes the accidental empty-string config lookup path.
- Switching to `Object.prototype.hasOwnProperty.call(...)` is a small but solid hardening improvement for config-object access.
- Extending disturb handling to look at `policy.targets` is still the right direction. It brings app-disturb behavior closer to the existing multi-target policy model.
- `_getAppNameFromTarget()` / `_getAppNames()` keep the target-parsing logic localized instead of duplicating it inline in `_registerPolicy()`.

- No linked issue found.

### ⚠️ Issues found
- **High:** The multi-app merge semantics are still an unverified behavior change:
  - `rateLimit` uses the minimum
  - `dropPacketRate` uses the maximum
  - `increaseLatency` uses the maximum
  - `disableQuic` uses OR

  That may be “strictest,” but it is not obviously the correct product rule. If a policy unintentionally resolves to multiple apps, the resulting disturb behavior can become much harsher than expected.
- **Medium:** `_getAppNames(policy)` still merges names from three different sources:
  - `policy.targets`
  - `policy.app_name`
  - parsed `policy.target`

  with only de-duplication, not precedence or consistency checking. If those fields disagree, the code silently combines all of them instead of surfacing a malformed policy state.
- **Medium:** `_getAppNameFromTarget()` still hardcodes the `TLX-fw-` prefix assumption. That keeps the disturb path coupled to one specific target encoding format and will silently mis-handle future target variants.
- **Medium:** The new logging is still fairly verbose for a hot policy-registration path:
  - full `appNames`
  - full `defaultDisturbVal`

  That may be useful during rollout, but it can become noisy in normal operation.
- **Low:** The code now falls back to general config only when **no** app config is found at all. If multiple app names are present and only some have app-specific config, the result is a partial merge with no general-level fill-in for missing disturb-level fields per app. That may be intended, but it is worth being explicit about.

### 💡 Suggestions
- Define the merge rule for multi-app disturb policies explicitly in product terms. Right now the implementation is reasonable but still speculative.
- Add validation for conflicting app sources. If `app_name`, `target`, and `targets` resolve to different apps unexpectedly, log a warning or reject instead of silently merging.
- Centralize `TLX-fw-` parsing/generation if there is already a canonical helper elsewhere in the policy stack.
- Consider reducing the new `info` logs to `debug` once the rollout is verified.
- Add tests for:
  - empty/no-app policy input
  - single app via `app_name`
  - single app via `target`
  - multiple apps via `targets`
  - conflicting `app_name` vs `targets`
  - partial app config coverage with fallback behavior

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8940  
Head SHA: `41920f5b2ceada097c331cd57ed2b44744e4e1f4`  
Checked at: 2026-04-16 17:16:39 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a meaningful improvement over the earlier multi-target disturb implementation. The core direction is right:
  - split one logical disturb policy into per-target effective QoS policies
  - allocate distinct QoS handlers via `qosSubKey`
  - keep QUIC blocking consolidated at the parent-policy level to avoid collisions
- Moving the per-target parameter resolution into `_resolveDisturbParams()` and `_buildEffectivePolicies()` makes the behavior easier to reason about than the previous “single merged policy” shape.
- The `QoS.js` changes are directionally correct. Keying handler allocation by `{ pid, subKey }` addresses the main correctness problem where multiple targets under one parent policy would otherwise fight over the same QoS slot.
- The refactor in the app-disturb config refresh path is also good:
  - `registeredPolicies[pid]` now stores structured state
  - the update loop correctly dereferences `state.policy`

  That avoids treating the stored runtime state as if it were still the raw policy object.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** `_shortHash()` uses MD5 truncated to 8 hex chars to derive `qosSubKey`. This is probably fine most of the time, but it is still a collision-prone identifier for something that now controls QoS handler separation. A collision across targets under the same parent policy would silently merge two effective policies back onto one handler, which recreates the class of bug this PR is trying to fix.
- **Medium:** `_getPolicyTargets(policy)` does `_.uniq(_.compact(_.concat(policy.targets, policy.target)))`, but there is no normalization step before hashing or storing targets. If semantically equivalent targets can differ in representation or case, the same logical target may get:
  - duplicate effective policies
  - duplicate QUIC block entries
  - unstable `qosSubKey` values across runs

  That makes behavior depend on raw input formatting.
- **Medium:** `allocateQoSHanderForPolicy()` now stores `qos_${i} -> ref.pid`, not the full `{ pid, subKey }` identity. That may be enough for current cleanup flows, but it weakens observability/debuggability because reverse lookup no longer tells you which sub-policy owns the handler. For a feature that just introduced sub-policy allocation, that missing fidelity will make debugging much harder.
- **Low:** The change relies on ordered enforcement/unenforcement across `effectivePolicies` plus one parent-level QUIC block policy. That seems reasonable, but it does increase lifecycle complexity. If future code partially updates one target inside a registered policy, this structure will be more fragile than the old single-policy path.
- **Low:** `PolicyManager2` now passes an object to `allocateQoSHanderForPolicy()` / `getQoSHandlerForPolicy()` / `deallocateQoSHandlerForPolicy()`, but the function name typo `Hander` remains. Not a functional issue, just a minor maintainability paper cut in a newly expanded API surface.

### 💡 Suggestions
- Consider making `qosSubKey` deterministic without truncation, or at least use a lower-collision scheme if the Redis field length is not a real constraint.
- Normalize targets before dedupe/hash generation so logically identical targets cannot allocate different sub-policies.
- Store richer reverse metadata in `POLICY_QOS_HANDLER_MAP_KEY`, or add a parallel debug mapping, so a handler can be traced back to both `pid` and `subKey`.
- Add regression tests for:
  - multi-target disturb with different per-target params
  - multi-target disturb with overlapping QUIC-disabled targets
  - reenforce/unenforce symmetry
  - handler allocation uniqueness across targets under the same `pid`

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8940  
Head SHA: `26cb0b187f1915b60430b41fa6be7775fb5cd35d`  
Checked at: 2026-04-17 17:45:54 CST

jasonlyc (approved):

jasonlyc (approved):

PR #8941, opened 2026-04-16
Bug Fix

melo-crypto:
fix
[[Bug] OVPN config is mis-recoginized net_gateway config](https://github.com/firewalla/firecommit/issues/6908)
[[Bug] Box shouldn't always use first IP address in wireguard.allowedIP for VPN alarm](https://github.com/firewalla/firecommit/issues/6918)

### DIFF
--- a/extension/routing/routing.js
+++ b/extension/routing/routing.js
@@ -227,13 +227,23 @@ async function removeRouteFromTable(dest, gateway, intf, tableName, preference =
   }
 }
 
-async function flushRoutingTable(tableName, dev = null, proto="boot", af = null) {
+async function flushRoutingTable(tableName, dev = null, proto="boot", af = null, type = null) {
   const cmds = [];
-  if (af === 4 || af === null) {
-    cmds.push(`sudo ip route flush ${dev ? `dev ${dev}` : "" } proto ${proto} table ${tableName}`);
-  } 
-  if (af === 6 || af === null) {
-    cmds.push(`sudo ip -6 route flush ${dev ? `dev ${dev}` : ""} proto ${proto} table ${tableName}`);
+  if (type) {
+    // flush by route type (e.g. "throw", "unreachable"); dev and proto are irrelevant in this mode
+    if (af === 4 || af === null) {
+      cmds.push(`sudo ip route flush type ${type} table ${tableName}`);
+    }
+    if (af === 6 || af === null) {
+      cmds.push(`sudo ip -6 route flush type ${type} table ${tableName}`);
+    }
+  } else {
+    if (af === 4 || af === null) {
+      cmds.push(`sudo ip route flush ${dev ? `dev ${dev}` : "" } proto ${proto} table ${tableName}`);
+    }
+    if (af === 6 || af === null) {
+      cmds.push(`sudo ip -6 route flush ${dev ? `dev ${dev}` : ""} proto ${proto} table ${tableName}`);
+    }
   }
 
   for (const cmd of cmds) {
--- a/extension/vpnclient/OpenVPNClient.js
+++ b/extension/vpnclient/OpenVPNClient.js
@@ -100,6 +100,10 @@ class OpenVPNClient extends VPNClient {
     return `${f.getHiddenFolder()}/run/ovpn_profile/${this.profileId}.subnet${ipFamily === 6 ? '6' : ''}`;
   }
 
+  _getBypassSubnetFilePath(ipFamily=4) {
+    return `${f.getHiddenFolder()}/run/ovpn_profile/${this.profileId}.subnet${ipFamily === 6 ? '6' : ''}_bypass`;
+  }
+
   _getLocalIPFilePath(ipFamily=4) {
     return `${f.getHiddenFolder()}/run/ovpn_profile/${this.profileId}.ip${ipFamily === 6 ? '6' : '4'}`;
   }
@@ -407,6 +411,45 @@ class OpenVPNClient extends VPNClient {
     return results;
   }
 
+  async getBypassSubnets() {
+    const results = [];
+    // IPv4 bypass subnets (xx.xx.xx.xx/255.255.255.0 format)
+    const subnets = await fs.readFileAsync(this._getBypassSubnetFilePath(), "utf8")
+      .then((content) => content.trim().split("\n"))
+      .catch(() => []);
+    for (const subnet of subnets) {
+      const [network, mask] = subnet.split("/", 2);
+      if (!network || !mask)
+        continue;
+      try {
+        const ipSubnet = iptool.subnet(network, mask);
+        results.push(`${ipSubnet.networkAddress}/${ipSubnet.subnetMaskLength}`);
+      } catch (err) {
+        log.error(`Failed to parse bypass cidr subnet ${subnet} for profile ${this.profileId}`, err.message);
+      }
+    }
+    // IPv6 bypass subnets (CIDR format)
+    const subnet6s = await fs.readFileAsync(this._getBypassSubnetFilePath(6), "utf8")
+      .then((content) => content.trim().split("\n"))
+      .catch(() => []);
+    for (const subnet of subnet6s) {
+      const [network, mask] = subnet.split("/", 2);
+      if (!network || !mask)
+        continue;
+      try {
+        const addr = new Address6(network);
+        if (addr.isValid()) {
+          const prefix = parseInt(mask);
+          if (prefix >= 0 && prefix <= 128)
+            results.push(`${addr.correctForm()}/${prefix}`);
+        }
+      } catch (err) {
+        log.error(`Failed to parse bypass IPv6 subnet ${subnet} for profile ${this.profileId}`, err.message);
+      }
+    }
+    return results;
+  }
+
   async _isLinkUp() {
     const remoteIP = await this._getRemoteIP();
     if (remoteIP) {
@@ -438,7 +481,8 @@ class OpenVPNClient extends VPNClient {
     const filesToDelete = [
       this._getProfilePath(), this._getRuntimeProfilePath(), this._getUserPassPath(), this._getPasswordPath(), 
       this._getGatewayFilePath(), this._getPushOptionsPath(), this._getSubnetFilePath(), this._getIP4FilePath(),
-      this._getGatewayFilePath(6), this._getSubnetFilePath(6), this._getIP6FilePath()
+      this._getGatewayFilePath(6), this._getSubnetFilePath(6), this._getIP6FilePath(),
+      this._getBypassSubnetFilePath(), this._getBypassSubnetFilePath(6)
     ];
     for (const file of filesToDelete)
       await fs.unlinkAsync(file).catch((err) => {});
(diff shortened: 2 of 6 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The OpenVPN route-filtering change is directionally correct. Skipping pushed routes whose gateway is not the VPN gateway is the right way to avoid mistakenly treating ISP-bound `net_gateway` routes as VPN-routed subnets.
- The new `route_idx` extraction makes the IPv4 loop clearer and avoids repeating the `awk` expression inside multiple variable names.
- In `WgvpnConnSensor`, preferring `/32` for IPv4 and `/128` for IPv6 is a sensible improvement. For WireGuard peers with multiple allowed IPs, that is a better heuristic for identifying the actual peer address than blindly taking index 0.
- The event payload and log message were updated consistently in the WireGuard sensor, so the new selected peer IPs propagate end-to-end.

- No linked issue found.

### ⚠️ Issues found
- **High:** The IPv6 loop in `ovpn_remove_pushed_routes.sh` still looks wrong:
  - it iterates `for route_network_option_name in ${!route_ipv6_network_*}; do`
  - but reads `route_network="${!route_network_option_name}"`

  That variable expansion is okay, but the naming is misleading and easy to get wrong. More importantly, the script assumes the matching gateway variable is always named `route_ipv6_gateway_${idx}` and comparable to `$ifconfig_ipv6_remote`. OpenVPN IPv6 route env naming/semantics can vary, so this needs stronger validation.
- **Medium:** The new route filtering treats any route with a gateway different from `$route_vpn_gateway` / `$ifconfig_ipv6_remote` as “should go through ISP” and skips it. That may be correct for `net_gateway`, but it also excludes any pushed route that legitimately uses another next hop inside the tunnel setup. The assumption feels broader than the comment.
- **Medium:** `ovpn_remove_pushed_routes.sh` still uses unquoted variable expansions in several places:
  - `echo -n "" > $SUBNET_FILE`
  - `echo "${subnetwork}" >> $SUBNET_FILE`
  - `chown pi $SUBNET_FILE`

  In practice these paths are probably stable, but shell/network scripts should be stricter about quoting.
- **Medium:** In `WgvpnConnSensor`, picking the first `/32` or `/128` is a better heuristic, but it is still a heuristic. If a peer has multiple host routes or unusual allowed-IP layouts, the chosen address may still be arbitrary. That may affect downstream identification or UI display.
- **Low:** The WireGuard log line prints the peer public key directly on each accepted connection. That may already be existing behavior, but it is still fairly verbose/sensitive operational logging.
- **Low:** The shell comments say “config without gateway_option will not be filtered,” which is true, but that also means legacy/partial envs retain previous behavior while gateway-bearing envs change behavior. This split behavior should be called out in tests or release notes.

### 💡 Suggestions
- Add concrete test coverage for the OpenVPN script behavior with:
  - `net_gateway`
  - VPN gateway
  - missing gateway env vars
  - IPv6 pushed routes
- Tighten the shell script with consistent quoting on file paths and variables.
- Revisit whether “gateway != vpn gateway => skip” is too broad for all OpenVPN pushed-route cases, especially IPv6.
- For `WgvpnConnSensor`, consider documenting the peer-IP selection rule explicitly, since `/32` / `/128` preference is policy, not an obvious truth.
- If downstream consumers need a stable identity, consider including the full allowed-IP lists as optional metadata instead of only one chosen address.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8941  
Head SHA: `0aed72c60026a7e3cd64305a76cf3debefa7fb3c`  
Checked at: 2026-04-16 17:16:05 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The overall direction makes sense: preserving OpenVPN-pushed routes that should bypass the tunnel by materializing them into separate runtime files and then turning them into `throw` routes in the VPN routing table is a reasonable design.
- The change is nicely layered:
  - `ovpn_remove_pushed_routes.sh` classifies pushed routes into routed vs bypass
  - `OpenVPNClient` reads and normalizes them
  - `VPNClient` passes them through
  - `VPNClientEnforcer` applies them before the default route
- Cleaning up the new bypass files in `destroy()` is important and was handled.
- The `WgvpnConnSensor` tweak is a good small correctness fix: preferring `/32` and `/128` addresses over the first entry is a better heuristic for peer IP reporting.

- No linked issue found.

### ⚠️ Issues found
- **High:** The gateway classification in `ovpn_remove_pushed_routes.sh` is too broad and likely wrong for real OpenVPN push semantics:
  - IPv4 bypass if `route_gateway_X` is present and `!= $route_vpn_gateway`
  - IPv6 bypass if `route_ipv6_gateway_X` is present and `!= $ifconfig_ipv6_remote`

  This assumes “gateway differs from VPN gateway” means “should go through ISP”. That is not always true. A pushed route can legitimately use another next hop reachable through the tunnel or through topology-specific VPN routing. Converting all such routes into `throw` routes can misroute traffic instead of preserving the server’s intent.
- **High:** `VPNClientEnforcer.enforceVPNClientRoutes()` adds bypass routes as `throw` routes without adding any corresponding positive route in another table. That only works if the policy-routing flow later falls back exactly as expected. If the calling policy path depends on this table being authoritative, traffic to those subnets may become unroutable instead of “use ISP”.
- **Medium:** The OpenVPN script intentionally does **not** filter routes with no explicit gateway option:
  - `config without gateway_option will not be filtered`

  That means the same semantic destination may be treated differently depending on whether the server happened to include an explicit gateway field. This makes route handling dependent on push syntax details rather than routing intent.
- **Medium:** The IPv6 gateway comparison against `ifconfig_ipv6_remote` is heuristic and may be brittle. IPv6 route push behavior can involve link-local or topology-specific next hops, and string inequality is a weak basis for deciding tunnel-vs-bypass semantics.
- **Medium:** `getBypassSubnets()` logs parse failures but otherwise silently drops malformed entries. That is okay operationally, but if the bypass file format is wrong due to shell-script bugs, the final result will just be incomplete routing rather than a clear failure signal.
- **Low:** `parseInt(mask)` in `getBypassSubnets()` does not specify radix. In modern JS it is usually fine, but this is still better written explicitly as base 10.
- **Low:** `routing.addRouteToTable(..., "throw")` is wrapped in `.catch(() => {})`, so failures to install bypass routes are completely swallowed. That makes routing-debugging harder when this feature misbehaves.

### 💡 Suggestions
- Revisit the core rule for classifying pushed routes as bypass. “Gateway != VPN gateway” is too heuristic by itself. It would be safer to derive this from explicit OpenVPN semantics or constrain it to the exact push patterns you know Firewalla servers emit.
- Add targeted tests/examples for pushed-route combinations, especially:
  - route with no gateway option
  - route with explicit VPN gateway
  - route with explicit non-VPN gateway that is still tunnel-reachable
  - IPv6 pushed routes
- Consider logging bypass-route installation failures at least at warning/debug level instead of swallowing them completely.
- Make the IPv6 and IPv4 classification logic more explicit/documented so future maintainers understand which OpenVPN push forms are intentionally treated as ISP-bypass.
- Use `parseInt(mask, 10)` for clarity.
- If the intended behavior really is “bypass to main/ISP,” it would help to validate end-to-end that a `throw` route in this table actually produces that outcome for the policy-routing chain used by VPN clients.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8941  
Head SHA: `1b79da0d7b5743eb65c6877dd96a13bff2249cbe`  
Checked at: 2026-04-17 11:01:11 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The routing-layer change is directionally sound: adding explicit support for flushing route entries by `type` in `routing.flushRoutingTable()` is a useful primitive and makes the bypass-route lifecycle cleaner.
- The OpenVPN bypass-route flow is well threaded end-to-end:
  - `ovpn_remove_pushed_routes.sh` separates routed vs bypass subnets
  - `OpenVPNClient.getBypassSubnets()` parses the persisted files
  - `VPNClient` passes bypass subnets into enforcement
  - `VPNClientEnforcer` clears old throw routes before re-adding current ones
- Flushing existing `throw` routes before re-applying is an improvement over the earlier version; it avoids stale bypass routes accumulating across reconnects.
- Cleaning up the new bypass files in `OpenVPNClient.destroy()` is correct and important.
- The `WgvpnConnSensor` tweak is a reasonable correctness improvement: preferring `/32` and `/128` addresses gives better peer-IP reporting than blindly taking the first entry.

- No linked issue found.

### ⚠️ Issues found
- **High:** The core route classification logic in `ovpn_remove_pushed_routes.sh` is still too heuristic. Treating a pushed route as “bypass/ISP” whenever:
  - `route_gateway_X` exists and differs from `route_vpn_gateway`
  - or `route_ipv6_gateway_X` differs from `ifconfig_ipv6_remote`

  is not generally equivalent to “should bypass the tunnel.” A pushed route can legitimately use another next hop that is still part of the VPN topology. Converting those to `throw` routes can change the server’s intended routing semantics.
- **High:** `throw` routes only work if the surrounding policy-routing chain falls back exactly as expected. This PR adds:
  - `addRouteToTable(..., "throw")`

  but does not show an explicit positive route for those destinations in another table. If the lookup path does not fall back to the desired ISP/main route in all cases, traffic to those subnets can become unreachable rather than bypassed.
- **Medium:** The shell script still intentionally leaves routes without an explicit gateway in the normal VPN-routed file:
  - `config without gateway_option will not be filtered`

  That means route behavior depends on how the server encoded the push, not just on routing intent. Two semantically similar route pushes can end up treated differently.
- **Medium:** The IPv6 classification remains especially fragile. Comparing `route_ipv6_gateway_X` to `ifconfig_ipv6_remote` by string inequality is a weak proxy for “not through VPN,” and IPv6 next-hop behavior is often more topology-specific than that.
- **Medium:** `OpenVPNClient.getBypassSubnets()` still uses `parseInt(mask)` without radix and silently drops malformed entries after logging. That is survivable, but if the shell-side classification is wrong or file contents drift, the runtime behavior becomes incomplete rather than explicitly failed.
- **Low:** The bypass-route installation still swallows route-add failures:
  - `.catch((err) => {})`

  That makes diagnosing real routing problems harder.
- **Low:** `flushRoutingTable(tableName, null, null, null, "throw")` ignores `proto`/`dev` intentionally, which is fine, but it broadens the flush scope for all throw routes in that table. If other throw routes are ever added to the same VPN table for different reasons, they will also be wiped.

### 💡 Suggestions
- Revisit the bypass classification rule in `ovpn_remove_pushed_routes.sh`. “gateway differs from VPN gateway” is not a strong enough condition by itself to mean “send via ISP.”
- Add tests or documented examples for the exact OpenVPN push forms this feature is meant to support, especially:
  - explicit VPN gateway
  - explicit non-VPN gateway
  - missing gateway option
  - IPv6 routes
- Consider logging throw-route add failures at least at warning/debug level instead of swallowing them entirely.
- Use `parseInt(mask, 10)` for clarity.
- If the product assumption truly is “throw route in VPN table means fall back to ISP/main,” it would help to add an integration/regression test for that policy-routing outcome, not just route installation.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8941  
Head SHA: `0fc84133fee878d9599fd3810c4c9cef4acd956c`  
Checked at: 2026-04-17 12:46:38 CST

jasonlyc (approved):

PR #8943, opened 2026-04-16
fix empty regional ip6 set

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The refactor is moving in a sensible direction overall:
  - temp ipsets are no longer kept around permanently
  - temp set naming is being consolidated behind the main helpers
  - `updateIPSetByDomain*()` call sites are cleaner now that IPv4/IPv6/temp selection is handled by one API
- Creating temp ipsets only during recycle should reduce long-lived kernel memory usage compared with always pre-creating all temporary sets.
- The country-updater changes are directionally reasonable if the goal is to treat IPv4 as bulk static data and only maintain IPv6 dynamically from observed traffic.
- The added comment in `net2/Iptables.js` is useful. That comparison path is subtle, and calling out the exact `iptables-save` parity requirement helps prevent future duplicate-rule regressions.
- Passing the already-known address family into `IPUtil.isPrivate()` / `isPublic()` is a nice small cleanup and avoids redundant `net.isIP()` calls in hot paths.

- No linked issue found.

### ⚠️ Issues found
- **High:** `createTempIpsets()` has a precedence bug in the `dstType` assignment:
  - `const dstType = liveMeta && liveMeta.type || this.constructor.name === 'CountryUpdater' ? 'hash:net' : 'hash:ip';`

  In JavaScript this is parsed as:
  - `((liveMeta && liveMeta.type) || this.constructor.name === 'CountryUpdater') ? 'hash:net' : 'hash:ip'`

  So if `liveMeta.type` is any truthy string, `dstType` becomes `'hash:net'` regardless of the real live-set type. That means normal category temp sets can be recreated with the wrong type, which can break later add/swap behavior.
- **High:** `CountryUpdater.updateIP()` changed delete behavior from removing the original `ip` to removing all derived `CIDRs`:
  - old: `zremAsync(key, ip)` / `Ipset.del(ipset, ip)`
  - new: `zremAsync(key, CIDRs)` / `CIDRs.forEach(entry => Ipset.del(ipset, entry))`

  But the add path stores `CIDRs` only after geo lookup at the current point in time. If a later geo lookup returns a different CIDR normalization set for the same IPv6 address, the previous entries may not be removed correctly. This makes remove semantics dependent on recomputing the exact same derived CIDR list.
- **Medium:** `createTempIpsets()` now creates fresh temp sets instead of flushing existing ones, but it only carries over:
  - `type`
  - `comment`
  - `maxelem`

  It drops other live-set creation parameters such as `hashsize`. If swap correctness depends on matching set characteristics closely, this can change behavior or performance for large categories.
- **Medium:** `CountryUpdater.addDynamicEntries()` switched from `Ipset.testAndAdd(entries, ipsetName, 60)` to `entries.forEach(entry => Ipset.add(ipsetName, entry))`. That removes dedup/test behavior and batching semantics. If the dynamic key contains repeated entries, this could generate more ipset operations and noisier failures.
- **Medium:** `setupCategoryEnv()` no longer pre-creates temp sets, and `createTempIpsets()` is only called from recycle paths. Any other path that still assumes temp sets already exist when `useTemp: true` is passed will now fail unless it is always preceded by recycle setup.
- **Low:** `CountryUpdater.updateIP()` now silently returns for any non-IPv6 input instead of throwing on invalid input. That may be intentional, but it also hides bad callers and malformed data that previously surfaced immediately.
- **Low:** `hook/DestIPFoundHook.js` changes the logger from explicit `'info'` to default level. That may be fine, but it is an incidental behavior change in a PR that is otherwise about ipset/category logic.

### 💡 Suggestions
- Fix the `dstType` expression immediately by adding explicit parentheses, or better yet split it into separate statements for readability.
- Add tests around `createTempIpsets()` that verify temp sets inherit the correct type for:
  - normal category sets
  - country sets
  - IPv4 and IPv6 variants
- Re-check whether temp-set creation should preserve more metadata from the live set, especially `hashsize`.
- Add regression tests for `CountryUpdater.updateIP()` remove behavior to ensure previously added IPv6 CIDRs are actually removable across repeated geo lookups.
- Audit all `useTemp: true` call paths to confirm they only happen after temp-set creation in the new lifecycle.
- Consider keeping some validation/logging when `updateIP()` receives invalid or unexpected non-IPv6 input, even if the new behavior is to ignore it.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8943  
Head SHA: `074a78d918104611b87474e016c6423881b6be30`  
Checked at: 2026-04-16 20:31:13 CST

jasonlyc (approved):

PR #8944, opened 2026-04-16
mesh_script

Chetarokh:
1-hop mesh performance

### DIFF
--- a/test/mesh_test.py
+++ b/test/mesh_test.py
@@ -1,0 +1,109 @@
+import subprocess
+import time
+import re
+
+# Configure your mesh nodes
+mesh_nodes = [
+    {"ip": "192.168.201.84", "role": "ROOT"},
+    {"ip": "192.168.10.198", "role": "SATELLITE"}
+]
+
+IPERF_DURATION = 10  # seconds
+
+
+def run_command(cmd):
+    try:
+        result = subprocess.run(
+            cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+            text=True
+        )
+        return result.stdout
+    except Exception as e:
+        return f"Error: {e}"
+
+
+# -----------------------------
+# Ping Test
+# -----------------------------
+def ping_test(host):
+    output = run_command(["ping", "-c", "4", host])
+
+    packet_loss = "N/A"
+    latency = "N/A"
+
+    for line in output.split("\n"):
+        if "packet loss" in line:
+            packet_loss = line.split(",")[2].strip()
+        if "round-trip" in line or "rtt" in line:
+            latency = line.split("=")[1].strip()
+
+    return packet_loss, latency
+
+
+# -----------------------------
+# iPerf Test
+# -----------------------------
+def iperf_test(server_ip):
+    print(f"🚀 Running iperf3 test to {server_ip}...")
+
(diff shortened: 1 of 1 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The script is small and easy to read.
- Separating the workflow into `ping_test()`, `iperf_test()`, and `extract_receiver_bitrate()` is a sensible structure for a quick diagnostic utility.
- Using `subprocess.run([...])` with argument lists instead of shell strings avoids shell-injection risk.
- Capturing both ping loss/latency and iperf throughput in one loop is useful for a basic mesh-health smoke test.

- No linked issue found.

### ⚠️ Issues found
- **High:** The test is effectively unbounded. `for i in range(99999):` means this will run for days and continuously hammer both mesh nodes with repeated ping + `iperf3` traffic. That is risky for a test script in `test/`, especially if someone expects it to finish as part of a normal validation flow.
- **High:** `run_command()` ignores command failures because it never checks `returncode`. If `ping` or `iperf3` fails, exits non-zero, or writes the useful error to `stderr`, the caller still treats the run as successful and parses `stdout`. That can easily produce misleading output like `"Not found"` instead of a real failure reason.
- **Medium:** `extract_receiver_bitrate()` relies on one very specific iperf output shape:
  - a bracketed stream id
  - a receiver summary line
  - units limited to `[GMK]bits/sec`

  This is brittle across iperf versions, localization, parallel streams, JSON mode, or outputs that use lowercase/other spacing conventions.
- **Medium:** The script hardcodes live lab IPs directly in the test file. That makes the test environment-specific and non-portable, which is unusual for something placed under `test/`.
- **Medium:** The script prints results but never asserts anything, returns pass/fail status, or exits non-zero on bad conditions. As written, it is more of a manual soak tool than an actual test.
- **Medium:** `ping_test()` parses `packet loss` with `line.split(",")[2]`, which assumes the exact macOS/Linux summary format. That is fragile and can break if ping output varies.
- **Low:** There is no timeout on subprocess execution. If `iperf3` hangs or a command blocks unexpectedly, the whole script can stall indefinitely.
- **Low:** Variable naming/style is inconsistent (`TPUT` uppercase while most variables are lowercase), which makes the script feel ad hoc.

### 💡 Suggestions
- Replace the `99999` loop with a bounded iteration count or a duration-based flag so the script finishes predictably.
- Make command execution fail loudly:
  - check `returncode`
  - include `stderr` in error reporting
  - consider `timeout=` in `subprocess.run(...)`
- If this is intended to be a real automated test, add explicit pass/fail thresholds for:
  - max packet loss
  - max latency
  - min throughput
- Move node IPs to CLI arguments, env vars, or a config file instead of hardcoding them in the test source.
- Consider using `iperf3 --json` and parsing JSON instead of regexing human-readable output.
- If this is really a manual soak/diagnostic tool, place it outside `test/` or rename/document it so people do not mistake it for a normal unit/integration test.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8944  
Head SHA: `6b9394f36ef453ae28e7c03ca026d0f26f537025`  
Checked at: 2026-04-17 01:30:58 CST

MelvinTo (approved):

PR #8945, opened 2026-04-16
Add reboot AP via serial script

Chetarokh:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The script is straightforward and easy to follow for a manual serial-console workflow.
- Logging the captured serial output to a file is useful for postmortem analysis when a reboot cycle fails.
- The sequence of collecting a few basic facts before rebooting:
  - `fwap --version`
  - firmware version
  - `ethtool eth0 | grep Speed`

  is sensible for a quick bring-up/reboot smoke loop.
- Using `errors="ignore"` on decode avoids crashing on noisy serial bytes.

- No linked issue found.

### ⚠️ Issues found
- **High:** This is an infinite reboot loop with no stop condition:
  - `while True:`

  That is dangerous in a `test/` script because it can repeatedly reboot hardware indefinitely if started accidentally, and it provides no built-in max iteration count, timeout budget, or safety confirmation.
- **High:** The script never verifies login state or command success before issuing `reboot`. It opens the serial port and immediately sends shell commands, but:
  - there is no prompt detection
  - no login handling
  - `username` / `password` are defined but unused

  On many boot states, those commands will either be ignored or land at the wrong prompt, and the script may report misleading output.
- **High:** Serial-port open errors are unhandled. If `/dev/ttyUSB0` is missing, busy, renamed, or permission-denied, `serial.Serial(...)` will raise and terminate the whole loop abruptly.
- **Medium:** The script relies entirely on fixed `sleep()` delays (`2s`, `5s`, `150s`) instead of reading until a prompt or known readiness signal. That makes it fragile across slower/faster boots and different AP states.
- **Medium:** `ser.read(ser.in_waiting)` can return empty output if bytes have not arrived yet. The script treats that as normal and continues, so failures can look like blank command output rather than explicit errors.
- **Medium:** The log file path is hardcoded to `serial_output.log` in the current working directory. Repeated infinite runs can grow that file indefinitely.
- **Low:** `print("\r Port speed of AP is:", output.decode(errors="ignore").replace(" ", " "))` performs a no-op replace and looks accidental.
- **Low:** `sys` is imported but unused.
- **Low:** `username` and `password` are dead variables, which makes the script look unfinished and creates confusion about whether authentication is required.

### 💡 Suggestions
- Add a hard safety bound:
  - `--iterations N`
  - or `--duration`
  - or at least a configurable max loop count
- Handle serial setup failures explicitly with clear errors and retry/backoff logic.
- Replace fixed sleeps with prompt/readiness detection where possible:
  - wait for login prompt
  - send credentials if needed
  - wait for shell prompt before running commands
- Validate each command result and fail loudly if expected output is missing before issuing `reboot`.
- Make the serial device, baud rate, and wait times CLI arguments instead of hardcoding `/dev/ttyUSB0`.
- Add log rotation or per-run log filenames to avoid unbounded log growth.
- Remove unused imports/variables and clean up the no-op string replacement.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8945  
Head SHA: `459d6a87d49f53344ffaf9976cdb98fe2a46339d`  
Checked at: 2026-04-17 05:00:59 CST

MelvinTo (approved):

PR #8946, opened 2026-04-17
fix bug: add debug log to DomainUpdater

caixingcheng:
add debug log to print job queue length

### DIFF
--- a/control/DomainUpdater.js
+++ b/control/DomainUpdater.js
@@ -65,6 +65,7 @@ class DomainUpdater {
       });
 
       setInterval(async () => {
+        log.debug(`_priorityJobQueue.length:${this._priorityJobQueue.length}, _standardJobQueue.length:${this._standardJobQueue.length}`);
         if (this._needRefresh) {
           log.debug("DomainUpdater refreshing connmark for updated ipsets");
           await this.scheduleRefreshConnmark();


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The change is minimal and low-risk.
- Adding queue-length visibility around the refresh loop can help diagnose whether `DomainUpdater` is falling behind on priority vs standard work.
- Because this is a `debug` log, it is at least scoped below normal production info/warn noise.

- No linked issue found.

### ⚠️ Issues found
- **Low:** This runs inside a recurring `setInterval`, so even a debug log here can become very noisy if the interval is short or debug logging is enabled for an extended period. That can make troubleshooting harder rather than easier, especially if the queue is usually idle.
- **Low:** The log is emitted unconditionally on every tick, even when both queues are empty and `_needRefresh` is false. In practice, the most useful signal is probably when backlog exists or changes.

### 💡 Suggestions
- Consider logging only when one of these is true:
  - `_priorityJobQueue.length > 0`
  - `_standardJobQueue.length > 0`
  - `_needRefresh === true`
- If this is intended as temporary observability, it may be better to remove it after the underlying issue is understood.
- If queue health matters long term, consider a rate-limited or state-change-based debug log instead of per-tick logging.

### Verdict
APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8946  
Head SHA: `db681357d1e70870003a5fd3f730a2b9a88cd5a7`  
Checked at: 2026-04-17 18:01:24 CST

jasonlyc (approved):

PR #8949, opened 2026-04-20
more fixes on ipset/iptables control

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a good correctness fix for a precedence bug in a sensitive code path.
- Adding parentheses makes the intended fallback explicit:
  - use `liveMeta.type` when available
  - otherwise choose `hash:net` only for `CountryUpdater`
  - otherwise fall back to `hash:ip`
- The change is small, targeted, and easy to reason about.
- This is exactly the kind of bug that can silently create the wrong temp ipset type, so fixing it at the expression level is the right move.

- No linked issue found.

### ⚠️ Issues found
- I do not see a blocking issue in this diff.
- The pre-fix expression was definitely ambiguous/wrong because `?:` has lower precedence than `||`, so without parentheses the condition did not evaluate the way the surrounding comment suggests.

### 💡 Suggestions
- Consider avoiding mixed `&&` / `||` / ternary expressions here entirely for readability, e.g. split into:
  - `const liveType = liveMeta && liveMeta.type;`
  - `const defaultType = ...;`
  - `const dstType = liveType || defaultType;`
- A small unit test around `createTempIpsets()` fallback behavior would be worthwhile, especially for:
  - live set present with explicit type
  - `CountryUpdater` without live metadata
  - non-country updater without live metadata

### Verdict
APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8949  
Head SHA: `bfd0b86f6e55507279eb8f9b20b9c005e8ef63bc`  
Checked at: 2026-04-20 10:46:14 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This looks like a solid cleanup/refactor pass with a couple of correctness fixes mixed in.
- The most important correctness fix is in `CategoryUpdaterBase.createTempIpsets()`:
  - adding parentheses around the ternary removes the operator-precedence bug
  - without that, `CountryUpdater` could incorrectly force `hash:net` regardless of `liveMeta.type`
- The `Ipset.read()` / `Ipset.isReferenced()` changes are improvements:
  - using structured `read(..., true)` metadata is more robust than shelling out to `ipset list | grep | cut`
  - empty-response logging was downgraded from warn to verbose, which should reduce noise for normal “set missing / empty” cases
- The move from raw `exec(wrapIptables(...))` toward `IptablesControl` + `Rule` is directionally good:
  - more consistent rule construction
  - less stringly-typed shell command assembly
  - easier future batching / centralized error handling
- `IpsetControl` now avoids repeatedly warning on missing `del` operations and logs duplicate `create`s explicitly, which should help operational clarity without flooding logs.

- No linked issue found.

### ⚠️ Issues found
- **Low:** `extension/diag/app.js` now applies the same NAT redirect twice on create/remove paths:
  - once through `iptc.addRule(new Rule(...))`
  - once through `exec(wrapIptables(...))`

  The comment says this is intentional “in case FireMain is not up”, but in the normal case both paths will run. That can create duplicate-insert / duplicate-delete behavior and makes success semantics dependent on how `IptablesControl` deduplicates or tolerates existing rules. It is probably harmless, but it is not cleanly idempotent.
- **Low:** several modules (`clash`, `clash_tun`, `DockerBaseVPNClient`, `ModeManager`, `STUNSensor`) were switched from awaited shell execution to fire-and-forget `iptc.addRule(...)`. If `iptc.addRule()` is queued/asynchronous internally, these call sites may no longer provide the same sequencing guarantees as the old `await exec(...)` behavior. I don’t see an obvious concrete regression in this diff, but it is worth confirming for startup/teardown-sensitive paths.

### 💡 Suggestions
- For `extension/diag/app.js`, consider choosing one path explicitly:
  - use `IptablesControl` when available,
  - fall back to raw `exec(...)` only when the controller/runtime is unavailable,
  rather than always doing both.
- For the `IptablesControl` migrations, add or confirm tests/validation around:
  - redirect enable/disable ordering
  - VPN client start/stop NAT rules
  - STUN temporary accept rule add/remove lifecycle
- If `iptc.addRule()` is intentionally non-blocking, it may be worth documenting that in these call sites where previous code relied on `await exec(...)`.

### Verdict
APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8949  
Head SHA: `32527742f5ed0adb989ea25209e5fb8036f36fce`  
Checked at: 2026-04-20 14:48:19 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The `CategoryUpdaterBase` fix is the most important part here:
  - switching `activeCategories[category]` from a boolean-ish marker to the actual ipset type fixes the temp-ipset creation path much more robustly than relying on the previous precedence-buggy expression
  - `CountryUpdater` now explicitly sets `'hash:net'`, which matches the intended country ipset semantics
- The added logging in `createTempIpsets()` / `swapIpset()` should help debug category lifecycle issues.
- `Ipset.read()` / `Ipset.isReferenced()` are cleaner and less shell-fragile now that they use structured metadata instead of `ipset list | grep | cut`.
- Downgrading repetitive device-presence-disabled logs from `info` to `verbose` is a good noise reduction.
- The `IptablesControl` / `Rule` migrations continue the cleanup away from string-built shell commands.

- No linked issue found.

### ⚠️ Issues found
- **Low:** `extension/diag/app.js` still programs the same redirect rule through two paths:
  - `iptc.addRule(new Rule(...))`
  - `await exec(wrapIptables(...))`

  The comment explains why, but in the normal case this is still a duplicate-application model rather than a true fallback path.
- **Low:** several call sites that previously `await`ed `exec(wrapIptables(...))` now just call `iptc.addRule(...)`. If `IptablesControl` is queued/non-blocking, that may subtly change sequencing guarantees during startup/teardown-sensitive paths such as STUN temp rule setup, compatible spoof toggling, Clash redirect enable/disable, or Docker VPN lifecycle.
- **Low:** `createTempIpsets()` now returns early if `activeCategories[category]` is missing. That is probably safer than creating the wrong type, but it also means callers now silently skip temp ipset creation on missing state after only logging an error. If this state can drift, downstream behavior may become partial instead of obviously failing fast.

### 💡 Suggestions
- For `extension/diag/app.js`, consider making the exec path a real fallback only when `IptablesControl` is unavailable, instead of always doing both.
- Add/confirm tests around the category ipset lifecycle:
  - `activateCategory()` stores the correct type
  - `createTempIpsets()` uses that type for temp sets
  - country category path stays `hash:net`
- For the `IptablesControl` migrations, verify ordering-sensitive flows explicitly:
  - STUN temporary accept add/remove
  - Docker VPN NAT add/remove
  - Clash redirect enable/disable
  - compatible spoof toggle on/off
- If `activeCategories[category]` missing is a hard invariant violation, consider surfacing it more strongly than a log-and-return.

### Verdict
APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8949  
Head SHA: `8d5e59b9d3314cfe9ec3d10a79ca7f6f3e8f16f8`  
Checked at: 2026-04-20 20:01:58 CST

jasonlyc (approved):

PR #8950, opened 2026-04-20
Bug Fix and Enhancement

melo-crypto:
send batch of operations to MSP instead of one by one
https://github.com/firewalla/firecommit/issues/7586
Support both plain text list and bloomfilter for one target list from cloud/msp


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Batching ops before compress/encrypt/send is a sensible optimization.
- The queue ordering is handled carefully:
  - `dequeueOps()` pops from the head in order
  - `pushOps()` reverses before `LPUSH`, which preserves original order when requeueing on failure
- The new `this.socket && this.isSocketConnected()` guard is safer than checking only `this.socket`.
- Reducing per-op compression/encryption overhead into one payload should materially improve throughput and socket efficiency.

- No linked issue found.

### ⚠️ Issues found
- **High:** this changes delivery semantics from many small emits to one large emit with up to `100` ops, but there is no payload size guard. If queued ops are large, the combined JSON → deflate → encrypt blob can become too large for the socket/server path, causing repeated batch failure and requeue loops. The previous implementation had an implicit size cap because it sent one op at a time.
- **Medium:** the old code paced sends with `await delay(200)`. This patch removes pacing entirely and emits one large batch immediately. That may be fine, but it also changes backpressure characteristics significantly. If the MSP side expects a steadier flow or has message-size-sensitive handling, this could make failures burstier.
- **Medium:** on any failure, the entire batch is pushed back, even if the failure happens after building a valid encrypted payload but before/during emit handling. Since `socket.emit(...)` itself is not acked here, the code still has at-least-once / possible-duplicate semantics, and batching amplifies the duplicate surface from one op to up to 100 ops at once.
- **Low:** `dequeueOps()` performs repeated `LPOP` calls in a loop. I understand the Redis 6.0.16 comment, but this is still multiple round trips and partially weakens the batching benefit on the dequeue side.

### 💡 Suggestions
- Add a size-based split, not just a count-based split. `MAX_OPS_PER_TIME = 100` is too coarse by itself.
- If possible, add an application-level ack or batch ID so retries are less duplication-prone.
- Consider making batch size configurable via MSP/cloud config so it can be tuned safely in production.
- If Redis version is a constraint, this is acceptable for now, but document that `dequeueOps()` is a compatibility workaround rather than ideal batching.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8950  
Head SHA: `b66b8369e42601cead5234373e1fc1fbf7f3d336`  
Checked at: 2026-04-20 12:17:25 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Batching here is directionally sensible:
  - fewer Redis round-trips
  - fewer compress/encrypt cycles
  - fewer socket emits
- `pushOps()` correctly reverses before `LPUSH`, so queue order is preserved when requeuing a failed batch.
- `dequeueOps()` is a reasonable compatibility fallback given the Redis 6.0.16 note about `LPOP key count`.
- The socket guard is slightly safer now with `this.socket && this.isSocketConnected()` instead of only checking `this.socket`.

- No linked issue found.

### ⚠️ Issues found
- **High:** this changes the wire contract from sending one op at a time to sending `data: ops` as an array in a single `replyid: "sync_op_from_box"` message. If MSP currently expects `data` to be a single op object, this will silently break processing even though the local code looks fine. I don’t see any paired receiver-side change in this diff.
- **Medium:** batching up to 100 ops into one compressed/encrypted payload increases message size substantially. The old code naturally paced itself at 20 ops max with 200 ms spacing. This version can emit one much larger payload immediately, which raises risk around:
  - socket/frame size limits
  - downstream parser/memory pressure
  - retries becoming more expensive because one failure replays the whole batch
- **Medium:** `dequeueOps()` removes all items before the batch is sent. The previous implementation had the same general at-least-once pattern, but the failure window is now much larger: up to 100 ops can be in-flight and lost on process crash between dequeue and successful emit/requeue.
- **Low:** removing the per-op pacing changes runtime behavior significantly. If the MSP side depends on gradual ingestion or ordering across multiple emits, this can create burstiness that did not exist before.

### 💡 Suggestions
- Verify the MSP receiver contract first. If it expects a single op, keep the external message shape unchanged and only batch internal compression/encryption/emit if the receiver is updated in lockstep.
- Consider a smaller batch size or a payload-size cap, not just a count cap.
- If crash-safety matters, consider a reserve/ack pattern instead of destructive dequeue-before-send.
  - receiver handling of batched `data`
  - queue order preservation after failed requeue
  - oversized/boundary batches
  - reconnect/failure behavior with partially drained queues

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8950  
Head SHA: `d1b34b7f8fd8a969e67dac800890ec76d3e30124`  
Checked at: 2026-04-21 12:31:07 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- There is some good intent in this change:
  - batching MSP sync ops should reduce per-message compression/encryption overhead
  - `pushOps()` preserves original queue order correctly by reversing before `LPUSH`
  - checking `this.isSocketConnected()` before syncing is safer than only checking `this.socket`
- The added category IP loading in `CategoryUpdateSensor` is directionally useful if the goal is to enrich category data with Bone-provided IP entries.

- No linked issue found.

### ⚠️ Issues found
- **High:** `sensor/Guardian.js` changes the wire shape of `replyid: "sync_op_from_box"` from a single op object to `data: ops` where `ops` is an array. That is a protocol/contract change, not just an internal batching optimization. Unless the MSP receiver is explicitly updated to accept an array for this reply type, this can break sync handling outright.
- **High:** because the entire batch is emitted as one message, any receiver/parser failure now blocks up to 100 ops at once instead of just one. That increases blast radius and makes recovery/debugging harder.
- **Medium:** `dequeueOps()` repeatedly `LPOP`s and `JSON.parse`s entries without per-item error handling. A single malformed queue item will throw and abort the whole batch path, potentially leaving the queue partially drained and the remaining ops unsent for that cycle.
- **Medium:** the new category IP import path in `CategoryUpdateSensor` appends IPv4/IPv6 addresses for every matching category update, but I don’t see any dedupe/replace semantics here. If `addIPv4Addresses` / `addIPv6Addresses` are additive, repeated refreshes could accumulate stale or duplicate data unless those helpers already handle replacement internally.

### 💡 Suggestions
- Keep the batching optimization internal unless/until the MSP side contract is explicitly versioned or updated. Safer options:
  - emit multiple single-op messages after one dequeue batch
  - or introduce a new reply type for batched payloads instead of changing `sync_op_from_box`
- Add defensive parsing in `dequeueOps()` so a malformed queue entry does not take down the entire batch.
- If batching is required, add an explicit receiver compatibility test for `sync_op_from_box` with array payloads.
- For the category IP sync path, verify and document whether `addIPv4Addresses` / `addIPv6Addresses` replace, merge, or dedupe existing entries. If not, consider clearing/reconciling old values before re-adding.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8950  
Head SHA: `242f8b693a6d4a9bcfdb7600efca42a57c7bed15`  
Checked at: 2026-04-22 10:31:43 CST

jasonlyc (approved):

PR #8951, opened 2026-04-20
Reimplement extra time

caixingcheng:
1. introduce bypass rule to handle extra time request
2. Fixed bugs found in the previous version.


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The PR addresses a real gap in the access-request flow: instead of disabling the original block/disturb policies, it introduces a dedicated bypass layer that can expire independently and preserve the underlying rule set.
- Several pieces are directionally solid:
  - `getAppsFromPolicy()` centralizes app extraction for multi-target policies.
  - `findMatchingTimeLimitRules()` now separates `bypassPolicies` from `blockOrDisturbPolicies`, which is a better model than overloading one list.
  - `appTimeUsage.onQuotaReached === "unenforce"` is a clean extension point in `AppTimeUsageManager`.
  - `time_limits.js` now includes multi-target rules in the `"all"` listing path instead of dropping them outright.
- The new bypass-path plumbing is reasonably comprehensive:
  - Redis tracking
  - policy serialization/comparison support for `affectedPids`
  - dnsmasq bypass entries
  - policy activation/deactivation synchronization via `AccessRequestSensor`

- No linked issue found.

### ⚠️ Issues found
- **High:** `matchTarget()` still explicitly rejects multi-target rules during general matching:
  - `if (policy.targets && policy.targets.length > 1) return false;`

  That means normal app-based lookups still cannot match a multi-target time-limit rule unless the caller uses the synthetic `policy:<pid>` path. This is still the core correctness gap from the earlier access-request review: multi-target rules are not actually participating in normal matching semantics.
- **High:** the bypass rule keying and lookup still depend on exact app-array identity/order. `approveRequest()` builds the bypass key from `apps` derived from the selected policy / request path, and `time_limits.js` later reconstructs the key from `appQuotaInfo.apps.sort()`. This is much better than the earlier `[appQuotaInfo.app]` bug, but it is still fragile when multiple overlapping rules produce different app sets for the same displayed entry. A bypass tied to one affected policy-set can be missed or misapplied when the “best match app” chosen for display is not the same app-set that created the bypass.
- **High:** bypass cleanup lifecycle is incomplete. The PR adds:
  - `affectedPids`
  - `active_bypass_policy`
  - chain creation / flush / delete logic
  - sensor-based affectedPid updates

  but there is no obvious cleanup of the Redis `bypass:extraTime:${userId}:${apps.join(',')}` pointer when:
  - the bypass policy is deleted early,
  - the bypass policy save/update path changes pid,
  - or the policy expires and is removed outside the exact `approveRequest()` path.

  That leaves stale pointers likely, which can cause future approvals/lookups to bind to nonexistent bypass policies.
- **Medium:** `FWRuleDiagViewController`-style inversion bug equivalent does not exist here, but there is a notable logic risk in `AppTimeUsageManager`:
  - for `onQuotaReached === "unenforce"`, `refreshPolicy()` now enforces the bypass policy whenever usage is *below* quota and unenforces it once quota is reached.

  That is the intended high-level behavior, but it assumes `applyPolicy()`/`unapplyPolicy()` are fully symmetric and idempotent for this new action. Given the new bypass implementation spans iptables + dnsmasq + per-policy chain management, this path really needs regression coverage; otherwise stale enforcement state is easy to leave behind.
- **Medium:** `AccessRequestSensor.updateRelatedBypassPolicies()` iterates all tags/users and performs policy fetch/update/re-enforcement on every relevant activation/deactivation. That is functional, but it can become noisy and expensive in policy-heavy systems, and it broadens the set of code paths that can mutate a bypass policy after approval. Without strong tests around concurrent rule updates, this looks brittle.
- **Medium:** `control/Bypass.js` only implements `setupTagsRules()`. `PolicyManager2._applyBypass()` parses both `intfs` and `tags`, but ultimately routes everything through `Bypass.setupTagsRules({ pid, affectedPids, intfs, tags, action, targets })`. The name is misleading, and more importantly the implementation only handles tag/interface contexts in one specific way. If a future bypass policy targets other scope types, this will silently diverge from normal policy enforcement semantics.
- **Low:** `getRequest()` mutates deserialized payloads by deleting fields when `req.note == 'null'` / `req.mac == 'null'`. That cleanup is harmless, but it hides upstream serialization issues instead of fixing them at write time.

### 💡 Suggestions
- Remove the multi-target short-circuit in `matchTarget()` and make general matching honor `policy.targets` properly. Otherwise multi-target rules remain second-class and only work through the special `policy:<pid>` flow.
- Add explicit lifecycle cleanup for the Redis bypass key:
  - on policy delete
  - on expiry auto-delete
  - on broken/stale referenced pid
- Add regression tests for the full bypass lifecycle:
  - approve request for single-app block rule
  - approve request for multi-target block rule
  - approve request for disturb rule
  - usage below quota → bypass enforced
  - usage reaches quota → bypass unenforced
  - expiry/delete cleanup for iptables + dnsmasq + Redis pointer
  - affected policy activation/deactivation updates
- Consider storing a canonical bypass-policy mapping on the policy itself (or a dedicated reverse index) rather than relying only on `userId + sorted apps` string keys.
- Rename or generalize `Bypass.setupTagsRules()` if it is intended to be the central bypass enforcement entrypoint.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8951  
Head SHA: `554e8048c0d3724d4176945cd816d62443b2537b`  
Checked at: 2026-04-20 14:47:44 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision moves the access-request flow in the right direction:
  - multi-app matching is now modeled explicitly with comma-separated app sets
  - `getAppsFromPolicy()` is reused consistently
  - `findMatchingTimeLimitRules()` now returns separate `bypassPolicies` and `blockOrDisturbPolicies`
  - `time_limits.js` now includes multi-target rules in the general listing path instead of forcing a `policy:<pid>` special case
- The bypass-policy plumbing is broader and more complete than the earlier version:
  - `affectedPids` is persisted on `Policy`
  - bypass policies are tracked in `PolicyManager2`
  - iptables/dnsmasq bypass support is wired in
  - `AccessRequestSensor` keeps `affectedPids` synchronized when related policies activate/deactivate
- The `onQuotaReached: "unenforce"` path in `AppTimeUsageManager` is a reasonable extension for temporary bypass behavior.

- No linked issue found.

### ⚠️ Issues found
- **High:** `matchTarget()` is still incorrect for single-app lookups against multi-target rules. In the new code:
  - if `app === "youtube"` and a policy has `targets: ["TLX-fw-youtube", "TLX-fw-netflix"]`
  - `apps.length` is `1`, so execution goes into the `else` branch
  - that branch only checks `policy.target`, not `policy.targets`

  Since multi-target rules usually do not have a singular `policy.target`, they now fall through and return `true`. That means a single-app lookup can incorrectly match unrelated multi-target rules.
- **High:** the Redis pointer created via `getBypassPolicyKey()` is still not cleaned up when the referenced bypass policy is deleted or expires. The PR removes bypass policies from `policy_active` / `active_bypass_policy`, but the separate key:
  - `bypass:extraTime:${userId}:${apps.join(',')}`

  can still become stale and point to a nonexistent pid. Future lookups/approvals will then bind through dead state.
- **Medium:** bypass matching is still keyed purely by the sorted app list string. That is better than the earlier single-app bug, but it is still fragile when multiple overlapping rules share app sets differently or when policy scope/tag relationships evolve independently from the app list. The mapping is convenient, but not a strong identity.
- **Medium:** the bypass lifecycle remains fairly complex and cross-cutting:
  - `AccessRequestManager`
  - `AppTimeUsageManager`
  - `PolicyManager2`
  - `Block`
  - `Bypass`
  - `dnsmasq`
  - `AccessRequestSensor`

  I don’t see regression coverage in this diff for create/update/expire/delete/reenforce behavior. Given the amount of state mutation here, that is still risky.

### 💡 Suggestions
- Fix `matchTarget()` so single-app lookups still verify membership in `policy.targets` when `policy.target` is absent. Right now multi-target rules overmatch.
- Add cleanup for the `bypass:extraTime:...` Redis pointer when:
  - the bypass policy expires
  - the bypass policy is deleted
  - the stored pid no longer resolves
- Add regression tests for:
  - single-app lookup does **not** match unrelated multi-target rules
  - multi-app lookup matches the exact app-set rule
  - bypass creation/update for multi-app rules
  - quota reached → bypass unenforced
  - expire/delete cleanup for iptables + dnsmasq + Redis pointer
  - `affectedPids` updates when related policies activate/deactivate
- Consider a stronger reverse mapping than only `userId + sorted apps`, or at least validate the stored bypass pid aggressively before trusting it.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8951  
Head SHA: `f1f6d9d97cb4bf0e39c8d75e64ca691f80c3cd90`  
Checked at: 2026-04-21 10:31:27 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is moving in the right direction compared with the earlier version:
  - multi-app rules are modeled explicitly instead of the old `policy:<pid>` escape hatch
  - bypass policies are now first-class in `Policy`, `PolicyManager2`, `dnsmasq`, and the access-request flow
  - `AccessRequestSensor` attempts to keep `affectedPids` synchronized as related policies activate/deactivate
- The separation in `findMatchingTimeLimitRules()` between:
  - `bypassPolicies`
  - `blockOrDisturbPolicies`

  is a cleaner interface than returning one mixed list.
- The `onQuotaReached: "unenforce"` addition in `AppTimeUsageManager` is a reasonable mechanism for temporary bypass behavior.

- No linked issue found.

### ⚠️ Issues found
- **High:** `matchTarget()` is still incorrect for single-app lookups against multi-target policies. If `app` is a single value, the new logic only validates `policy.target` and does not check membership in `policy.targets`. That means a rule like `["youtube","netflix"]` can still incorrectly match a lookup for just `"youtube"` simply because the function falls through to `true`.
- **High:** the bypass Redis reverse pointer is still not fully lifecycle-safe. The PR stores:
  - `bypass:extraTime:${userId}:${apps.join(',')}`

  and gives it a TTL, but I still do not see symmetric cleanup when the underlying bypass policy is deleted early, removed as broken data, or otherwise disappears before the key expires. The system can still resolve a stale pid and operate on dead state.
- **Medium:** the updated bypass-quota merge logic regressed from the prior revision. When an existing bypass policy already has consumed time beyond its stored `extraQuota`, the old version adjusted `quota` upward to preserve the current effective state. That corrective logic is now gone, and the update path only increments `extraQuota`. This can skew remaining quota/accounting for an already-active bypass policy.
- **Medium:** bypass state management is still spread across many subsystems:
  - `AccessRequestManager`
  - `AppTimeUsageManager`
  - `PolicyManager2`
  - `Block`
  - `Bypass`
  - `dnsmasq`
  - `AccessRequestSensor`

  For something this stateful, I still do not see regression coverage for create/update/expire/delete/reenforce flows, which keeps the change high-risk.

### 💡 Suggestions
- Fix `matchTarget()` so single-app lookups explicitly check `policy.targets` membership when the rule is multi-target.
- Add hard cleanup for the `bypass:extraTime:...` Redis key whenever:
  - the bypass policy is deleted
  - the bypass policy expires and is auto-deleted
  - the stored pid no longer resolves
- Restore the quota-adjustment behavior when updating an existing bypass policy, or add a clearer invariant proving the current accounting remains correct after partial usage.
- Add regression tests for:
  - single-app lookup not matching unrelated multi-target rules
  - exact multi-app matching
  - bypass create/update/expire/delete flows
  - stale reverse-pointer cleanup
  - `affectedPids` synchronization on policy activation/deactivation
  - `onQuotaReached === "unenforce"` enforcing/unapplying symmetrically

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8951  
Head SHA: `46e62113c9bdfdb301530db8e1cd5ab1cba0173a`  
Checked at: 2026-04-21 11:31:05 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is stronger than the earlier one:
  - bypass policies are now tracked separately via `active_bypass_policy`
  - `PolicyManager2` can load active bypass policies directly
  - `sortedActiveRulesCache` now attempts to suppress block/app_block rules that are currently bypassed instead of relying only on lower-layer iptables/dnsmasq behavior
- The general architecture is becoming more coherent:
  - `AccessRequestManager` creates/updates bypass policies
  - `AppTimeUsageManager` enforces/unenforces them based on quota
  - `PolicyManager2` now knows how to activate/deactivate and enumerate them
  - `AccessRequestSensor` tries to keep `affectedPids` in sync with policy lifecycle
- The time-limits API changes are also directionally better:
  - multi-app rules are exposed as app-set strings instead of `policy:<pid>`
  - bypass policy quota is surfaced back into the listing path

- No linked issue found.

### ⚠️ Issues found
- **High:** `matchTarget()` is still wrong for single-app lookups against multi-target rules. When `app` is a single value, the function only validates `policy.target` and never checks whether that app is actually present in `policy.targets`. So a multi-target rule can still incorrectly match a single-app lookup simply by falling through to `true`.
- **High:** the Redis reverse pointer for bypass policies is still not lifecycle-safe. The code creates:
  - `bypass:extraTime:${userId}:${apps.join(',')}`

  and gives it a TTL, but I still do not see symmetric cleanup when:
  - the underlying bypass policy is deleted early
  - the policy disappears as broken/stale data
  - `pm2.getPolicy(bypassPolicyId)` returns null before the Redis key expires

  That means later lookups can still resolve stale bypass state.
- **Medium:** the update path for an existing bypass policy still regresses quota/accounting compared with the earlier revision. It now only increments `extraQuota`, but does not adjust `quota` when the current usage has already moved past the previously stored effective quota. That can leave remaining quota calculations inconsistent for already-active bypass policies.
- **Medium:** the new `sortedActiveRulesCache` bypass filtering adds another stateful layer, but I don’t see proof that cache invalidation is fully symmetric for all bypass-policy changes. Since `isBypassed(rule)` depends on:
  - active bypass policy membership
  - `affectedPids`
  - `appTimeUsed`
  - `extraQuota` / expiry

  any stale cache here can make policy visibility diverge from actual runtime enforcement.
- **Medium:** bypass matching/identity is still keyed primarily by `userId + sorted apps`, which is convenient but fragile if overlapping rule scopes evolve or multiple logically distinct bypass contexts share the same app set.

### 💡 Suggestions
- Fix `matchTarget()` so single-app queries explicitly check membership in `policy.targets` when the rule is multi-target.
- Add hard cleanup for `bypass:extraTime:...` whenever:
  - a bypass policy is deleted
  - a bypass policy expires and auto-deletes
  - a referenced bypass pid no longer resolves
- Restore the quota-adjustment logic for existing bypass policies, or otherwise prove the accounting invariant with tests.
- Audit cache invalidation around `sortedActiveRulesCache` for all of:
  - bypass creation
  - bypass update
  - quota exhaustion
  - bypass unenforce
  - bypass deletion/expiry
  - single-app vs multi-target matching
  - bypass create/update/expire/delete
  - stale reverse-pointer recovery
  - `affectedPids` lifecycle updates
  - cache correctness when a bypass becomes active/inactive

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8951  
Head SHA: `a46e7079d9ac78561350281a39c0da3c61b6a33f`  
Checked at: 2026-04-21 11:46:10 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is architecturally better than the earlier version in a few important ways:
  - bypass handling is now modeled as an explicit policy action (`action: "bypass"`) instead of trying to disable/shadow block rules directly
  - `PolicyManager2` now tracks active bypass policies separately, which is a cleaner foundation for enforcement and matching
  - `AppTimeUsageManager` gaining `onQuotaReached === "unenforce"` is a sensible mechanism for “temporary allowance until quota is consumed”
  - `getAppsFromPolicy()` and the tighter multi-app handling in `matchTarget()` are directionally better than the old single-target assumptions
- The overall intent is good:
  - preserve the original block/disturb rules
  - add a time-bounded bypass overlay
  - automatically re-enforce the original rules when the allowance is consumed or expires

- No linked issue found.

### ⚠️ Issues found
- **High:** `matchTarget()` is still not safe for single-app matching against multi-target policies. In the new logic:
  - multi-app requests validate exact membership/length
  - but the single-app branch only checks `policy.target`
  - if `policy.targets` exists and `policy.target` does not, the function falls through to `return true`

  That means a rule with `targets: ["youtube", "tiktok"]` can still incorrectly match a request for just `"youtube"`. This is still a blocker because it can attach the wrong rules to an access request.
- **High:** the Redis reverse-pointer lifecycle for bypass policies still looks incomplete. The key:
  - `bypass:extraTime:${userId}:${apps.join(',')}`

  is created/updated on approval, but this diff does not show symmetric cleanup when the bypass policy is:
  - deleted manually
  - auto-deleted on expiry
  - removed through other policy cleanup paths

  If the pointer survives while the policy is gone or replaced, later approvals can resolve stale IDs and update the wrong/nonexistent policy.
- **High:** bypass quota merge/update behavior is still fragile. The code mixes:
  - request-level `leftQuota`
  - current app usage from `AppTimeUsageManager`
  - rule-level `appTimeUsage.quota/extraQuota`
  - existing bypass policy state

  but the merge logic is not obviously monotonic or idempotent across repeated approvals in the same day. In particular, `quotaLeft < totoalQuotaLeft` style adjustments can drift when:
  - app usage changes between approvals
  - an existing bypass already has extra quota
  - multiple requests for overlapping app sets are approved
- **High:** bypass identity is still keyed primarily by `userId + sorted apps`. That is too coarse. Different underlying rule sets affecting the same user/app combination can collapse onto the same bypass policy even if:
  - the matched blocked rules changed
  - schedule context changed
  - enforcement scope differs

  This can cause old bypass state to be reused in situations where the effective protected rules are no longer the same.
- **Medium:** `sortedActiveRulesCache` bypass filtering still risks divergence. The cache filters active rules by consulting current bypass policies, but I don’t see a fully symmetric invalidation story for:
  - bypass create
  - bypass update
  - bypass unenforce
  - bypass delete
  - expiry-triggered auto-delete

  Any missed invalidation means alarm/rule matching can continue using stale “bypassed vs not bypassed” state.
- **Medium:** `affectedPids` lifecycle is still underdefined. The bypass policy stores `affectedPids`, but updates replace that set with `Array.from(affectedPids)` from the current approval pass. That risks:
  - dropping previously affected rules still intended to stay bypassed
  - or reusing stale ones depending on request order
- **Medium:** `onQuotaReached === "unenforce"` is directionally right, but the enforcement symmetry is still delicate. In `refreshPolicy()` the bypass gets re-applied when usage is below quota and unapplied when above quota, which is correct in spirit, but this now depends on the timing and consistency of:
  - `appTimeUsed` updates
  - bypass policy refresh
  - original rule enforcement state

  That path needs stronger proof/tests because any lag can leave the original block rule in the wrong state.
- **Low:** there is an unused import:
  - `const { include } = require('underscore');`
- **Low:** typo/naming issues like `TARGET_PERFIX_LEN` / `totoalQuotaLeft` / `calaculateQuotaLeft` make already-complex logic harder to audit.

### 💡 Suggestions
- Fix `matchTarget()` so single-app requests do **not** match multi-target rules unless the policy explicitly represents exactly that single app.
- Add explicit reverse-pointer cleanup for `bypass:extraTime:${userId}:${apps.join(',')}` on every bypass-policy removal path:
  - manual delete
  - auto-expire delete
  - policy cleanup helpers
- Revisit bypass identity so it is not keyed only by `userId + sorted apps`; it should reflect the actual protected rule set or another stable identity boundary.
- Make `affectedPids` merge/update semantics explicit:
  - either union with existing still-valid affected rules
  - or prove replacement is always correct
- Ensure `sortedActiveRulesCache` is invalidated symmetrically on all bypass lifecycle transitions.
- Add regression tests for:
  - single-app request vs multi-target rule non-match
  - repeated approvals for same user/apps on same day
  - bypass policy expiry and Redis pointer cleanup
  - `affectedPids` updates across multiple approvals
  - original block rule re-enforcement after bypass quota is consumed
  - iptables/dnsmasq symmetry for bypass create/update/delete

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8951  
Head SHA: `89453fd55201e94a1509e7a67ed8ed1a3b664bd4`  
Checked at: 2026-04-21 19:01:22 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The overall direction makes sense: instead of pausing/disabling the original time-limit/block rules, this revision tries to model approved extra time as an explicit temporary **bypass** policy tied back to the affected rules. That is a cleaner abstraction than mutating multiple unrelated policies in ad hoc ways.
- A few implementation details are solid:
  - `affectedPids` gives the bypass rule an explicit linkage to the blocked/disturb rules it is overriding.
  - adding `onQuotaReached === "unenforce"` support in `AppTimeUsageManager` is a reasonable primitive for temporary bypass semantics.
  - storing bypass policies in a separate active zset can make lookup/cleanup cheaper than scanning all policies every time.
  - `getAppsFromPolicy()` is a useful helper for normalizing single-target vs multi-target app rules.

  - No linked issue found.

### ⚠️ Issues found
- **High:** `matchApp()` no longer checks policy action up front, which broadens `findMatchingTimeLimitRules()` in a dangerous way.
  - Previously `matchApp()` rejected anything that was not `app_block`, `block`, or `disturb`.
  - Now it only checks tag/target/schedule, and the caller later splits matching rules into `bypassPolicies` vs `blockOrDisturbPolicies`.
  - That means unrelated actions with matching tags/targets/schedules can now be considered “matching time limit rules,” which changes semantics and makes the matching helper unsafe for reuse.

- **High:** bypass detection in `PolicyManager2.getSortedActiveRules()` only checks `affectedPids` and remaining quota, but not whether the bypass rule actually matches the same tag/scope/target context as the candidate rule.
  - If a bypass policy references a pid and still has quota left, the original rule is removed from active rules globally.
  - That is too coarse for a rule engine: it assumes pid membership alone is sufficient to bypass all effective contexts of that policy.
  - If the referenced policy is broader than the bypass rule’s intended tag/device/app context, this can under-enforce blocking.

- **High:** the new bypass policy creation path can silently change rule semantics for multi-rule approvals.
  - `approveRequest()` aggregates `affectedPids`, computes one `apps` array from the request, then creates a single bypass rule with:
    - `type` hardcoded to `"category"` or `"mac"` for internet,
    - `target` based only on `apps[0]`,
    - `targets` based only on request apps, not the original policy shapes.
  - If affected policies differ in type/shape/targets, one synthetic bypass policy may not be semantically equivalent to bypassing each original rule individually.

- **Medium:** `findMatchingTimeLimitRules()` changed return type from `Policy[]` to `{ bypassPolicies, blockOrDisturbPolicies }`, which is a breaking API change in a core helper.
  - In the shown diff only one main caller was updated, but this function is exported and may have other call sites.
  - Without updating all consumers, this change can cause runtime bugs that are easy to miss in JS.

- **Medium:** `approveRequest()` now passes `includeDisabled: true` into `findMatchingTimeLimitRules()`, but that option is still translated into `pm2.loadActivePoliciesAsync({ includingDisabled: includeDisabled })`.
  - The local variable is named `includeDisabled`, but the downstream API expects `includingDisabled`.
  - This was already awkward before, and the new logic relies on it more heavily. If that contract drifts even slightly, disabled rules stop participating and bypass creation becomes inconsistent.

- **Medium:** `removeFromRefrencedBypassPolicies()` updates or deletes bypass policies when a referenced rule is deleted, but it does not update the Redis lookup key created by `AccessRequestManager.getBypassPolicyKey(...)`.
  - If the bypass policy gets deleted because `affectedPids` becomes empty, the cached key can still point to a deleted pid until TTL expiry.
  - That creates stale lookup state and can cause approval flows to try to reuse nonexistent policies.

- **Medium:** the new chain-management logic in `PolicyManager2` creates `FW_<pid>_BYPASS` chains during enforce of block/disturb-like policies, but the lifecycle is tightly coupled to bypass support and now spread across multiple paths.
  - This is pretty invasive networking behavior for a PR that also changes access-request semantics.
  - I don’t see guardrails here for partial failures between chain creation, bypass application, and cleanup. A failed mid-flight update risks leaving orphaned chains or inconsistent policy state.

- **Low:** there are several correctness/maintainability nits in `AccessRequestManager` that make this code harder to trust:
  - `TARGET_PERFIX_LEN` is misspelled
  - `totoalQuotaLeft` / `calaculateQuotaLeft` are misspelled
  - `include` is imported from `underscore` but unused
  - `getRequest()` deletes `req.note` / `req.mac`, but the request object stores `reason` / `deviceMac`, so this cleanup looks stale or wrong

- **Low:** the checked-in built asset rename (`index-Bge_N4b4.js` → `index-Drdu33fs.js`) is noise in a backend-heavy PR.
  - It makes review harder and increases churn without helping validate the logic changes shown here.

### 💡 Suggestions
- Restore the action filter inside `matchApp()`, or make the helper name/contract explicit if it is now intended to match more than block/disturb time-limit rules.
- Rework bypass suppression in `getSortedActiveRules()` so it validates more than `affectedPids` + quota-left. At minimum, confirm the bypass rule still matches the same tag/scope/target dimensions before suppressing the original rule.
- Be careful about collapsing multiple affected rules into one synthetic bypass policy. If original rules can differ materially, keep per-rule bypass semantics or normalize more explicitly.
- Audit all call sites of `findMatchingTimeLimitRules()` since its return type changed.
- When deleting or replacing a bypass policy, also clean/update the corresponding `bypass:extraTime:*` Redis key to avoid stale pid pointers.
- Add targeted tests for:
  - multi-app approvals against mixed single-target and multi-target rules
  - deletion of a referenced block rule while a bypass policy is active
  - bypass quota expiry causing original rule reactivation
  - `getSortedActiveRules()` behavior when a bypass only partially overlaps an affected rule
- If possible, split generated frontend asset updates from policy-engine changes in future PRs.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8951  
Head SHA: `f48edefe472fa5923b9b967eb717c2fcadf51d8b`  
Checked at: 2026-04-22 14:01:29 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The refactor is aiming at a cleaner model for extra-time approvals:
  - instead of pausing/disabling original rules directly, it introduces an explicit temporary **bypass** policy tied to the affected rule set
  - `affectedPids` gives the bypass policy a clear linkage back to the original block/disturb rules
  - `onQuotaReached: "unenforce"` in `AppTimeUsageManager` is a sensible primitive for “temporary allow until quota is consumed”
- A few helper additions are useful in isolation:
  - `getAppsFromPolicy()` improves normalization of `target` vs `targets`
  - `getBypassPolicyKey()` makes the Redis lookup intent explicit
  - tracking `leftQuota` on the request object is directionally useful for approval math

  - No linked issue found.

### ⚠️ Issues found
- **High:** `matchApp()` no longer filters by action, which makes rule matching too broad.
  - Previously this helper only matched `app_block`, `block`, and `disturb`.
  - Now it matches any tagged/scheduled policy with matching targets, and the caller filters actions later.
  - That is dangerous because `findMatchingTimeLimitRules()` is exported and now has a much looser contract than its name suggests.

- **High:** bypass suppression in `PolicyManager2.getSortedActiveRules()` is too coarse.
  - A rule is treated as bypassed if any active bypass policy still has quota left and `affectedPids.includes(rule.pid)`.
  - That ignores whether the bypass policy still matches the same tag/scope/target context as the original rule.
  - In practice this can under-enforce: a broad rule may be removed from active rules globally just because one bypass policy references its pid.

- **High:** `approveRequest()` collapses all affected rules into one synthetic bypass policy using the request’s app list, which can change semantics.
  - The bypass policy shape is derived from `req.app`, not reconstructed from each original rule.
  - If the affected policies differ in type/targets/scope details, a single bypass policy may not faithfully override all of them.
  - This is especially risky for mixed single-target vs multi-target rules.

- **Medium:** `findMatchingTimeLimitRules()` changed return type from an array to an object:
  - old: `Policy[]`
  - new: `{ bypassPolicies, blockOrDisturbPolicies }`
  - That is a breaking API change for a shared exported helper. If any other caller still expects an array, this will fail at runtime.

- **Medium:** stale bypass-policy Redis pointers are still possible.
  - `AccessRequestManager` stores `bypass:extraTime:<user>:<apps> -> pid`.
  - But when referenced rules are removed and `removeFromRefrencedBypassPolicies()` deletes the bypass policy, this lookup key is not cleaned up immediately.
  - That can leave a stale pid until TTL expiry and cause reuse/update paths to look up a nonexistent policy.

- **Medium:** `getRequest()` cleanup logic appears inconsistent with the stored request schema.
  - Requests are created with `deviceMac` and `reason`.
  - The read path deletes `req.note` and `req.mac`, which looks stale and likely ineffective.
  - That suggests this code path was adapted from an older schema and not fully updated.

- **Low:** there are several maintainability issues that reduce confidence in a policy-engine change of this size:
  - unused import: `include` from `underscore`
  - comment typo: `matchingPolices`
  - renamed typo fixed in one place, but overall this is still a large behavioral change mixed with low-signal churn
- **Low:** the built frontend asset rename is review noise in a backend-heavy policy PR and makes the diff harder to audit.

### 💡 Suggestions
- Restore the action filter inside `matchApp()`, or split the helper into:
  - a generic target/schedule matcher
  - and a time-limit rule matcher with explicit action constraints
- Tighten bypass suppression in `getSortedActiveRules()` so it validates more than `affectedPids` and quota-left.
  - At minimum, confirm the bypass rule still matches the same tag/scope/target dimensions before suppressing the original rule.
- Reconsider collapsing multiple affected rules into one bypass policy.
  - If original rules can differ materially, keep per-rule bypass policies or normalize the source rules more explicitly.
- Audit all call sites of `findMatchingTimeLimitRules()` since its return type changed.
- When a bypass policy is deleted or replaced, also clean/update the `bypass:extraTime:*` Redis key immediately.
- Fix `getRequest()` to clean fields that actually exist in the current request schema, or remove that dead cleanup entirely.
- Add targeted tests for:
  - mixed single-target and multi-target approvals
  - deleting a referenced rule while a bypass policy is active
  - stale Redis bypass-key reuse
  - `getSortedActiveRules()` with partially overlapping bypass scope

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8951  
Head SHA: `ef328a4e31fa4dbb1f150d496d43ec9e3748f760`  
Checked at: 2026-04-22 16:32:09 CST

jasonlyc (approved):

PR #8952, opened 2026-04-20
fix(iperf3): DMZ bypass for Wi-Fi test in bridge mode

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The bridge-mode problem statement makes sense, and the fix is targeted:
  - bypass DMZ DNAT in `FW_PREROUTING_DMZ_HOST`
  - allow the traffic in `FW_INPUT_ACCEPT`
- Persisting the bypass port in `BYPASS_PORT_FILE` is a practical way to clean up stale rules even if the iperf3 process dies before `iperf3:stop`.
- The start path is reasonably defensive:
  - remove lingering stale rules first
  - if bypass setup fails, stop iperf3 and clear state
- The inline comments explain the networking rationale clearly, which is useful here.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** `iperf3:start` cleans up only the previously stored bypass port before starting a new server, but it does not stop an existing live iperf3 process first. If a prior process is still running and its actual listening port differs from the stored file state, this flow can leave the old process alive while creating a new bypass mapping assumption around the current PID/port lookup. The sensor seems to assume one server instance and one authoritative stored port, but the lifecycle is not fully reconciled.
- **Medium:** the bypass rules are inserted with a generic fixed comment `"iperf3 wifi test"` and removed by exact rule deletion. If duplicate identical rules are inserted for the same port through repeated starts/races, cleanup may not fully restore the chain to the expected state. There is no explicit idempotency guard beyond the port file.
- **Low:** `readStoredBypassPort()` treats invalid file content as `null` but does not delete the bad file. That means a corrupt port file can persist indefinitely and silently disable stale-rule cleanup on future runs.
- **Low:** `removeBypassRules()` logs failures as errors even during benign cleanup cases where the rule is already absent. That can create noisy logs for normal recovery paths.

### 💡 Suggestions
- Ensure the lifecycle is single-instance safe:
  - either stop any existing iperf3 server before starting a new one,
  - or explicitly verify that the existing PID/port is the one represented by the stored bypass state.
- Consider making the bypass rule identity more explicit, e.g. include the port in the comment, so diagnostics and cleanup are easier.
- If `readStoredBypassPort()` sees invalid content, delete the file so future runs can recover cleanly.
- Downgrade expected “rule already absent” cleanup failures from error to warn/debug if possible.
- Add tests or scripted validation for:
  - bridge-mode start/stop happy path
  - crash/restart cleanup using `BYPASS_PORT_FILE`
  - invalid stored port file
  - repeated `iperf3:start` calls without an intervening stop

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8952  
Head SHA: `6dc0fe54f631889eb139235a5ab109d299534b0b`  
Checked at: 2026-04-20 17:15:58 CST

jasonlyc (approved):

PR #8953, opened 2026-04-20
add cache in HostTool

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The refactor is directionally good:
  - more MAC-entry access now goes through `HostTool` instead of open-coded Redis operations
  - cache invalidation is wired into most write paths
  - helper methods like `incrFieldInMAC`, `deleteMac`, `updateSpoofingDataInMAC`, and `updateKeysInMAC` reduce duplicated Redis logic
- `Host.saveWithCompare()` is a useful optimization idea:
  - avoids unnecessary writes
  - skips ultra-frequent heartbeat-only updates for `lastActiveTimestamp`
  - keeps the active timestamp logic centralized
- Switching `DNSManager.resolveMac()` and `SpooferManager.loadManualSpoof()` to `hostTool.getMACEntry()` improves consistency.
- The cache-invalidation event (`MSG_HOST_MAC_ENTRY_CACHE_INVALIDATE`) is a sensible addition for cross-process coherence.

- No linked issue found.

### ⚠️ Issues found
- **High:** `HostTool.getMACEntry()` now caches MAC entries for **1 day** (`MAC_ENTRY_CACHE_TTL_MS = 86400000`). That is too long for host state that changes frequently:
  - spoofing flags
  - IP bindings
  - activity timestamps
  - security alarm counts
  - manual spoof state

  Even with invalidation on many local writes, a long TTL increases the blast radius of any missed invalidation path and can leave consumers reading stale host state for hours.
- **Medium:** `saveWithCompare()` uses loose inequality (`!=`) to decide changed fields:
  - `if (obj[key] != obj2[key])`

  Redis values are strings, while `redisfy()` may produce numbers/booleans. That means some semantically different states can compare equal or vice versa in surprising ways. For a write-suppression layer, this kind of coercive comparison is risky.
- **Medium:** `updateSpoofingDataInMAC()` can skip writes entirely if the cached `spoofing` boolean is considered equal, even though the payload also includes timestamp fields:
  - `spoofingTime`
  - `unspoofingTime`

  The shortcut only checks `hash.spoofing` vs cached `spoofing`, so it can suppress timestamp updates when the spoofing state value itself does not change. That looks like a correctness bug.
- **Low:** `Host.saveWithCompare()` updates `this.o.lastActiveTimestamp` when it skips a heartbeat write, but does not similarly preserve `lastFrom` if the skipped update included both `lastActiveTimestamp` and `lastFrom`. That can leave in-memory state partially divergent from the skipped persisted state.
- **Low:** the new cache prune interval is never obviously cleared, which is fine for a singleton process lifetime but worth noting if test/runtime environments recreate instances.

### 💡 Suggestions
- Reduce the MAC-entry cache TTL substantially, or make it configurable. For host-state metadata, minutes is much safer than a full day.
- Tighten `saveWithCompare()` field comparison:
  - compare normalized string forms for Redis-backed fields, or
  - use field-aware comparisons for booleans/numbers/timestamps.
- Fix `updateSpoofingDataInMAC()` so it only short-circuits when **all** relevant fields are unchanged, not just `spoofing`.
- Consider adding tests for:
  - cached read after local update/invalidation
  - missed invalidation fallback behavior
  - spoofing timestamp updates when spoofing state is unchanged
  - `saveWithCompare()` number/string/boolean comparison cases
  - skip-window behavior for `lastActiveTimestamp` + `lastFrom`

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8953  
Head SHA: `34c6bb44b393a00812163649b1f850f63af3d38a`  
Checked at: 2026-04-20 19:31:08 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Centralizing MAC-entry reads/writes behind `HostTool` is a solid refactor:
  - fewer ad hoc Redis mutations
  - easier cache invalidation
  - better consistency across `Host`, `DNSManager`, `SpooferManager`, and alarm counting
- The cache invalidation fanout via `MSG_HOST_MAC_ENTRY_CACHE_INVALIDATE` is the right shape for multi-process consumers.
- `saveWithCompare()` is a useful optimization in principle. Skipping redundant host writes can reduce a lot of Redis churn on noisy host-heartbeat updates.
- Moving `deleteMac()` to also clean up `REDIS_KEY_HOST_ACTIVE` in one place is cleaner than duplicating that responsibility at call sites.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** `macEntryCache` now keeps host metadata for **1 hour**:
  - `this.macEntryCache = new LRU({max: 500, maxAge: 3600 * 1000});`

  That is a long TTL for mutable fields like spoofing state, IP metadata, activity timestamps, backup names, etc. The invalidation event helps, but if any write path misses invalidation or an event is dropped, consumers can read stale host state for a long time.
- **Medium:** `saveWithCompare()` still uses loose inequality:
  - `if (obj[key] != obj2[key])`

  Redis-backed values often differ by type (`1` vs `"1"`, `true` vs `"true"`). Using loose comparison in a generic persistence path is risky because it can hide real state transitions or make persistence behavior depend on incidental type coercion.
- **Medium:** when the heartbeat-style write is skipped, the code only restores `this.o.lastActiveTimestamp` from Redis, not `lastFrom`. In the `onlyLastActiveAndFrom` case, this can leave the in-memory object diverged from persisted state.
- **Medium:** `updateSpoofingDataInMAC()` can suppress timestamp updates incorrectly. If cached `spoofing` already matches the incoming boolean, it returns early and skips writing:
  - `spoofingTime`
  - `unspoofingTime`

  That means callers can lose a legitimate timestamp refresh.
- **Low:** `pruneExpiredMacEntryCache()` is introduced but not used anywhere in this diff. If expiry is only opportunistic in this LRU implementation, that may be okay, but right now this helper looks unused.

### 💡 Suggestions
- Reduce the MAC-entry cache TTL unless there is strong evidence that a 1-hour staleness window is acceptable for all host fields.
- Replace loose `!=` comparison in `saveWithCompare()` with explicit normalization or strict comparison on normalized values.
- In the skip-write path, keep both `lastActiveTimestamp` and `lastFrom` aligned with Redis.
- Rework `updateSpoofingDataInMAC()` so it only skips when the full effective payload is unchanged, not just the boolean spoofing flag.
  - Redis/string vs runtime/typed value comparisons in `saveWithCompare()`
  - skipped writes preserving `lastFrom`
  - spoofing timestamp updates when spoofing state does not change
  - cache invalidation across multi-process read/write flows

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8953  
Head SHA: `d44972f4f149e090f686c4b7ecf06c433baafdb0`  
Checked at: 2026-04-21 17:02:52 CST

jasonlyc (approved):

PR #8954, opened 2026-04-21
Support regular expression in target list block rules

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The regex support is being added in the right layers, not just one:
  - parsing in `CategoryEntry`
  - storage/caching in `CategoryUpdater`
  - dnsmasq enforcement in `createCategoryMappingFile()`
  - app-side attribution in `PolicyManager2.matchPolicy()`
- Centralizing regex validation in `CategoryEntry.isValidDomainRE()` is a good cleanup. It avoids the previous duplicated validation logic in `PolicyManager2`.
- The `CategoryUpdateSensor` changes are thoughtful:
  - regex entries are peeled off before IP/hash/domain classification
  - regex state is flushed on strategy transitions
  - unsupported filter/BF strategy behavior is explicitly handled
- Mirroring regex matching in `matchPolicy()` is important. Without that, alarm attribution would diverge from what dnsmasq actually blocks.

- No linked issue found.

### ⚠️ Issues found
- **High:** the `domain_re` apply path can still report success after `dnsmasq.addPolicyFilterEntry()` fails. In `PolicyManager2`:
  - the call is wrapped in `.catch((err) => { log.error(...) })`
  - on failure, `flag` becomes `undefined`
  - `if (flag !== "skip_restart") dnsmasq.scheduleRestartDNSService();` still runs
  - `skipFinalApplyRules = true` is still set
  
  So a failed dnsmasq write can still look like a successfully applied policy from the caller’s perspective. For a correctness-sensitive policy path, that is a real bug.
- **Medium:** regex support is now part of enforcement, storage, and alarm matching, but I don’t see any regression coverage in this diff for the new lifecycle:
  - parse `regex:...`
  - persist to Redis
  - render into dnsmasq `re-match`
  - match in `matchPolicy()`
  - flush on category disable / strategy transition
  
  Given how cross-cutting this is, lack of tests is risky.
- **Low:** `getCompiledRegexDomains()` recompiles and caches per category, which is fine, but `matchPolicy()` now does regex evaluation on every category match attempt. For large target lists this may become noticeably more expensive than the old exact/wildcard-only path, especially if regex-heavy categories start being used more broadly.

### 💡 Suggestions
- Make `domain_re` application fail closed:
  - if `dnsmasq.addPolicyFilterEntry()` throws, return immediately instead of continuing
  - only schedule restart and set `skipFinalApplyRules = true` when the add actually succeeded
- Add regression tests for:
  - `CategoryEntry.parse("regex:...")`
  - invalid regex rejection
  - dnsmasq config generation with `re-match`
  - `matchPolicy()` matching regex category members
  - regex flush on category disable / filter-strategy transition
- If regex target lists may grow, consider measuring or bounding regex count per category so app-side matching cost stays predictable.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8954  
Head SHA: `637ff0ca1b165df226449df1a7b74fe14830ef8a`  
Checked at: 2026-04-21 16:47:06 CST

jasonlyc (approved):

PR #8958, opened 2026-04-22
add intelDnsCache in IntelTool.js

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The optimization direction makes sense:
  - maintaining an in-memory negative cache for `inteldns:*` keys can cut a lot of repeated Redis `HGETALL` misses on busy boxes
  - updating the cache on successful `addDomainIntel()` is the right basic write-through behavior
  - removing a key from the cache when `hgetall` comes back empty is a reasonable self-healing path
- Restricting the preload/refresh behavior to `firewalla.isMain()` is also sensible.
- Using a lock around cache reload/update avoids obvious concurrent mutation races inside the process.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** the cache is still only partially invalidated. It is refreshed every 6 hours and updated on `addDomainIntel()`, but I don’t see corresponding invalidation for other mutation paths such as delete/rename/expiry-driven cleanup outside this code path. That means the in-memory set can remain stale for a long time and suppress legitimate Redis reads for newly-created keys until the next refresh if they are written by another process/path.
- **Medium:** `_reloadIntelDnsCache()` does a full `scanResults('inteldns:*')` and replaces the whole set. That may be acceptable at this scale, but it is still a fairly blunt operation and should be watched on larger datasets or slower boxes.
- **Low:** the startup delay means the cache is `null` for the first minute, so behavior changes after warm-up. That is probably fine, but it does mean performance characteristics are bimodal and can make debugging “why was this a Redis hit earlier but not later?” less obvious.
- **Low:** the cache only tracks key existence, not emptiness/TTL freshness, so it is strictly a best-effort negative cache. That’s okay, but it should be documented clearly because it is easy to over-assume its correctness from the name.

### 💡 Suggestions
- Add invalidation/update hooks for every domain-intel mutation path, not just `addDomainIntel()`.
- If other processes can write these Redis keys, consider either:
  - shortening the periodic refresh interval
  - or subscribing to a pub/sub / event signal for cache invalidation
- Add a comment near `getDomainIntel()` clarifying that this is a negative existence cache only, not a source of truth.
- A regression test would help for:
  - miss cached as absent
  - later key created by another path/process
  - lookup still succeeds after intended invalidation/refresh

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8958  
Head SHA: `7947324dbe955ce1f0b150d7803c1908fc51b5e2`  
Checked at: 2026-04-22 11:46:01 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The intent is solid: adding a negative in-memory cache for `inteldns:*` avoids repeated Redis lookups for known-missing domain intel on busy boxes.
- The implementation updates the cache on write (`addDomainIntel`) and self-heals on stale positives (`getDomainIntel` removes keys when Redis no longer has data), which is better than a pure startup snapshot.
- Scoping the background preload to `firewalla.isMain()` is the right choice and avoids duplicating this cache across non-main processes.
- Guarding cache mutations with `AsyncLock` is a reasonable way to avoid inconsistent Set updates from concurrent refresh/add/delete paths.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** the startup `setTimeout(async () => { ... setInterval(async () => { ... }) })` has no error handling.
  - If `_reloadIntelDnsCache()` throws during the first delayed load, the interval is never installed.
  - That leaves the process permanently without this cache until restart, which is a fragile failure mode for a background optimization.

- **Medium:** the cache can remain stale for up to 6 hours for deletes that happen outside `getDomainIntel()`.
  - `addDomainIntel()` adds keys, and `getDomainIntel()` removes stale ones when that exact domain is queried.
  - But if an `inteldns:*` key expires or is deleted elsewhere and never gets queried again, the Set keeps reporting it as present until the next full reload.
  - That does not break correctness badly, but it weakens the “negative cache” assumption and still causes unnecessary Redis reads on stale positives for long periods.

- **Medium:** `scanResults('inteldns:*')` every 6 hours can still be a noticeable full-keyspace operation on constrained boxes.
  - The comment says ~1k keys / ~10 KB on one high-traffic box, but that is observational rather than guaranteed.
  - Without a bound or telemetry, this optimization may shift cost from many point reads to periodic full scans on larger deployments.

- **Low:** `getDomainIntel()` now has asymmetric behavior before vs after cache initialization.
  - During the first 60 seconds, all misses still hit Redis.
  - After initialization, known-absent keys return early.
  - That is probably acceptable, but the performance profile changes sharply based on uptime, which may complicate debugging.

- **Low:** the lock protects Set replacement/mutation, but reads of `this.intelDnsCache` happen unlocked.
  - In Node this is unlikely to corrupt memory, but it does mean readers can observe the old Set while a reload is replacing it.
  - Since this is just a performance cache, that’s probably fine, but it’s worth calling out as eventual-consistency rather than strict consistency.

### 💡 Suggestions
- Wrap the delayed preload and periodic reload in `try/catch`, and ensure the interval is installed even if the first reload fails.
- Consider recording reload failures and cache size in metrics/logs so it’s easier to confirm this optimization is working in production.
- If stale positives are still a concern, consider:
  - a shorter refresh interval,
  - or a lightweight TTL/version on the in-memory snapshot,
  - or only using the Set as a negative cache after a successful full refresh timestamp check.
- If `scanResults('inteldns:*')` might grow over time, add a note or metric guard so the operational cost can be monitored rather than assumed.
- Add tests for:
  - preload failure not disabling future refreshes,
  - expired/deleted Redis key causing Set eviction on read,
  - `addDomainIntel()` updating the cache when initialized,
  - correct fallback behavior before cache initialization.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8958  
Head SHA: `6c741ef10b14a9b9b24a9a364f3b9037f347a390`  
Checked at: 2026-04-22 18:02:50 CST

jasonlyc (approved):

PR #8959, opened 2026-04-22
wait for category init before ipset restore

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The startup ordering fix is directionally sound:
  - waiting for category/country initialization before `blockControl.finishInitialization()` addresses a real race where the first atomic ipset swap can otherwise happen before data is ready
  - adding a bounded timeout avoids turning this into an unbounded startup hang
- Marking dns-only categories as initialized in `CategoryUpdater` is a good catch; without that, the new wait loop would have blocked on categories that never call `recycleIPSet()`.
- Moving `initializedCategories` into `CategoryUpdaterBase` is a clean way to share the initialization state across both category and country updaters.
- The logging around pending initialization is useful for debugging slow startup and data-source failures.

- No linked issue found.

### ⚠️ Issues found
- **Medium:** `const initWait = fc.getConfig().timing['policy.category.init_wait'] || 120` treats falsy configured values as unset. That is usually fine, but it also means an explicit `0` cannot disable the wait. If `0` is ever intended as “don’t wait,” this implementation will silently ignore it.
- **Medium:** the new wait loop checks only “active but not initialized,” but it does not distinguish between “still loading” and “already failed permanently.” In failure cases, startup now always burns the full timeout window before proceeding, which may be acceptable, but it is a blunt failure mode.
- **Low:** `initializedCategories[category] = true` is now written from multiple places (`CategoryUpdater`, `CountryUpdater`, and comments imply `recycleIPSet` also participates). That works, but it makes the lifecycle a bit diffuse and easier to get wrong later if another path forgets to update the flag.
- **Low:** `pending.join(', ')` in repeated logs can get noisy if many categories are active; not wrong, just something to watch at startup.

### 💡 Suggestions
- If disabling the wait via config is useful, consider using nullish handling instead of `||`, e.g. preserve an explicit `0`.
- Consider separating “loading,” “initialized,” and “failed” states instead of a single boolean map. That would let startup stop waiting early on terminal failures while still surfacing which categories failed.
- It may be worth centralizing the “mark initialized” transition in one helper to keep the state lifecycle easier to reason about.
- A regression test around startup ordering would be valuable here:
  - active ipset-backed category finishes late → startup waits, then proceeds
  - dns-only category initializes without `recycleIPSet()` → startup does not stall indefinitely
  - failing remote source → startup resumes after timeout

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8959  
Head SHA: `899f1896c70d85419a9e4610e17ac845d63e58c6`  
Checked at: 2026-04-22 12:15:58 CST

jasonlyc (approved):

PR #8960, opened 2026-04-22
adds nchist

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Nice small utility with a clear, focused scope: inspect `history:networkConfig`, show snapshots, and diff two versions.
- The interface is easy to understand:
  - no args → history listing
  - one ref → show config
  - two refs → diff
  - `-u` → unified diff
- Using `subprocess.run([...], check=True)` avoids shell-injection issues.
- The path-based diff output is practical for large JSON blobs; the digest summary in list view is also useful for quick browsing.
- The implementation is readable and self-contained, which is exactly what a script like this should be.

  - No linked issue found.

### ⚠️ Issues found
- **Low:** score parsing assumes Redis zset scores are always integer strings.
  - `score = int(out[i + 1])` will fail if the sorted-set score is ever emitted as a float string (which Redis allows).
  - If this history key ever moves from integer-second timestamps to fractional timestamps, the script breaks immediately.

- **Low:** `cmd_show()` prints the raw stored JSON string instead of normalized/prettified JSON.
  - That makes output less readable and less stable for manual inspection, especially if stored values are minified or inconsistently formatted.

- **Low:** list/resolve behavior overloads a numeric ref as “seq if `< len(entries)` else timestamp”.
  - In practice this works for real Unix timestamps, but it is still an ambiguous API contract.
  - It would be safer to make the distinction explicit or at least document that small numeric timestamps cannot be addressed directly.

### 💡 Suggestions
- Parse scores more defensively, e.g. accept float-form Redis scores and normalize before display/comparison.
- Pretty-print JSON in `cmd_show()` so the single-entry view is easier to read.
- Consider an explicit ref syntax such as `seq:<n>` / `ts:<n>` if you want to remove the numeric ambiguity entirely.
- Optional polish: catch `json.JSONDecodeError` and print a clearer error if a stored entry is malformed rather than letting the traceback bubble up.

### Verdict
APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8960  
Head SHA: `f0e724121d3306c9d99078c53505337e79ace9dc`  
Checked at: 2026-04-22 13:01:02 CST

jasonlyc (approved):

PR #8963, opened 2026-04-22
Enhancement implementation

melo-crypto:
write currently-using family protect dns server to config when it hav
e no server specified

add dns server test api for family protect

https://github.com/firewalla/firecommit/issues/7611

### DIFF
--- a/net2/HostManager.js
+++ b/net2/HostManager.js
@@ -726,8 +726,11 @@ module.exports = class HostManager extends Monitorable {
       extdata['portforward'] = portforwardConfig;
 
     const fpp = await sensorLoader.initSingleSensor('FamilyProtectPlugin');
-    const familyConfig = await fpp.getFamilyConfig()
-    if (familyConfig) extdata.family = familyConfig
+    const familyConfig = await fpp.getFamilyConfig();
+    log.debug(`FamilyConfig: ${JSON.stringify(familyConfig)}`);
+    const effectiveServers = familyConfig && familyConfig.servers && familyConfig.servers.length > 0
+      ? familyConfig.servers : await fpp.familyDnsAddr();
+    extdata.family = Object.assign({}, familyConfig, { servers: effectiveServers });
 
     const ruleStatsPlugin = await sensorLoader.initSingleSensor('RuleStatsPlugin');
     const initTs = await ruleStatsPlugin.getFeatureFirstEnabledTimestamp();
--- a/sensor/FamilyProtectPlugin.js
+++ b/sensor/FamilyProtectPlugin.js
@@ -37,6 +37,8 @@ let FAMILY_DNS = null;
 const fs = require('fs');
 const Promise = require('bluebird');
 Promise.promisifyAll(fs);
+const dns = require('dns');
+const util = require('util');
 
 const DNSMASQ = require('../extension/dnsmasq/dnsmasq.js');
 const Constants = require('../net2/Constants.js');
@@ -125,6 +127,37 @@ class FamilyProtectPlugin extends Sensor {
         sem.sendEventToFireMain({
           type: 'FAMILY_RESET',
         })
+      });
+      
+      extensionManager.onGet('familyDnsTest', async (_msg, data) => {
+        const maxServers = 10;
+        const servers = data && data.servers;
+        const domains = data && data.domains;
+        if (!Array.isArray(servers) || servers.length === 0)
+          throw new Error("servers is required");
+        if (servers.length > maxServers)
+          throw new Error(`servers exceeds limit of ${maxServers}`);
+        if (!Array.isArray(domains) || domains.length === 0)
+          throw new Error("domains is required");
+
+        const testOne = (server, domain) => new Promise((resolve) => {
+          const resolver = new dns.Resolver();
+          resolver.setServers([server]);
+          const resolve4Async = util.promisify(resolver.resolve4.bind(resolver));
+          const timer = setTimeout(() => {
+            resolver.cancel();
+            resolve({ domain, addresses: [], error: 'timeout' });
+          }, 5000);
+          resolve4Async(domain).then(
+            (addresses) => { clearTimeout(timer); resolve({ domain, addresses }); },
+            (err) => { clearTimeout(timer); resolve({ domain, addresses: [], error: err.message }); }
+          );
+        });
+
+        return Promise.all(servers.map(async (server) => {
+          const results = await Promise.all(domains.map(domain => testOne(server, domain)));
+          return { server, results };
+        }));
       });
     }
 
--- a/sensor/FamilyProtectPlugin.js
+++ b/sensor/FamilyProtectPlugin.js
@@ -129,8 +129,13 @@ class FamilyProtectPlugin extends Sensor {
         })
       });
       
-      extensionManager.onGet('familyDnsTest', async (_msg, data) => {
+      extensionManager.onCmd('familyDnsTest', async (msg, data) => {
         const maxServers = 10;
+        const maxDomains = 20;
+        // 64 bytes per server (IPv6+port worst case), 253 bytes per domain (RFC 1035 max FQDN)
+        const maxPayloadBytes = maxServers * 64 + maxDomains * 253 + 1000;
+        if (data && JSON.stringify(data).length > maxPayloadBytes)
+          throw new Error(`payload exceeds limit of ${maxPayloadBytes} bytes`);
         const servers = data && data.servers;
         const domains = data && data.domains;
         if (!Array.isArray(servers) || servers.length === 0)
@@ -139,6 +144,8 @@ class FamilyProtectPlugin extends Sensor {
           throw new Error(`servers exceeds limit of ${maxServers}`);
         if (!Array.isArray(domains) || domains.length === 0)
           throw new Error("domains is required");
+        if (domains.length > maxDomains)
+          throw new Error(`domains exceeds limit of ${maxDomains}`);
 
         const testOne = (server, domain) => new Promise((resolve) => {
           const resolver = new dns.Resolver();
--- a/test/test_netbot.js
+++ b/test/test_netbot.js
@@ -757,13 +757,13 @@ describe('test familyDnsTest', function() {
   });
 
   async function dnsTest(value) {
-    return call({ mtype: 'get', type: 'jsonmsg', data: { item: 'familyDnsTest', value }, target: '0.0.0.0' });
+    return call({ mtype: 'cmd', type: 'jsonmsg', data: { item: 'familyDnsTest', value }, target: '0.0.0.0' });
   }
 
   async function dnsTestRaw(value) {
     return netbot.msgHandler(gid, {
       mtype: 'msg',
-      message: { from: 'test', obj: { mtype: 'get', type: 'jsonmsg', data: { item: 'familyDnsTest', value }, target: '0.0.0.0' }, appInfo: { deviceName: 'test' }, type: 'jsondata', suppressLog: true }
+      message: { from: 'test', obj: { mtype: 'cmd', type: 'jsonmsg', data: { item: 'familyDnsTest', value }, target: '0.0.0.0' }, appInfo: { deviceName: 'test' }, type: 'jsondata', suppressLog: true }
     });
   }
 
--- a/test/test_netbot.js
+++ b/test/test_netbot.js
@@ -746,3 +746,50 @@ describe('test netbot', function(){
   });
 
 });
+
+describe('test familyDnsTest', function() {
+  this.timeout(30000);
+
+  before(async function() {
+    const sl = require('../sensor/APISensorLoader.js');
+    await sl.initSensors(netbot.eptcloud);
+    sl.run();
+  });
+
+  async function dnsTest(value) {
+    return call({ mtype: 'get', type: 'jsonmsg', data: { item: 'familyDnsTest', value }, target: '0.0.0.0' });
+  }
+
+  async function dnsTestRaw(value) {
+    return netbot.msgHandler(gid, {
+      mtype: 'msg',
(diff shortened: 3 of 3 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The host-side change is directionally useful: returning effective Family Protect DNS servers in `HostManager` should make the UI/config surface reflect what the box is actually using, instead of exposing an empty `servers` list.
- The new `familyDnsTest` hook is scoped as a read/test path and does not mutate policy state, which keeps the change relatively contained.
- The implementation uses a per-server `dns.Resolver()` instance rather than mutating global DNS process state, which is the right approach.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `extdata.family` is now always populated, even when `familyConfig` is falsy.
  - `Object.assign({}, familyConfig, { servers: effectiveServers })` will produce `{ servers: undefined }` when config is missing and `familyDnsAddr()` returns nothing.
  - Previously the field was omitted entirely when there was no config. That is a behavior/API change and may break consumers that treat presence of `extdata.family` as “feature configured”.

- **Medium:** the new debug log can leak more configuration detail than necessary.
  - `log.debug(\`FamilyConfig: ${JSON.stringify(familyConfig)}\`)` dumps the full family config into logs.
  - Even if this is “just DNS config”, it increases noise and may expose user/environment-specific settings unnecessarily.

- **Medium:** `familyDnsTest` has no input bound on `servers` or `domains`.
  - A caller can submit arbitrarily large arrays and trigger `servers × domains` DNS queries concurrently.
  - Since this is an extensionManager RPC handler, that opens the door to avoidable resource pressure and long event-loop stalls on the box.

- **Medium:** the timeout does not cancel the underlying DNS lookup.
  - `Promise.race([... , timeout])` only stops awaiting; the resolver query can still continue in the background.
  - With enough inputs, timed-out requests can stack up and create hidden load even after the caller already got a timeout result.

- **Low:** only `resolve4` is used.
  - If the purpose is validating whether the family DNS servers work for configured domains, IPv6-only or AAAA-preferred cases will be reported as failures even when resolution is actually healthy.
  - That may be acceptable, but it should be intentional.

- **Low:** `familyDnsAddr()` is awaited in `HostManager` whenever configured servers are absent.
  - If this method is slow or does I/O, `toJson`-style host data generation becomes a little more expensive on a frequently-read path.
  - Not a blocker from this diff alone, but worth validating.

### 💡 Suggestions
- Preserve the old response shape unless config truly exists, e.g. only set `extdata.family` when either:
  - `familyConfig` exists, or
  - there is a deliberate requirement to expose computed default servers as config.
- Replace the full config debug log with a minimal summary, such as server count or whether fallback DNS was used.
- Add hard limits for `familyDnsTest`, for example:
  - max servers
  - max domains
  - optional per-request concurrency cap
- If timeout behavior matters operationally, consider an approach that can actually stop work or at least serialize/bound the outstanding lookups.
- Clarify whether IPv4-only validation is intentional. If not, consider checking both A and AAAA, or document the scope as IPv4-only DNS reachability.
- Add tests for:
  - no `familyConfig`
  - empty `familyConfig.servers` with fallback from `familyDnsAddr()`
  - oversized `servers/domains` input rejection
  - timeout result formatting for one failing domain among otherwise successful queries

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8963  
Head SHA: `e537ed73abcd06aec6fbc3e713e823e00b45058f`  
Checked at: 2026-04-22 18:02:05 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is better than the previous one.
  - Adding a hard cap on `servers` in `familyDnsTest` meaningfully reduces the worst-case fanout.
  - Switching from `Promise.race(...timeout...)` to `resolver.cancel()` on timeout is a real improvement; it stops treating timeout as purely cosmetic and actively tears down the DNS request.
- The host-side fallback logic is still directionally useful:
  - returning effective Family Protect DNS servers helps the UI/config surface reflect actual runtime behavior rather than an empty configured list.
- Using per-request `dns.Resolver()` instances remains the right choice instead of mutating global resolver state.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `extdata.family` is still always populated, even when there is no actual family config.
  - `Object.assign({}, familyConfig, { servers: effectiveServers })` will still create a truthy object even if `familyConfig` is nullish.
  - Previously, `extdata.family` was omitted when no config existed. That is an API/shape change and may affect consumers that use field presence as a signal.

- **Medium:** the debug log still dumps the full family config.
  - `log.debug(\`FamilyConfig: ${JSON.stringify(familyConfig)}\`)` is noisy and exposes more config detail than needed.
  - A summarized log would be safer and easier to operate.

- **Medium:** only `servers` is bounded; `domains` is still unbounded.
  - The new cap reduces one axis, but the total work is still `servers × domains`.
  - A large `domains` array can still create a heavy burst of DNS lookups and timer objects.

- **Low:** the validation remains IPv4-only.
  - `resolve4` will report failure for IPv6-only domains or environments where AAAA is the expected path.
  - That may be acceptable if this is intentionally “A-record reachability only,” but it should be explicit.

### 💡 Suggestions
- Preserve the old response shape unless there is explicit config to expose.
  - If the goal is to show effective runtime DNS, consider adding a separate field rather than always materializing `extdata.family`.
- Replace the full config log with something like:
  - whether configured servers were present
  - how many effective servers were returned
- Add a limit for `domains` too, not just `servers`.
- If IPv4-only testing is intentional, document it in the handler name/comment or response shape to avoid confusing false negatives.
- Add tests for:
  - no `familyConfig`
  - fallback to `familyDnsAddr()`
  - `servers` over limit
  - timeout path returning `{ error: 'timeout' }`
  - resolver cancellation not hanging the request

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8963  
Head SHA: `e2963630d8e6a5fbff117f406eddeb5509155d8c`  
Checked at: 2026-04-22 19:15:50 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is stronger than the earlier one because it now includes tests for the new `familyDnsTest` handler:
  - happy path with valid server/domain
  - missing `servers`
  - missing `domains`
  - server-count limit enforcement
- The DNS test implementation itself is directionally solid:
  - per-request `dns.Resolver()` instances avoid touching global resolver state
  - timeout now actively calls `resolver.cancel()`, which is better than a passive `Promise.race(...)` timeout
  - limiting servers to 10 reduces one important axis of potential fanout
- Returning effective Family Protect DNS servers from `HostManager` is still a useful UI-facing improvement.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `extdata.family` is still always populated, even when no Family Protect config exists.
  - `Object.assign({}, familyConfig, { servers: effectiveServers })` creates a truthy object even if `familyConfig` is missing.
  - Previously, `extdata.family` was omitted when there was no config, so this is an API shape change that may affect callers relying on field presence.

- **Medium:** the debug log still dumps the full family config.
  - `log.debug(\`FamilyConfig: ${JSON.stringify(familyConfig)}\`)` is noisy and may expose more runtime configuration detail than necessary.
  - A summarized log would be safer and easier to operate.

- **Medium:** `domains` is still unbounded.
  - The new cap only limits `servers`, but total work is still `servers × domains`.
  - A large `domains` array can still create a substantial burst of DNS lookups, timers, and result objects.

- **Low:** the tests depend on external DNS reachability.
  - `8.8.8.8` + `www.google.com` makes the happy-path test more integration-style than unit-style.
  - That may be acceptable in this suite, but it can become flaky in restricted CI or offline environments.

- **Low:** the handler is still IPv4-only.
  - `resolve4` treats IPv6-only domains as failures.
  - That may be intended, but if the endpoint is presented as generic DNS validation, the behavior could be misleading.

### 💡 Suggestions
- Preserve the old response shape unless there is explicit Family Protect config.
  - If the goal is to expose runtime-effective DNS, consider a separate field rather than always materializing `extdata.family`.
- Replace the full-config debug log with a brief summary, such as:
  - whether configured servers were present
  - how many effective servers were returned
- Add a limit for `domains` too, not just `servers`.
- If possible, make the tests less network-dependent by mocking resolver behavior, or at least keep one mocked unit test alongside the current integration-style coverage.
- Clarify in comments/docs whether `familyDnsTest` is intentionally IPv4-only.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8963  
Head SHA: `2fdbaf90f78d2bec12ed165a3992d7f77ad69ce3`  
Checked at: 2026-04-22 20:31:00 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is stronger than the earlier one:
  - it adds both `maxDomains` and a payload-size limit, which closes the previous “servers capped but domains unbounded” gap
  - using `resolver.cancel()` on timeout is still the right approach
  - the new tests cover the basic validation/error paths and the success path
- Moving the extension hook from `onGet` to `onCmd` can also make sense if this endpoint is intended as an explicit active test action rather than a passive read.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** the test code still calls the endpoint as a `get`/`jsonmsg`, but the implementation changed from `extensionManager.onGet('familyDnsTest', ...)` to `extensionManager.onCmd('familyDnsTest', ...)`.
  - Unless the message-routing layer maps that exact request shape to `onCmd`, these tests are no longer exercising the code path that was changed.
  - More importantly, the real caller contract may now be inconsistent with the documented/tested invocation pattern.

- **Medium:** `extdata.family` is still always populated in `HostManager`, even when no Family Protect config exists.
  - `Object.assign({}, familyConfig, { servers: effectiveServers })` creates a truthy object even when `familyConfig` is absent.
  - Previously `extdata.family` was omitted unless config existed, so this is still an API-shape change.

- **Medium:** the full config debug log is still too noisy.
  - `log.debug(\`FamilyConfig: ${JSON.stringify(familyConfig)}\`)` dumps the entire config object into logs.
  - That is more detail than this path likely needs.

- **Low:** the payload-byte limit is approximate and based on `JSON.stringify(data).length`.
  - That is acceptable as a guardrail, but it is not a protocol-level byte count and may reject/allow slightly differently than the comment suggests.
  - Not a blocker, just something to be aware of.

- **Low:** the happy-path test still depends on public DNS reachability (`8.8.8.8`, `www.google.com`).
  - That makes it more integration-style and potentially flaky in restricted CI.

### 💡 Suggestions
- Make sure the test/request shape matches the new handler registration:
  - if `familyDnsTest` is now a command, update the client invocation and tests accordingly
  - otherwise keep it on `onGet`
- Preserve the old `extdata.family` response shape unless you explicitly want to expose “effective runtime servers” even without config.
- Replace the full family-config debug log with a short summary, like whether fallback DNS was used and how many servers were returned.
- If you keep the payload guard, consider adding one test for the payload-size rejection path too.
- If practical, mock the resolver for at least one unit test so the suite is not fully dependent on external DNS.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8963  
Head SHA: `09071f6acf71e4079f74c346b975c575c781d869`  
Checked at: 2026-04-23 11:16:21 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision fixes one of the earlier concerns: the tests now invoke `familyDnsTest` as a `cmd`, which matches the new `extensionManager.onCmd('familyDnsTest', ...)` registration.
- Adding both a domain-count cap and an overall payload-size guard is a good improvement over the earlier version. It prevents the obvious `servers × domains` blow-up from becoming unbounded.
- Using a dedicated `dns.Resolver()` per test and calling `resolver.cancel()` on timeout is the right direction for keeping long DNS waits from lingering indefinitely.
- The validation-path tests are useful and now line up with the updated handler contract.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `extdata.family` is still always materialized, even when there was previously no Family Protect config.
  - Before this change, `extdata.family` was only present when `getFamilyConfig()` returned something truthy.
  - Now `Object.assign({}, familyConfig, { servers: effectiveServers })` always creates an object, so callers can no longer distinguish “feature/config absent” from “feature present with fallback servers”.
  - That is an API-shape change in `HostManager`, not just a display enhancement.

- **Low:** the debug log is still too verbose.
  - `log.debug(\`FamilyConfig: ${JSON.stringify(familyConfig)}\`)` dumps the full config object into logs every time this path runs.
  - That is noisy, and if more fields get added later it becomes even less appropriate.

- **Low:** the happy-path test still depends on public DNS and external network reachability.
  - `8.8.8.8` + `www.google.com` makes this more of an integration test than a deterministic unit test.
  - It may be flaky in CI environments with restricted outbound DNS.

- **Low:** `familyDnsTest` remains IPv4-only because it calls `resolve4`.
  - That may be intentional, but if the feature is meant to validate “family DNS works” more generally, it currently ignores IPv6-only or IPv6-preferred resolution behavior.

### 💡 Suggestions
- Preserve the previous `extdata.family` contract unless the API change is intentional and coordinated with consumers.
  - For example, only set `extdata.family` when `familyConfig` exists, and expose fallback servers through a separate field if needed.
- Replace the full JSON config log with a concise summary, such as whether fallback servers were used and how many were returned.
- Add at least one deterministic test with a mocked resolver so the suite does not rely entirely on public DNS.
- If IPv4-only resolution is intended, document that in the handler or test name; otherwise consider whether this should test both A and AAAA resolution paths.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8963  
Head SHA: `85f540a821260e3bbddfad641625388453dd3cbe`  
Checked at: 2026-04-23 11:46:23 CST

jasonlyc (approved):

PR #8964, opened 2026-04-22
fix racing condition when in memory host object changes

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This refactor fixes a real debouncing problem: `mergeAndSave()` now accumulates pending updates per MAC instead of letting later calls overwrite earlier in-memory changes before the timer fires.
- Moving the merge point to `_mergeAndSave(mac, pending)` is directionally cleaner:
  - event handlers no longer mutate `host.o.detect` directly
  - write coalescing is centralized in one place
  - the `DetectUpdate` path now behaves more like the other producers
- The Firewalla-device fast path is also cleaner now:
  - it writes only the intended `nameBased` update
  - it no longer depends on mutating `host.o.detect` before scheduling the merge

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** nested source metadata is now merged incorrectly for repeated `DetectUpdate` calls.
  - In the old code, updates were applied directly to `host.o.detect[from]`, so keys like ``${key}.source`` could coexist alongside the base field.
  - In the new code, `mergeAndSave()` does `entry.updates[key] = Object.assign({}, entry.updates[key], updates[key])`, and `_mergeAndSave()` does another shallow `Object.assign`.
  - That means multiple updates to the same `from` bucket can clobber nested objects rather than merge them deeply. If two detectors update different subfields under the same source before the debounce fires, the earlier one can be lost.

- **Medium:** `DetectUpdate` now mutates the incoming `detect` object by appending `*.source` fields directly onto it.
  - If the emitter reuses that object or logs/inspects it elsewhere, this introduces side effects outside the sensor.
  - Previously the mutation happened on `host.o.detect[from]`; now it happens on the event payload object itself.

- **Medium:** `mergeAndSave(mac)` with no updates now always schedules a timer even when nothing changed.
  - `NewDeviceFound` / `RegularDeviceInfoUpdate` call `this.mergeAndSave(mac)` unconditionally.
  - `_mergeAndSave()` then re-reads the host and may still persist because of name-based inference, but if there is no detect state and no derived `nameBased` type, the timer/work was unnecessary.
  - Not a blocker, but this path can create avoidable churn under frequent device updates.

- **Low:** there is a subtle behavioral change when the host disappears during the debounce window.
  - The old code captured the `host` object up front.
  - The new code re-fetches via `hostManager.getHostFastByMAC(mac)` at execution time and silently drops the update if the host is gone.
  - That may be acceptable, but it is a changed delivery guarantee for transient devices.

### 💡 Suggestions
- If multiple detectors can update the same `from` bucket, switch from shallow merge to a deeper merge strategy for `detect` payloads, especially for metadata keys like `foo.source`.
- Avoid mutating the incoming `detect` object from the event; build a copied object before appending `*.source` fields.
- Consider short-circuiting `mergeAndSave(mac)` when there are no explicit updates and there is no likely derived work to do, to reduce timer churn on noisy device-update events.
- Add regression tests for:
  - two `DetectUpdate` events for the same MAC/source within the debounce window
  - preservation of both data fields and `*.source` metadata
  - host disappearing before the delayed merge executes
  - Firewalla MAC path plus later UA update coalescing correctly

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8964  
Head SHA: `9d7e093c0ee6dcefbcb602920217af44e0538683`  
Checked at: 2026-04-22 19:00:59 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision improves the earlier refactor.
  - Cloning `detect` into `sub` before adding `*.source` metadata removes the unwanted mutation of the incoming event payload.
  - Centralizing debounced updates by MAC is still the right direction; it avoids the old pattern where intermediate writes depended on mutating `host.o.detect` before the timer fired.
- The Firewalla-device path is cleaner now:
  - it schedules only the intended `nameBased` update
  - it no longer relies on pre-populating `host.o.detect`
- The new `mergeAndSave(mac, updates)` structure makes the coalescing intent much clearer than the old “capture host object and flush later” flow.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** pending updates for the same `from` bucket are still merged only shallowly.
  - `entry.updates[key] = Object.assign({}, entry.updates[key], updates[key])`
  - and later `detect[key] = Object.assign({}, detect[key], updates[key])`
  - If two `DetectUpdate` events for the same `from` arrive within the debounce window and both contain nested objects, later updates can still overwrite earlier nested structure rather than merge it deeply.
  - This is better than mutating the live host object directly, but it still risks losing partial detector state.

- **Medium:** `mergeAndSave(mac)` with no updates still schedules work for every `NewDeviceFound` / `RegularDeviceInfoUpdate`.
  - That means a timer, host lookup, and merge attempt even when there is no explicit detect delta.
  - It may still be justified for name-based inference, but on noisy device-update paths this can create avoidable churn.

- **Low:** the new debug log may be too verbose for a hot path.
  - `log.debug('mergeAndSave', mac, entry.updates)` can emit full pending detect payloads repeatedly for active networks.
  - Helpful during development, but probably noisy in production.

- **Low:** the host disappearance behavior is still changed from the old implementation.
  - The new code re-fetches the host at flush time and silently drops pending updates if the host no longer exists.
  - That may be acceptable, but it is a behavioral change worth being explicit about.

### 💡 Suggestions
- If detector payloads can contain nested objects, consider using a deep merge for accumulated `updates` and final `detect` application, not just `Object.assign`.
- Consider short-circuiting `mergeAndSave(mac)` when there are no explicit updates and no likely derived work to do, or at least document that empty calls are intentionally used for deferred name-based inference.
- Downgrade or remove the new `mergeAndSave` debug log before merge unless this path really needs payload-level tracing.
- Add regression tests for:
  - two `DetectUpdate` events for the same MAC/from within the debounce window
  - preservation of both base fields and `*.source` metadata
  - empty `mergeAndSave(mac)` calls not causing unnecessary persistence
  - host removal before delayed flush

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8964  
Head SHA: `78202d3b43e07bbd99536b2be13d62c5df2d5b83`  
Checked at: 2026-04-22 19:30:56 CST

jasonlyc (approved):

PR #8965, opened 2026-04-23
check app timeusage only for VideoAlarm and GameAlarm

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a clean, narrowly scoped change.
- The intent is reasonable: only run the “related app time usage” logic for alarm types that are actually app-category driven.
- The early return improves readability and avoids unnecessary downstream checks for unrelated alarm types.
- The allowlist is explicit, so the behavior change is easy to understand from the code.

  - No linked issue found.

### ⚠️ Issues found
- **Low:** this is still a behavior change, not just a refactor.
  - Before this patch, every alarm type flowed through the existing `p.dest.app.id` logic.
  - After this patch, all non-`ALARM_VIDEO` / non-`ALARM_GAME` alarms immediately return `true`.
  - If any other alarm type was intentionally relying on this method to suppress or gate behavior based on app usage, that logic is now silently bypassed.

- **Low:** the supported types are hardcoded locally inside the method.
  - That is fine today, but it creates a maintenance trap if additional app-usage-based alarm types are introduced later.
  - Someone updating alarm taxonomy could easily miss this branch and get incorrect behavior with no obvious failure.

- **Low:** there is no accompanying test coverage in the diff.
  - For a method that now short-circuits on alarm type, I would want at least one regression test showing the old path still applies to `ALARM_VIDEO` / `ALARM_GAME` and is skipped for other types.

### 💡 Suggestions
- Add focused tests for:
  - `ALARM_VIDEO` using the existing app-id path
  - `ALARM_GAME` using the existing app-id path
  - a non-target alarm type short-circuiting to `true`
- If this reflects a product rule, add a brief comment explaining why only these two alarm types participate in related app time-usage checks.
- Consider moving the type list to a shared alarm classification constant if the same distinction exists elsewhere.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8965  
Head SHA: `fe70263195324027bfd03dfabb4a1d9bdb51fc37`  
Checked at: 2026-04-23 11:45:50 CST

jasonlyc (approved):

PR #8970, opened 2026-04-23
support IP-based malware autoblock in active protect(strict mode)

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a focused change with a clear end-to-end intent:
  - detect strict-mode inbound IP intel hits in `FlowMonitor`
  - tag the alarm with `p.blockby = "ip_intel"`
  - generate an IP-based block policy in `AlarmManager2`
  - preserve that policy through auto-block revalidation
- The plumbing is coherent across the three touched files.
- The `AlarmManager2` change is especially important and looks correct:
  - forcing `p.type = "ip"` and `p.target = alarm["p.dest.ip"]` for `ip_intel` avoids relying on the existing `dest.name === dest.ip` heuristic
  - that makes the policy decision explicit instead of incidental
- Updating `AutoBlocksRevalidationSensor` to recognize `ip_intel` alongside `fastdns` is also the right follow-through. Without that, strict-mode IP auto-blocks would be treated inconsistently on revalidation.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `isActiveProtectStrictMode()` checks only the feature flag and raw `policy:system.dns_proxy`, but not whether the parsed policy is actually the active policy for the current traffic path.
  - If `dns_proxy` is enabled globally but not meaningfully governing the flow being inspected, this may over-apply strict-mode IP autoblock behavior.
  - The naming suggests “Active Protect strict mode”, but the implementation is effectively “dns_proxy policy has strict=true”.

- **Medium:** `checkIpAlarm()` now auto-blocks any inbound intel-category flow in strict mode without additional confidence gating.
  - The condition is:
    - strict mode enabled
    - `flowObj.fd === 'in'`
    - `flowObj.intel.category === 'intel'`
  - That is a fairly broad trigger.
  - If `intel.category === 'intel'` is noisy or includes lower-confidence cases, this could lead to overly aggressive IP autoblocks.

- **Low:** `isActiveProtectStrictMode()` silently swallows JSON parse/config errors.
  - Returning `false` is safe, but it also hides bad `policy:system` state and makes debugging harder if strict mode unexpectedly stops applying.

### 💡 Suggestions
- It would help to clarify the policy coupling in `isActiveProtectStrictMode()`:
  - if `dns_proxy.strict` is truly the product-level source of truth for strict mode, a short comment would make that clearer
  - otherwise consider naming the helper more narrowly
- If available elsewhere in the codebase, consider reusing a shared strict-mode / Active Protect policy helper instead of parsing `policy:system` inline here.
- I’d also consider logging parse failures at debug/warn level in `isActiveProtectStrictMode()` so broken policy state is easier to diagnose.
- Finally, if strict IP autoblock is intentionally broad, a short comment near the new `checkIpAlarm()` branch explaining the confidence model would help future reviewers.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8970  
Head SHA: `0320dae9734434b79cde6c41a92393baad76f0d8`  
Checked at: 2026-04-23 19:31:17 CST

monitor/FlowMonitor.js
```diff
@@ -924,6 +924,18 @@ module.exports = class FlowMonitor {
     return true;
   }
 
+  async isActiveProtectStrictMode() {
+    if (!fc.isFeatureOn("dns_proxy")) return false;
+    try {
+      const raw = await rclient.hgetAsync("policy:system", "dns_proxy");
```
jasonlyc: Use HostManager.getPolicyFast("dns_proxy")

jasonlyc (changes requested):

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The intent is reasonable: when Active Protect strict mode auto-blocks based on IP intel, the generated policy should stay IP-scoped instead of falling back to domain-based targeting.
- The changes are small and consistent across the pipeline:
  - `FlowMonitor` tags the alarm with `p.blockby = "ip_intel"`
  - `AlarmManager2` converts that into an IP policy
  - `AutoBlocksRevalidationSensor` recognizes the new auto-block origin
- Reusing the existing `blockby` pattern is a good choice. It keeps the change localized instead of introducing a parallel flag just for this path.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `isActiveProtectStrictMode()` is declared `async` even though it does not await anything.
  - That is not a correctness bug by itself, but it obscures the fact that this is just a synchronous config read.
  - It also adds avoidable promise churn on a hot path inside flow monitoring.

- **Medium:** the strict-mode IP auto-block condition in `FlowMonitor.checkIpAlarm()` is narrower than the policy conversion logic in `AlarmManager2`.
  - The trigger requires:
    - strict mode on
    - `flowObj.fd === 'in'`
    - `flowObj.intel.category === 'intel'`
  - But `AlarmManager2` only looks at `p.blockby === 'ip_intel'` and immediately forces an IP rule.
  - That means the semantics of “ip_intel” are spread across files and not self-validating. If another path ever sets `p.blockby = 'ip_intel'` without the same assumptions, policy generation will silently change behavior.

- **Medium:** the `flowObj.intel.category === 'intel'` check is pretty opaque and brittle.
  - `category` usually carries a taxonomy value, so comparing it to the literal string `"intel"` is not very descriptive.
  - If upstream intel classification names evolve, this path can quietly stop firing.

- **Low:** `AlarmManager2` copies `alarm["p.blockby"]` into `p.blockby`, but only special-cases one value.
  - That is fine for now, but it continues the pattern of “magic string” control flow without centralization.
  - This is starting to look like a place where enum-like constants would help.

### 💡 Suggestions
- Make `isActiveProtectStrictMode()` synchronous unless there is a near-term need for async behavior.
- Add or update tests for:
  - strict mode inbound IP-intel alarm produces `type = "ip"` and `target = p.dest.ip`
  - non-strict mode still follows the existing domain/IP fallback behavior
  - auto-block revalidation preserves `ip_intel` rules the same way it already preserves `fastdns`
- Consider replacing the literal `"intel"` / `"ip_intel"` strings with shared constants to reduce drift across `FlowMonitor`, `AlarmManager2`, and `AutoBlocksRevalidationSensor`.
- If `flowObj.intel.category === 'intel'` is the intended contract, add a short comment explaining what upstream data shape guarantees that value.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8970  
Head SHA: `52eb9912b390aa0e040fbd1f6002cd89340916b0`  
Checked at: 2026-04-27 10:16:36 CST

jasonlyc (approved):

PR #8971, opened 2026-04-24
fix a bug that conntrack cache is not removed with redis key

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a small, targeted cleanup.
- Replacing manual `unlinkAsync([...conntrack.getKey(...)...])` calls with `conntrack.delConnEntries(...)` is a good direction if the goal is to centralize conntrack-key deletion logic in one helper.
- The intent of the code remains clear:
  - for blocked UDP IP flows, clear conntrack state so later Zeek logs do not get recorded against stale flow state
- From a maintainability perspective, using the helper is better than re-encoding key construction at the call site.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** this change may broaden the deletion scope compared with the previous implementation.
  - The old code deleted exactly two explicit keys:
    - `src:sport -> dst:dport`
    - `dst:dport -> src:sport`
  - The new code delegates to `conntrack.delConnEntries(...)` twice, but the diff does not show whether that helper deletes exactly one key, both directions, or additional derived entries.
  - If `delConnEntries(...)` already removes both forward and reverse keys, the second call may be redundant. If it removes wildcard/family-related entries too, this is no longer a pure refactor.
  - So correctness depends entirely on the helper’s semantics.

- **Low:** this introduces two awaited delete operations instead of one batched unlink call.
  - The previous code issued one Redis unlink with both keys at once.
  - The new code may result in two round trips unless `delConnEntries(...)` batches internally.
  - That is probably negligible here, but it is still a small regression if the helper is just wrapping a single-key delete.

### 💡 Suggestions
- I’d verify and document `conntrack.delConnEntries(...)` semantics before merging:
  - does one call delete only the exact directional key?
  - both directions?
  - any related wildcard/timeout keys?
- If one call already handles both directions, drop the second call to avoid redundant work.
- If it does not, a short comment here would help explain why both calls are still required after the helper migration.
- If performance matters on this hot path, it may also be worth confirming the helper preserves the old non-blocking/batched unlink behavior.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8971  
Head SHA: `02a85fe7afde5e6c53dc2f305ae89a4e051ebdd3`  
Checked at: 2026-04-24 11:00:52 CST

jasonlyc (approved):

PR #8972, opened 2026-04-24
fix bug for Link-local addr is blocked incorrectly

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The direction here is sensible: introducing a network-level `list:set` wrapper and switching callers to `getNetListIpsetName(uuid)` is a clean way to make policy/rule code cover:
  - IPv4 network subnets
  - regular IPv6 subnets
  - link-local IPv6 subnets
  without duplicating call-site logic everywhere.
- The `NetworkProfile` changes are fairly comprehensive.
  - The new `c_net_all_*_set` is created alongside the existing per-family sets.
  - The list is populated with v4, v6, and link-local-v6 child sets.
  - The various “off” / monitored / tag membership paths were updated to include the new link-local set, which is the right follow-through.
- Splitting FE80 prefixes into a dedicated `hash:net,iface` set is also the right systems-level move.
  - Link-local IPv6 matching really is interface-scoped, so this is more correct than stuffing everything into a plain `hash:net` set.
- Updating policy/rule consumers like DNS redirect, bypass, block, AP isolation, NTP redirect, and policy manager to consume the aggregate set reduces future drift and is easier to maintain.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** several updated call sites now assign the same `list:set` name to both IPv4 and IPv6 variables, which may be incompatible with how the surrounding rule builders expect family-specific sets.
  - Example pattern:
    - `remoteSet4 = NetworkProfile.getNetListIpsetName(target);`
    - `remoteSet6 = NetworkProfile.getNetListIpsetName(target);`
  - If the iptables/ip6tables rule generation layer simply emits `--match-set <name> ...`, this may work because `list:set` can contain family-appropriate child sets.
  - But if any surrounding logic assumes `remoteSet4` and `remoteSet6` represent distinct family-specific set names for tuple counting, diagnostics, or later mutation, this is now a semantic change, not just a refactor.
  - The diff makes this look intentional, but it is worth re-checking the callers that use `remoteTupleCount = 2`; the underlying model is now effectively “one aggregate logical set”, not two family sets.

- **Medium:** `getNetListIpsetName(uuid)` creates a `list:set` and then adds child sets every time `ensureCreateEnforcementEnv()` runs, but the diff does not show protection against duplicate `Ipset.add(...)` behavior.
  - If `Ipset.add(list:set, member)` is not idempotent or logs noisy errors on duplicate members, repeated env creation could create churn or misleading logs.
  - This is probably okay if the wrapper suppresses “already added” cases, but the correctness depends on that helper behavior.

- **Medium:** the new link-local matching path depends on `realIntf` being valid when building `hash:net,iface` entries, but the diff does not show a guard around empty/undefined interface values before:
  - `executeIpsetCommands(netLinklocalIpsetName6, fe80Subnets, \`,${realIntf}\`)`
  - If `realIntf` is missing or stale during profile update, link-local entries could be malformed or silently wrong.
  - Given that link-local IPv6 is interface-scoped, this detail matters more than for ordinary subnet sets.

- **Low:** the new example-less call-site simplification obscures IPv4/IPv6 intent in a few places.
  - For instance, `set4Name` and `set6Name` in `APFeaturesPlugin` now point to the same aggregate set.
  - That is probably correct mechanically, but the names are now misleading and make the code harder to read.

### 💡 Suggestions
- Re-check the callers that still conceptually model “v4 set + v6 set” after switching to `getNetListIpsetName(...)`.
  - If both variables intentionally point to the same aggregate set, consider renaming them in follow-up cleanup so the code reflects the new model.
- Confirm `Ipset.add(netListIpsetName, childSet)` is safely idempotent in repeated `ensureCreateEnforcementEnv()` runs.
  - If not, guard or normalize it there.
- Add a defensive check/log around missing `realIntf` before populating the link-local `hash:net,iface` set.
- A few focused validation tests would materially increase confidence here:
  - network with IPv4 + global IPv6 + FE80 link-local subnet
  - rule matching through the aggregate `list:set` in both iptables and ip6tables
  - repeated `ensureCreateEnforcementEnv()` calls do not duplicate/error
  - cleanup flushes aggregate and child behavior as intended

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8972  
Head SHA: `bbcfe6a1a0cd8e08702e56de188a1182a045c306`  
Checked at: 2026-04-24 12:31:31 CST

jasonlyc (approved):

PR #8973, opened 2026-04-24
add a cap to exception

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a reasonable improvement for exception loading.
- The main behavior changes are coherent:
  - exception loading now supports pagination via `{ offset, number }`
  - default unbounded loads are capped by platform capacity
  - results are explicitly sorted newest-first by numeric EID
- Adding `getExceptionCapacity()` to the platform abstraction is also the right shape.
  - It keeps the capacity decision platform-specific instead of hardcoding it in `ExceptionManager`.
- The new tests are useful and cover the intended behaviors well:
  - platform capacity exists and is sane
  - paging works
  - default loads are recent-first
  - default loads honor the platform cap
  - explicit `number` can exceed the default cap

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `loadExceptionsAsync({ number: FILL })` no longer behaves like a true “unbounded load”, but the rest of the codebase may still assume the old semantics.
  - Previously, `loadExceptionsAsync()` returned all exception entries.
  - Now the default call returns only up to `platform.getExceptionCapacity()`.
  - That is a real API behavior change, not just a paging enhancement.
  - Any callers that expect full exception enumeration for cleanup, export, migration, metrics, or dedup checks may now silently see only a truncated subset unless they were updated too.

- **Low:** the `number` defaulting logic treats falsy values specially:
  - `const number = options.number || platform.getExceptionCapacity();`
  - This means `number: 0` cannot be requested and is interpreted as “use capacity”.
  - Same for `offset`, where `0` is fine but other falsy handling is less explicit than it could be.
  - Probably not a practical bug today, but it makes the API less precise.

- **Low:** sorting all IDs in memory after `smembers` can become inefficient if the exception set grows large.
  - Given the current platform caps, this is probably acceptable.
  - Still, the implementation is doing full set fetch + full sort + slice every time, even when only a small page is needed.

### 💡 Suggestions
- I’d re-check all call sites of `loadExceptionsAsync()` to make sure the new capped default is safe everywhere.
  - If some callers still need true full enumeration, consider an explicit opt-in like `{ number: Infinity }` or a separate “load all” path.
- Use nullish/default checks instead of truthy checks for paging options, e.g. treat `undefined` differently from `0`.
- If exception volume is expected to grow further, consider whether this data should eventually move to an ordered structure rather than `SMEMBERS + sort`.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8973  
Head SHA: `5fdd42da0c2d93af3053ae01f7a5c88e88e78ae1`  
Checked at: 2026-04-24 12:45:53 CST

jasonlyc (approved):

PR #8974, opened 2026-04-24
bug fixes

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Consolidating connmark refresh into `BlockControl` is a solid cleanup.
  - It removes duplicated timer/refresh logic from `PolicyManager2` and `DomainUpdater`.
  - Tying connmark refresh to the block-control pipeline reaching idle state is a better synchronization point than ad hoc `setTimeout(...)` calls in multiple modules.
- The `needRefreshConnmark` flag is a reasonable mechanism for coalescing multiple refresh requests while rule updates are still in flight.
- The iptables change is directionally smart:
  - explicitly fast-pathing established non-HTTP/HTTPS TCP/UDP traffic by connmark should reduce unnecessary traversal cost
  - lowering the connbytes threshold from `10` to `7` aligns with that optimization intent
- Raising `FW_PROBABILITY` from `0.99` to `0.999` on larger platforms is a plausible tuning improvement if the goal is to skip firewall traversal more aggressively for already-accepted flows.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `BlockControl.scheduleRefreshConnmark()` now only sets a boolean and depends on `_processQueue()` transitioning to idle to actually run `refreshConnmark()`.
  - That means callers outside the block-control queue lifecycle can request a connmark refresh and never get one if the queue is already idle and no more queued work arrives.
  - In particular, `DomainUpdater` now calls `blockControl.scheduleRefreshConnmark()` after its own processing interval, but it does not necessarily enqueue anything into `BlockControl`.
  - So this can regress from “eventually refresh after 5s” to “refresh never happens unless BlockControl happens to process something later.”

- **Medium:** the new `FW_FORWARD` fast-path exempts all non-80/443 TCP and non-443 UDP packets with the accepted connmark before the later firewall chains run.
  - That is a significant behavioral broadening compared with the old probabilistic bypass.
  - The correctness of this optimization depends on a strong assumption: once the connmark high bit is set, no later policy update needs those packets to be re-evaluated promptly.
  - This PR does add connmark refresh machinery, but given the issue above, that coupling is fragile.

- **Medium:** the TCP/UDP special casing is incomplete from a networking standpoint.
  - TCP only excludes ports `80` and `443`, and UDP only excludes port `443`.
  - If there are important policy-controlled protocols on other ports that still depend on early re-evaluation, they now get fully accepted once marked.
  - This may be intentional, but it is a bigger semantic change than a pure performance tune.

- **Low:** `refreshConnmark()` logs failures at `verbose` instead of `warn`, while the previous code used stronger logging in at least one path.
  - If conntrack updates start failing in production, this may become harder to notice.

### 💡 Suggestions
- I would fix the `BlockControl.scheduleRefreshConnmark()` execution model before merging.
  - If the queue is idle when a refresh is requested, it should probably schedule/run `refreshConnmark()` directly instead of waiting for a future queue transition that may never happen.
- Given the broader fast-path in `iptables_common.sh`, I’d want explicit validation for:
  - policy changes on established non-HTTP/HTTPS flows
  - allow/block transitions after connmark refresh
  - IPv4 and IPv6 behavior
- If the non-80/443 shortcut is intentional, document that assumption clearly in the firewall rules/comments.
- Consider keeping connmark refresh failures at least at `warn` level until the new centralized flow has proven stable.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8974  
Head SHA: `26c6ef1e231fa46b8f719d9890e1f707037a7ea4`  
Checked at: 2026-04-24 13:45:52 CST

drahnieR (approved):

PR #8975, opened 2026-04-24
fix LED race between cloud disconnect timer and reconnect

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a solid fix for an ordering/race problem in LED state handling.
- The main improvement is introducing `ledUpdateChain = Promise.resolve()` and routing both `ledNetworkDown()` and `ledNetworkUp()` through the same promise chain.
  - That makes LED updates serialize instead of racing each other.
  - For hardware-facing side effects like `firestatus`, that is exactly the right direction.
- Moving `this.disconnectCloud = false` to the start of both `reconnect` and `connect` handlers is also correct.
  - It lets any pending `ledNetworkDownJob` bail out before it enqueues a stale down-state transition.
- Clearing the timer and checking `disconnectCloud` before scheduling the down LED update makes the stale-offline case much less likely.
- Error handling on the LED chain is reasonable too; failures are logged and won’t permanently poison the chain.

  - No linked issue found.

### ⚠️ Issues found
- **Low:** `connect` and `reconnect` now both enqueue `platform.ledNetworkUp()` on the same chain, and depending on socket.io event ordering this may lead to duplicate “up” writes for one recovery.
  - That is probably harmless, but it is still extra hardware/state churn.
  - If `firestatus` reacts noisily to repeated identical updates, this could still produce unnecessary work.

- **Low:** the chain only serializes future calls; it does not deduplicate semantically identical target states.
  - So the race is fixed, but flapping connections can still enqueue a sequence like:
    - up
    - up
    - down
    - up
  - again, likely acceptable, just worth noting as follow-up polish rather than a correctness issue.

### 💡 Suggestions
- If you want to make this even tighter, consider tracking the last requested LED state and skipping redundant chained updates when the requested state is already the latest pending/current one.
- A regression test or harness around the sequence:
  - disconnect
  - reconnect before threshold
  - connect/reconnect callbacks both fire
  would be valuable, since this bug is very timing-dependent.
- Otherwise the fix looks good and addresses the real race.

### Verdict
APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8975  
Head SHA: `ef324dc85d4560d8b3770d9b9f9bc139e7048869`  
Checked at: 2026-04-24 18:15:54 CST

jasonlyc (approved):

PR #8976, opened 2026-04-24
add ut for dap api reset-stats

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a useful test follow-up for the new DAP stats-reset API.
- Good parts:
  - the helper split is sensible:
    - generic `postDapEncipherSimple(...)`
    - focused wrappers for `/reset-stats-all` and `/reset-stats/<mac>`
  - the new tests cover the main happy paths:
    - reset one device
    - reset all devices
  - the assertions are meaningful:
    - they verify `learnedCount` actually increases first
    - then verify it returns to `0`
    - and confirm the device remains in `learning` state after reset
- Also nice: fixing the `cleck` typo while touching the file.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** the non-existent-device test is asserting the wrong contract.
  - The API introduced in the corresponding backend change returns `404` for host-not-found / no-policy cases.
  - But this test expects:
    - `statusCode === 500`
    - message matching `DAP API call failed: 404`
  - That bakes an adapter-layer failure shape into the test instead of the intended endpoint behavior.
  - If the proxy/CLI layer gets fixed to preserve `404`, this test will fail even though behavior improved.

- **Low:** the new helper rejects on any non-2xx response, which makes negative-path assertions clumsier and hides the actual response body/status from the normal success shape.
  - For API contract testing, it is often cleaner to always resolve `{ statusCode, body }` and let the test assert on the status explicitly.

- **Low:** there is some unnecessary noise / leftover variables.
  - `body` is captured in a few tests and never asserted.
  - there are several commented-out waits / query-flow lines that reduce readability.

### 💡 Suggestions
- I’d update the negative test to assert the intended API semantics, not the current proxy failure wrapping.
  - Ideally:
    - expect a `404`
    - and assert the returned body/error payload identifies `host_not_found` or equivalent
- Consider changing `postDapEncipherSimple(...)` to always resolve with `{ statusCode, body }`.
  - That makes both success and failure-path tests much cleaner.
- Small cleanup:
  - remove unused `body` bindings
  - drop old commented-out debug lines if they’re no longer needed
- If possible, add one more negative-path test for “device exists but has no DAP policy,” since that is a distinct backend branch from “host not found.”

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8976  
Head SHA: `c4fa8f63a407291f7255b5bc366867f6cbbcf25b`  
Checked at: 2026-04-24 19:00:47 CST

jasonlyc (approved):

PR #8977, opened 2026-04-24
fix race condition that may cause incorrect VPN client connectivity status after stopped

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a sensible race-hardening change in the VPN connectivity checker.
- The early-return split makes the logic clearer:
  - stopped client → explicitly mark cached state as down
  - restarting client → bail out
  - startup grace window → bail out without mutating cached state
- The added post-check `_started` revalidation is the key improvement.
  - It addresses a real async edge case where long-running connectivity tests could complete after the client has already been stopped.
  - Preventing a stale success result from overwriting stopped state is the right behavior.
- The change is small, focused, and low-risk in terms of surface area.

  - No linked issue found.

### ⚠️ Issues found
- **Low:** the new guard only re-checks `_started`, not `_restarting`, after the async connectivity probes complete.
  - There is already a later `_restarting` check, so this is probably fine in practice.
  - But the intent is slightly asymmetric: stop gets a dedicated post-async correction path, while restart only gets a passive return.
  - If restart semantics require clearing/avoiding stale cached state the same way stop does, this may still leave a small edge-case gap.

- **Low:** the startup grace-path now returns without touching cached state unless `_started` is false.
  - That appears intentional, and is likely better than forcing a false negative during warm-up.
  - Still, it slightly changes behavior from the old combined branch, so it’s worth confirming that callers do not rely on cache refresh during the first 70 seconds after start.

### 💡 Suggestions
- Consider making the post-async state guard fully symmetric:
  - re-check both `_started` and `_restarting` immediately after the async probes
  - optionally document why only `_started` needs explicit cache correction
- Add a focused regression test for the exact race this fixes:
  - start connectivity check
  - stop client before async probe completes
  - verify no stale “connected” cached state is written afterward
- If there are existing tests around the startup grace window, update them to reflect the more explicit branch behavior.

### Verdict
APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8977  
Head SHA: `0903b6afc1dd615844acf72937ca3f9b75304273`  
Checked at: 2026-04-24 20:16:28 CST

jasonlyc (approved):

PR #8981, opened 2026-04-26
sync gse 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- There are several worthwhile improvements in this PR:
  - the new bypass-policy model is a cleaner concept than mutating / pausing block rules ad hoc
  - `AppTimeUsageManager` gaining `onQuotaReached: "unenforce"` is a good abstraction for temporary extra-time behavior
  - category regex matching in `PolicyManager2.match()` improves consistency between alarm attribution and dnsmasq enforcement
  - `checkValidDomainRE()` delegating to `CategoryEntry.isValidDomainRE()` is a solid cleanup that removes duplicate validation logic
  - `ExceptionManager.loadExceptionsAsync()` now supports ordered paging instead of returning an unordered set
- I also like the direction of tracking `affectedPids` on bypass policies.
  - That makes cleanup / reverse mapping much more explicit than older shadow-policy style approaches.

  - No linked issue found.

### ⚠️ Issues found
- **High:** `alarm/AccessRequestManager.js` `matchTarget()` is incorrect for single-app requests when the policy uses `targets`.
  - In the single-app branch, the code only checks `policy.target`.
  - If a policy has `targets: [...]` and no `target`, the function falls through to `return true`.
  - That means a single-app request can incorrectly match an unrelated multi-target policy.

- **High:** `matchApp()` dropped the explicit action filter.
  - Previously it only matched:
    - `app_block`
    - `block`
    - `disturb`
  - Now it matches on tag/target/schedule alone, and action filtering is deferred to the caller.
  - That broadens helper semantics in a risky way and makes it easier for unrelated policies to leak into time-limit matching.

- **High:** bypass policy construction for app sets involving `internet` is fragile.
  - The new policy shape depends on `apps[0]` after sorting.
  - For mixed requests, `type` / `target` are derived from the first sorted app, while `targets` contains the whole set.
  - That makes behavior order-sensitive in a way that does not look semantically safe.

- **Medium:** `calculateQuotaLeft()` can return `NaN`.
  - It does:
    - `ruleQuota + ruleExtraQuota - Number(policy.appTimeUsed)`
  - If `policy.appTimeUsed` is missing / undefined / malformed, later comparisons become unreliable.

- **Medium:** quota math in `approveRequest()` is hard to reason about and may over-grant.
  - `totalQuotaLeft = req.leftQuota + minutes`
  - then each matched rule is adjusted against that total.
  - This mixes user-visible remaining quota with per-rule remaining quota, which is tricky when multiple matched rules already have their own extra quota state.

- **Medium:** `AppTimeUsageManager` still assumes `this.enforcedPolicies[pid][uid]` exists in the non-`unenforce` branch.
  - The new code added a guard in one branch, but not consistently everywhere.
  - With more dynamic enforce/unenforce transitions, this becomes a more realistic source of runtime errors.

- **Medium:** `PolicyManager2` introduces `active_bypass_policy` as a second active index.
  - A lot of cleanup paths were updated, which is good.
  - But this kind of duplicated index is easy to let drift over time unless covered by dedicated tests.

- **Medium:** `unenforce()` for bypass policies schedules deletion with a timer.
  - That works in-process, but it is lifecycle-fragile across restarts or timer loss.
  - If this is the intended cleanup path, it needs explicit validation elsewhere.

- **Low:** `AccessRequestManager.js` has an unused import:
  - `const { include } = require('underscore');`
  - Not a correctness bug, but it’s a sign this area needs a cleanup pass.

- **Low:** `AlarmManager2.hasRelatedAppTimeUsage()` now only special-cases `ALARM_VIDEO` and `ALARM_GAME`.
  - That may be intentional, but it is a narrowing of behavior that should be confirmed.

### 💡 Suggestions
- Fix `matchTarget()` first:
  - for single-app requests, if `policy.targets` exists, ensure the requested app is actually included
- Restore explicit action filtering inside `matchApp()`, or rename/split the helper so its broader semantics are obvious and safe
- Rework bypass-policy construction so mixed app sets do not depend on `apps[0]`
- Harden quota math:
  - coerce `policy.appTimeUsed` to `0`
  - add tests for multiple matched policies with existing extra quota
  - add tests for `leftQuota` undefined / zero / stale values
- Add dedicated tests for:
  - single-app request vs multi-target policy
  - mixed `internet` + app bypass creation
  - `affectedPids` cleanup after source policy deletion
  - `active_bypass_policy` index maintenance
  - `onQuotaReached: "unenforce"` transitions
- Remove unused imports before merge

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8981  
Head SHA: `8d6faddbf88fadc8bb556bb6292a1790d1bbe8e5`  
Checked at: 2026-04-26 12:17:31 CST

PR #8983, opened 2026-04-26
sync orange 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- There are several meaningful improvements in this PR:
  - `AccessRequestManager` now models bypass-style approvals much more explicitly instead of only mutating block rules in place.
  - Multi-app handling is better thought through:
    - target parsing is centralized
    - policy apps can be derived from both `target` and `targets`
    - comma-separated app requests are handled intentionally
  - `AppTimeUsageManager` adding `onQuotaReached === "unenforce"` is a useful abstraction.
    - That makes “temporary bypass until quota is consumed” much cleaner than overloading block-only semantics.
  - `ExceptionManager.loadExceptionsAsync()` now pages/sorts exceptions instead of returning an unordered set, which is a practical improvement.
  - `PolicyManager2.checkValidDomainRE()` delegating to `CategoryEntry.isValidDomainRE()` is a good cleanup that reduces duplicated regex validation logic.
  - Matching category alarms against compiled regex domains is also a good correctness improvement.

  - No linked issue found.

### ⚠️ Issues found
- **High:** `alarm/AccessRequestManager.js` imports `include` from `underscore` but never uses it.
  - That is minor by itself, but it is also a smell that this code path was edited without a clean pass.
  - In a PR this large, unused imports make me less confident the behavior was fully exercised.

- **High:** `matchApp()` no longer validates policy action type.
  - Previously it rejected anything except:
    - `app_block`
    - `block`
    - `disturb`
  - That guard is now gone, and `matchApp()` only checks tags, targets, and schedule.
  - `findMatchingTimeLimitRules()` later splits policies by action, but the helper semantics have changed in a way that can leak non-time-limit policies into matching logic unexpectedly.

- **High:** `matchTarget()` multi-app matching is too loose for single-target policies.
  - In the `apps.length === 1` branch, if `policy.targets` exists and `policy.target` is absent, the function falls through to `return true`.
  - That means a policy with `targets: ['TLX-fw-youtube']` can match an unrelated single-app request like `netflix`, as long as the earlier supported-app checks pass.
  - This is a real correctness bug.

- **High:** bypass policy creation for mixed app sets including `internet` looks inconsistent.
  - The new bypass policy sets:
    - `type = "mac"` and `target = "TAG"` when `apps[0] === 'internet'`
    - but still writes `targets: apps.map(...)`
  - That means a multi-app request such as `internet,youtube` depends on `apps[0]` ordering for policy type/target shape.
  - Since the apps are sorted, behavior can be driven by lexicographic order rather than intended semantics.

- **Medium:** `approveRequest()` computes `totalQuotaLeft` from `req.leftQuota + minutes`, then compares that against per-rule quota left.
  - Conceptually this mixes:
    - remaining user-visible quota
    - newly approved increment
    - per-policy remaining allowance
  - The arithmetic may be right for some flows, but it is hard to reason about and easy to over-grant when multiple matched rules already have heterogeneous extra quota.

- **Medium:** `calculateQuotaLeft()` subtracts `policy.appTimeUsed` directly from `quota + extraQuota`.
  - If `appTimeUsed` is missing/undefined, this becomes `NaN`, which then poisons the later comparisons.
  - This helper should defensively coerce usage to `0`.

- **Medium:** `AppTimeUsageManager` uses `this.enforcedPolicies[pid][uid]` without guarding `this.enforcedPolicies[pid]` in the non-`unenforce` branch.
  - In the old code this may have been effectively safe due to call ordering, but with the new `unenforce` flow and more policy state transitions, this is more fragile.
  - A missing pid bucket here would throw.

- **Medium:** `PolicyManager2` introduces `active_bypass_policy` but cleanup/invalidation paths are only partially obvious.
  - The PR updates many deletion paths, which is good.
  - But adding a second active index increases the chance of drift if any future path forgets to maintain both sorted sets.
  - At minimum this needs dedicated tests.

- **Medium:** `unenforce()` for bypass policies schedules deletion using `setTimeout`, but only when `policy.autoDeleteWhenExpires == "1"` and `timeLeft > 0`.
  - If process restarts or timers are lost, cleanup relies on other paths.
  - That may be acceptable, but it is another lifecycle edge that deserves explicit test coverage.

- **Low:** `AlarmManager2.hasRelatedAppTimeUsage()` now only checks `ALARM_VIDEO` and `ALARM_GAME`.
  - That may be intentional, but it narrows prior behavior substantially.
  - If other app-time-based alarm types exist now or later, this becomes a hidden allowlist.

### 💡 Suggestions
- Restore explicit action filtering inside `matchApp()`, or rename the helper to make its broader semantics clear.
- Fix `matchTarget()` single-app behavior so policies with `targets` must actually contain the requested app.
- Revisit bypass-policy construction for requests that include `internet` plus specific apps.
  - The resulting policy shape should not depend on sorted app order.
- Harden quota math:
  - coerce `policy.appTimeUsed` to `0`
  - add tests for multiple matched rules with existing extra quotas
  - add tests for `leftQuota = 0/undefined`
- Add dedicated tests for:
  - single-app request vs multi-target policy mismatch
  - multi-app bypass creation
  - bypass policy cleanup from `affectedPids`
  - `active_bypass_policy` index maintenance
  - `onQuotaReached: "unenforce"` transitions
- Clean up unused imports in this file set before merging.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8983  
Head SHA: `8d6faddbf88fadc8bb556bb6292a1790d1bbe8e5`  
Checked at: 2026-04-26 12:16:10 CST

PR #8986, opened 2026-04-27
fix bugs, correct time limit extra time calculation

caixingcheng:
1. ignore expired extra time
2. correct time_limits folder naem

### DIFF
--- a/alarm/AccessRequestManager.js
+++ b/alarm/AccessRequestManager.js
@@ -478,7 +478,11 @@ class AccessRequestManager {
           continue; // do nothing if the approved quota is already covered by the existing quota and extraQuota on the rule
         }
 
-        au.extraQuota = Number(au.extraQuota) || 0;
+        if (au.extraQuota && au.extraQuotaUntilTs && nowTs < au.extraQuotaUntilTs) {
+          au.extraQuota = Number(au.extraQuota) || 0;
+        } else {
+          au.extraQuota = 0;
+        }
         au.extraQuota += totalQuotaLeft - quotaLeft;
 
         au.extraQuotaUntilTs = endOfTodayTs;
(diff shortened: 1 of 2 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Both changes are narrowly scoped and easy to reason about.
- The `AccessRequestManager` fix is directionally correct:
  - it avoids carrying forward stale `extraQuota` from an already-expired window
  - that matches the surrounding logic better than blindly accumulating whatever value is currently stored
- The `assets.lst` packaging change also makes sense:
  - extracting into a dedicated `time_limits/` directory is cleaner
  - `--strip-components=1` is a good way to avoid archive-root nesting issues
  - clearing existing extracted contents before unpack helps avoid stale frontend assets lingering across deploys

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** the `rm -rf time_limits/*` cleanup is incomplete for hidden files.
  - Shell glob `*` does not remove dotfiles / dot-directories.
  - If a previous archive version left hidden assets behind, they will survive the cleanup and can still cause stale-content issues.
  - Since this change is specifically about making extraction deterministic, leaving hidden files behind weakens that guarantee.

- **Medium:** the new extraction command is not failure-atomic.
  - It clears the target directory before verifying that `tar -xzf` succeeds.
  - If the archive is missing, corrupted, or partially unreadable, the deploy ends with an empty or partially updated `time_limits/` tree.
  - For static UI assets, that can translate directly into broken runtime behavior after deploy.

- **Low:** the `AccessRequestManager` condition relies on truthiness for numeric fields:
  - `if (au.extraQuota && au.extraQuotaUntilTs && nowTs < au.extraQuotaUntilTs)`
  - If `extraQuota` is `"0"` / `0`, this intentionally falls into the reset branch, which is probably fine here.
  - But using truthiness rather than explicit null/expiry checks makes the logic slightly harder to audit and easier to misread later.

### 💡 Suggestions
- In `assets.lst`, make cleanup cover hidden files too, or replace the directory atomically.
  - Safer patterns would be:
    - extract into a temp dir and swap it into place, or
    - remove/recreate the whole target directory instead of only `time_limits/*`
- Consider making extraction failure-atomic:
  - unpack to `time_limits.tmp`
  - verify success
  - then rename/swap into `time_limits`
- In `AccessRequestManager`, consider rewriting the condition more explicitly, e.g. “only preserve `extraQuota` when `extraQuotaUntilTs` exists and is still in the future”.
  - That would make the intended semantics clearer than relying on JS truthiness.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8986  
Head SHA: `003f9ac9d03411d5189bcbb63175364da04fd33b`  
Checked at: 2026-04-27 10:15:57 CST

MelvinTo (approved):

PR #8987, opened 2026-04-27
fix scheduled disturb rule failing to register due to per-target cron collision

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This looks like a targeted fix for recurring `disturb` policies, and the direction makes sense.
- Prioritizing `cronTime` before `needDisturbRegister(policy)` is the right core change if recurring disturb policies were being incorrectly routed into `PolicyDisturbManager` instead of the scheduler.
- The same precedence fix is applied symmetrically in both enforce and unenforce paths, which is good.
- The `PolicyDisturbManager` change also looks consistent with the intent:
  - using a single `target`
  - avoiding `targets: [target]`
  likely prevents downstream code from treating the generated effective disturb policy as a multi-target rule when it really isn’t.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `targets: undefined` changes object shape rather than removing the field.
  - If downstream logic distinguishes between:
    - missing `targets`
    - empty `targets`
    - explicitly `undefined`
    this may still behave differently than intended.
  - In this codebase, some policy serialization paths normalize fields, but not all consumers are guaranteed to.

- **Medium:** moving `cronTime` ahead of disturb registration fixes precedence, but it also changes behavior for any disturb policy that intentionally relied on `PolicyDisturbManager` setup before scheduling.
  - That is probably the correct fix.
  - But it means recurring disturb policies now bypass disturb-manager registration entirely and depend on scheduler-managed lifecycle only.
  - If `PolicyDisturbManager` contains disturb-specific setup not replicated when the scheduled run fires, this should be double-checked.

- **Low:** the behavioral reason for `targets: undefined` is not documented.
  - This is one of those changes that will be hard to understand later without context.

### 💡 Suggestions
- If the goal is to fully remove `targets`, consider deleting the field rather than setting it to `undefined`, unless the current object/serialization model explicitly prefers `undefined`.
- Add or update tests for:
  - recurring disturb policy registers with scheduler, not `PolicyDisturbManager`
  - recurring disturb policy deregisters symmetrically
  - single-target effective disturb policy does not get treated as multi-target
- A short comment near the new `cronTime` precedence would help:
  - e.g. recurring policies must be scheduler-managed even when they are disturb policies

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8987  
Head SHA: `d3e833cf09f4c659c405e58c0195b6a811e0632c`  
Checked at: 2026-04-27 19:01:51 CST

jasonlyc (approved):

PR #8988, opened 2026-04-27
fix remove of tls rule with port set

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This looks like a targeted correctness fix rather than broad churn.
- The `PolicyManager2` reorder makes sense:
  - `commonOptions.remotePortSet` / `remotePortNegate` are now prepared before `__applyRules(...)`
  - that fixes an obvious ordering hazard where unenforce rules could be generated without the remote-port constraint being attached
- The `Block.js` reorder is also reasonable.
  - Building the iptables rule in a more stable order helps when match modules depend on protocol or when rule generation is sensitive to argument sequence.
  - In particular, moving `rule.pro(proto)` earlier is a sensible cleanup.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `PolicyManager2` still sets `remotePortSet = \`c_bp_${pid}_remote_port\`` without creating/populating the ipset in this branch.
  - The old code created the set and inserted `443`:
    - `await ipset.create(remotePortSet, "bitmap:port");`
    - `await Block.batchBlock(["443"], remotePortSet);`
  - This PR removes that work entirely.
  - If `remotePortSet` is not guaranteed to already exist from the enforce path, unenforce can now reference a nonexistent set and fail or partially clean up.

- **Medium:** the change assumes unenforce should reuse previously-created state rather than reconstruct it.
  - That may be correct, but only if lifecycle guarantees are solid.
  - If a restart, partial apply, or earlier cleanup removed the ipset while rules still need to be removed, this becomes fragile.

- **Low:** the `Block.js` argument reordering changes rule emission shape.
  - That is probably harmless or beneficial, but iptables/ipset matching order can be surprisingly touchy in edge cases.
  - This deserves at least a sanity check on generated commands for representative tuple combinations.

### 💡 Suggestions
- Confirm whether `remotePortSet` is always guaranteed to exist during unenforce.
  - If not, this PR may need to recreate it defensively or handle missing-set cleanup more explicitly.
- Add or update tests for:
  - domain/dns block rule with TLS handling and `dnsmasq_only = false`
  - enforce → unenforce round trip where remote port negation is required
  - restart/recovery scenario where unenforce runs after partial state loss
- A short comment in `PolicyManager2` explaining why the set creation was removed would help future readers.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8988  
Head SHA: `72bdd04798374e8b83a9d596d2b031fcb9a98a29`  
Checked at: 2026-04-27 19:16:43 CST

jasonlyc (approved):

PR #8989, opened 2026-04-28
fix close queue if not ready

xinnige:
https://github.com/firewalla/firecommit/issues/4812

### DIFF
--- a/event/EventRequestHandler.js
+++ b/event/EventRequestHandler.js
@@ -111,6 +111,15 @@ class EventRequestHandler {
             const queue = this.queueMap.get(requestKey);
             if (queue.getState() !== 'ready') {
                 log.info(`event queue ${requestKey} is not ready, state ${queue.state}, recreate the queue`);
+                if (queue.queue) {
+                    try {
+                        await queue.queue.close(3000);
+                    } catch (err) {
+                        log.error(`event queue ${requestKey} failed to close before re-setup, ${err.message}`);
+                    } finally {
+                        queue.queue = null;
+                    }
+                }
                 await queue.setupEventQueue(1, this.processStateEvent.bind(this));
             }
         });


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a focused defensive fix for a real lifecycle problem.
- Closing the old queue before recreating it is the right direction.
  - It reduces the chance of leaking stale consumers/resources when a queue leaves `ready` state.
  - Setting `queue.queue = null` in `finally` is also sensible so the object does not keep pointing at a half-closed instance.
- The error handling is pragmatic:
  - close failure is logged
  - re-setup still proceeds

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** this only closes `queue.queue`, but does not guard against concurrent re-entry into the same recovery path.
  - If multiple callers hit this block for the same `requestKey` while the queue is unhealthy, they can all observe the non-ready state, all try to close/recreate, and race through `setupEventQueue(...)`.
  - Depending on how `setupEventQueue(...)` behaves internally, that could still leave duplicate setup work or a last-writer-wins queue reference.

- **Low:** the log message uses `queue.state`, while the readiness check uses `queue.getState()`.
  - If `state` is not the canonical source, the log can report misleading information during debugging.
  - Since this change is specifically about recovery/diagnostics, it would be better to log the exact state value returned by `getState()`.

### 💡 Suggestions
- If this path can be entered concurrently, consider adding a per-queue recovery lock / in-flight promise so only one close-and-setup sequence runs per `requestKey`.
- Cache the state returned by `queue.getState()` into a local variable and use that both for the condition and the log message.
- If supported by the queue implementation, consider logging whether the close timed out versus threw another error; that would make future queue-health debugging easier.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8989  
Head SHA: `d546d8f0879c2276783f4a468244fd6dd68d5931`  
Checked at: 2026-04-28 11:00:58 CST

jasonlyc (approved):

PR #8990, opened 2026-04-28
optimize hit count accounting&add last hit flow info

melo-crypto:
optimize hit count accounting&add last hit flow info

https://github.com/firewalla/firecommit/issues/7648
https://github.com/firewalla/firecommit/issues/7524


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This is a focused feature addition around rule-hit observability.
- Persisting `lastHitFlow` alongside hit counters is useful operationally; it should make rule troubleshooting much easier than just `hitCount` / `lastHitTs`.
- Moving normal allow/disturb/route hit accounting into `BroDetect` is directionally sound.
  - Counting at flow generation time is a better fit for long-lived connections than relying only on audit events.
- The comments around where accounting should happen are helpful and show good awareness of the split between:
  - blocked flows with no conn.log
  - DNS-only enforcement
  - normal IP flows
  - Zeek-skipped DNS-port disturb cases
- The `return -> continue` fix in `RuleStatsPlugin.writeRuleStats()` is definitely correct; one missing policy should not abort syncing the whole batch.

  - No linked issue found.

### ⚠️ Issues found
- **High:** `recordLastHitFlow()` in `RuleStatsPlugin` can regress timestamp ordering because it uses `const ts = flow.ts || flow._ts || 0`.
  - If `flow.ts` is a string, lexicographic comparison in `ts > existing.ts` can behave incorrectly.
  - Elsewhere this codebase frequently normalizes timestamps with `Number(...)`; this path should too, since it controls which flow becomes the persisted “last” hit.

- **Medium:** `lastHitFlow` is only flushed every 60 seconds, but `PolicyManager2.resetStats()` now deletes it immediately and emits `Policy:StatsReset`.
  - The event clears in-memory buffers for the current process, which is good, but if multiple workers/processes are accounting rules, only the local plugin instance receiving the event is protected.
  - Another worker could still write a stale `lastHitFlow` after reset if it buffered it before receiving/processing the reset event or if it does not share the same event bus timing.
  - The new design is better than before, but still not fully robust cross-process.

- **Medium:** `BroDetect` picks only the first matched allow PID (`matchedPIDs[0]`) when backfilling `apid` by host.
  - If multiple allow rules can legitimately match, the chosen PID may be nondeterministic unless `getMatchedPids()` guarantees ordering by precedence.
  - Since this feeds user-visible rule stats, incorrect “winner” selection can make hit attribution misleading.

- **Low:** `lastHitFlow` stores only a reduced subset of flow fields.
  - That is reasonable for storage size, but the new persistence contract is now part of policy data.
  - Without a comment/documented schema near the policy reader side, this field can become loosely defined and harder to evolve safely.

### 💡 Suggestions
- Normalize `ts` in `recordLastHitFlow()` explicitly, e.g. `const ts = Number(flow.ts ?? flow._ts ?? 0)`.
- If `getMatchedPids()` has a precedence guarantee, document that assumption near `matchedPIDs[0]`; otherwise consider resolving the winner more explicitly.
- Consider whether `lastHitFlow` should be flushed:
  - on plugin shutdown/disable
  - on reset before clearing
  - or more frequently than every 60s if the freshness matters for UX/debugging
- Add a small regression test for:
  - stats reset clearing `lastHitFlow`
  - multiple policy entries in `writeRuleStats()` when one policy key is missing
  - long-connection fragments not double-counting hits

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8990  
Head SHA: `ac06fca6ce00f0752fd6516dac50f46a733526e7`  
Checked at: 2026-04-28 11:15:55 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision addresses the biggest correctness gap from the earlier version: stats reset is now timestamped and propagated into `RuleStatsPlugin`, instead of only clearing local buffers blindly.
- The new reset model is directionally strong:
  - `PolicyManager2.resetStats()` writes `statsResetTs`
  - emits `Policy:StatsReset` with `policyIDs` and `resetTime`
  - `RuleStatsPlugin` tracks both global and per-policy reset timestamps
  - buffered rule hits and pending `lastHitFlow` writes are filtered against reset time
- The timestamp normalization in `recordLastHitFlow()` is also a good fix.
  - Using `Number(...)` avoids the earlier risk of string-based timestamp ordering bugs.
- Changing `return` to `continue` when a policy key is missing in `writeRuleStats()` is still the right fix.
- Overall, this version is much more careful about avoiding stale stat resurrection after reset.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `recordLastHitFlow()` still stores `f.ts` using `Number(flow.ts || flow._ts || 0)`, while the comparison key is `const ts = Number(flow._ts || flow.ts || 0)`.
  - That means ordering prefers `_ts` over `ts`, but the persisted payload prefers `ts` over `_ts`.
  - If those differ, the “winning” flow can be selected based on one timestamp and persisted with another, which is subtle but inconsistent.

- **Medium:** `applyStatsReset(policyIDs, resetTime)` clears `lastHitFlowMap` entries for specific PIDs, but it does not clear `policyStatsResetTs` entries when a later global reset occurs except via full `clear()`.
  - That is functionally okay now, but this reset-state cache can grow over time if there are many per-policy resets without a global reset.
  - It’s more of a maintainability / long-lived process hygiene concern than an immediate bug.

- **Low:** `recordBuffer` now stores `_statsTs: Date.now() / 1000`, while flow/accounting hit timestamps still use `_ts || ts`.
  - That distinction is intentional for reset filtering, but it’s subtle.
  - Without a short comment where `_statsTs` is introduced, it’s easy for a future maintainer to “simplify” it away and reintroduce stale-hit-after-reset bugs.

### 💡 Suggestions
- Make persisted `f.ts` use the same normalized source as the ordering key, e.g. reuse the computed `ts`.
- Consider pruning `policyStatsResetTs` opportunistically for policies that no longer exist or after successful flush cycles, just to avoid unbounded map growth in very long-lived processes.
- Add a short comment near `_statsTs` in `accountRule()` explaining that it is the ingestion/accounting timestamp used to suppress stale buffered hits after stats reset.
- If there are tests around this area, the most valuable regression cases would be:
  - reset occurs after hit buffered but before flush → hit not persisted
  - per-policy reset only suppresses that PID, not others
  - `lastHitFlow` does not reappear after reset from delayed flush

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8990  
Head SHA: `184a6c402bef4f39459c8d70d4da05b9c0ce37cd`  
Checked at: 2026-04-28 12:31:25 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision looks much stronger than the initial version.
- The reset flow is now properly timestamped and propagated:
  - `PolicyManager2.resetStats()` writes `statsResetTs`
  - emits `Policy:StatsReset` with `policyIDs` and `resetTime`
  - `RuleStatsPlugin` suppresses buffered hits and pending `lastHitFlow` writes that predate the reset
- Persisting `lastHitFlow` as an object field on `Policy` is useful and the schema update in `Policy.OBJ_VALUE_KEYS` is appropriate.
- The `return` → `continue` fix in `RuleStatsPlugin` is definitely correct; a missing policy key should not abort the whole batch.
- Moving allow/disturb/route hit accounting to `BroDetect` for normal flow-generation cases is directionally sound and the comments explain the split well.

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `BroDetect` still assigns `tmpspec.apid = matchedPIDs[0]` when backfilling allow matches by host.
  - If `getMatchedPids()` can return multiple valid policies without a strict precedence guarantee, stats attribution becomes order-dependent.
  - Since this affects user-visible policy hit stats, it would be better to either document that ordering contract or resolve the winning rule more explicitly.

- **Low:** `policyStatsResetTs` can grow over time in long-lived processes.
  - Per-policy reset timestamps are added to a `Map`, but there is no pruning except when a global reset clears the whole map.
  - That is not an immediate correctness issue, but it is a small maintenance/memory hygiene concern for daemons that run a long time and see many one-off policy resets.

- **Low:** the distinction between `_statsTs` and the actual hit timestamp (`_ts` / `ts`) is subtle.
  - It is the right idea for reset safety, but it is easy for a future refactor to miss why both exist.
  - Without a very explicit comment, this is the kind of thing that tends to get “simplified” later and break stale-hit suppression.

### 💡 Suggestions
- If `getMatchedPids()` is already ordered by rule precedence, add a short comment near `matchedPIDs[0]` saying that explicitly.
- Consider opportunistic pruning of `policyStatsResetTs` for deleted/nonexistent policies if this process is expected to be very long-lived.
- Add a short comment where `_statsTs` is attached in `accountRule()` explaining that it is the accounting-time timestamp used to suppress buffered writes after `Policy:StatsReset`.
- If there are tests around this area, the highest-value cases are:
  - buffered hit before reset does not get written after reset
  - per-policy reset suppresses only that PID
  - `lastHitFlow` does not reappear after reset from delayed flush
  - multi-match allow flow chooses the intended PID deterministically

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8990  
Head SHA: `36d90f1eb7cf1b808be4486a2512d0dd4f4ec1ed`  
Checked at: 2026-04-28 12:45:54 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The overall direction is still good:
  - persisting `lastHitFlow` on policy objects is useful for debugging and UI visibility
  - timestamped stats reset via `statsResetTs` + `Policy:StatsReset` is a solid improvement
  - `RuleStatsPlugin` now guards buffered writes against resets, which is important for correctness
  - changing the missing-policy path from `return` to `continue` is definitely the right fix
- Restricting ACL-audit-side accounting to block cases is also directionally cleaner if BroDetect is intended to be the primary source for emitted-flow hit accounting.

  - No linked issue found.

### ⚠️ Issues found
- **High:** this revision appears to reintroduce double-counting risk for long connections.
  - In the earlier safer version, BroDetect captured whether the connection had already been tracked and used that to avoid accounting repeated long-conn fragments.
  - In this diff, the rule-hit accounting block now runs for all emitted flows except `reverseLocal`:
    - `if (!reverseLocal && ruleStatsPlugin) { ... }`
  - That means repeated `conn_long.log` emissions for the same long-lived flow can increment allow/disturb/route hit counts multiple times.
  - Given the comment change from “skip duplicates on long conn fragments” to “with emitted flows and only skip the reverse local pass,” this looks intentional, but it regresses hit-count correctness.

- **Medium:** `tmpspec.apid = matchedPIDs[0]` is still order-dependent unless `getMatchedPids()` has a strict precedence guarantee.
  - If multiple allow rules can match by host, stats attribution may be misleading.
  - Since this becomes user-visible policy-hit data, that assumption should be explicit.

- **Low:** `policyStatsResetTs` can still grow over time in long-lived processes with many per-policy resets.
  - Not an immediate bug, but worth keeping in mind.

- **Low:** the `_statsTs` vs hit timestamp (`_ts` / `ts`) distinction is still subtle.
  - The implementation is reasonable, but this is the kind of logic that future cleanup can accidentally break.

### 💡 Suggestions
- Restore the earlier long-connection dedupe behavior in BroDetect.
  - At minimum, avoid counting repeated fragments for UIDs already tracked as long connections.
  - If you intentionally want one hit per emitted flow instead of one hit per connection, that should be called out explicitly because it changes the semantics of policy stats.
- If `getMatchedPids()` returns policies in priority order, document that near `matchedPIDs[0]`.
- Add or update regression tests for:
  - long-lived connections generating multiple Zeek emissions do not inflate hit counts
  - stats reset suppresses buffered stale writes
  - block-only ACL audit accounting still covers cases BroDetect cannot emit

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8990  
Head SHA: `26b92fe167e13c9143add9d4cf824ae16684da4e`  
Checked at: 2026-04-28 13:46:24 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is a meaningful improvement over the earlier version.
- The main architecture now hangs together much better:
  - `lastHitFlow` is persisted as structured object data on the policy
  - app-facing policy reads go through formatting/enrichment paths instead of exposing raw internal payloads
  - stats reset now clears `hitCount`, `lastHitTs`, and `lastHitFlow` together
  - `Policy:StatsReset` plus `_statsTs` / `statsResetTs` is a solid way to suppress stale buffered writes after reset
- I also like these specific fixes:
  - changing missing-policy handling from `return` to `continue`
  - formatting `lastHitFlow` for both single-policy and list APIs
  - distinguishing `flow` vs `audit` sources so the app can reuse existing enrichment logic

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `tmpspec.apid = matchedPIDs[0]` is still order-dependent.
  - If multiple allow rules can match the same host/flow, policy attribution depends on whatever order `getMatchedPids()` happens to return.
  - Since this becomes user-visible “last hit” / stats data, that precedence should be deterministic and documented.

- **Low:** `policyStatsResetTs` can still grow without any pruning strategy.
  - In a long-lived process with lots of targeted resets, this map can accumulate entries for policies that may never be touched again.
  - It is not an immediate correctness bug, but it is worth addressing before it becomes quiet process-state debt.

- **Low:** the `_statsTs` vs actual hit timestamp split is correct but subtle.
  - The code now relies on:
    - `ts` / `_ts` for “when the traffic happened”
    - `_statsTs` for “when this stat record entered local buffering”
  - That distinction is important for reset correctness, but it is non-obvious and easy for future cleanup to break.

### 💡 Suggestions
- Define or document precedence for `matchedPIDs[0]`.
  - Ideally `getMatchedPids()` should guarantee stable order by rule priority/specificity.
- Consider pruning `policyStatsResetTs` entries:
  - on policy removal
  - or opportunistically after successful writes / periodic maintenance
- Add a short code comment where `_statsTs` is introduced and where it is compared against reset time, explaining why it must differ from flow hit timestamp.
- Optional follow-up test coverage would be valuable for:
  - deterministic allow-rule attribution when multiple rules match
  - targeted reset followed by delayed buffered writes
  - app formatting of `lastHitFlow` for both `flow` and `audit` sources

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8990  
Head SHA: `bea2253946d4c06e5d7987a3fca182f11983f913`  
Checked at: 2026-04-28 18:00:56 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- This revision is in much better shape overall.
- The core design now feels consistent:
  - `lastHitFlow` is persisted as structured object data
  - policy reads for app-facing surfaces go through a dedicated formatting path
  - reset semantics now cover `hitCount`, `lastHitTs`, and `lastHitFlow` together
  - buffered stat writes are guarded by `_statsTs` vs `statsResetTs`, which is the right direction
- A few specific improvements are especially good:
  - `getPolicyForApp(...)` / `loadActivePoliciesForApp(...)` avoid leaking raw internal snapshots to the app
  - `LogQuery.enrichSimpleLog(...)` is a useful small abstraction
  - changing the missing-policy case from `return` to `continue` in `RuleStatsPlugin` fixes a real bug
  - moving allow/route/disturb conntrack recording earlier in `ACLAuditLogPlugin` makes the flow-generation path cleaner

  - No linked issue found.

### ⚠️ Issues found
- **Medium:** `tmpspec.apid = matchedPIDs[0]` is still order-dependent.
  - If multiple allow rules match the same host/flow, policy attribution depends on whatever ordering `getMatchedPids()` returns.
  - Since this data is now exposed through app-facing policy stats / last-hit detail, that precedence should really be deterministic and documented.

- **Low:** `policyStatsResetTs` can still grow indefinitely in a long-lived process.
  - Every targeted reset inserts/updates entries, but there is no pruning path for removed or long-unused policy IDs.
  - Not a blocker, but worth watching.

- **Low:** the `_statsTs` concept is correct but subtle.
  - The code now distinguishes:
    - traffic time (`ts` / `_ts`)
    - local buffering/write eligibility time (`_statsTs`)
  - That’s important for post-reset correctness, but it is easy for a future refactor to accidentally collapse them.

### 💡 Suggestions
- Define deterministic precedence for `matchedPIDs[0]`.
  - Ideally `getMatchedPids()` should guarantee stable ordering by policy priority/specificity, or the caller should explicitly choose.
- Add a short explanatory comment where `_statsTs` is introduced and where it is compared against reset timestamps.
  - That would make the reset logic much easier to preserve safely.
- Consider a future cleanup/prune strategy for `policyStatsResetTs`, especially if policy IDs churn frequently.
- Useful follow-up tests would be:
  - multiple matching allow rules produce stable attribution
  - targeted reset suppresses delayed buffered writes
  - `lastHitFlow` formatting works for both `flow` and `audit` snapshots

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8990  
Head SHA: `18516d1707833bbec86d081fe77a30da21fd00d0`  
Checked at: 2026-04-28 19:16:21 CST

jasonlyc (approved):

PR #8991, opened 2026-04-28
alway flush _dm set when current mode is domainonly

caixingcheng:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The functional change here is small but sensible:
  - when switching into `domainOnly`, flushing the `_dm` ipsets is the right cleanup step
  - the comment updates also better reflect what the code is actually doing
- Expanding the flush condition from:
  - `lastRecyclemode === "default" && currentRecyclemode === "domainOnly"`
  to:
  - `currentRecyclemode === "domainOnly"`
  is defensible if the invariant is “domain-only mode should never retain `_dm` IP members regardless of how we got there.”

  - No linked issue found.

### ⚠️ Issues found
- **Low:** the broader flush condition changes behavior for all entries into `domainOnly`, not just `default -> domainOnly` transitions.
  - That may be correct, but it is a semantic broadening, not just a TODO cleanup.
  - If any caller relied on preserving `_dm` contents during no-op refreshes while already in `domainOnly`, this will now clear them every time.

- **Low:** `Ipset.flush(...)` is still fire-and-forget here.
  - If flush can fail or race with concurrent updates, the code does not surface that explicitly.
  - This is not introduced by the PR, but the new behavior makes the flush path more important.

### 💡 Suggestions
- Add a short comment explaining *why* `_dm` should always be flushed whenever `currentRecyclemode === "domainOnly"`, not only on the transition from `default`.
- If this path is important for correctness, a targeted test around recycle-mode switching would help:
  - `default -> domainOnly` clears `_dm`
  - staying in `domainOnly` after refresh also leaves `_dm` empty
  - `domainOnly -> default` re-registers updater behavior as expected

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8991  
Head SHA: `2f676ca40efdef213ff8691ac018634cd1529495`  
Checked at: 2026-04-28 18:30:53 CST

jasonlyc (approved):

PR #8997, opened 2026-04-30
sync pse 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Good broad feature direction: extra time requests, bypass-policy handling, app-facing policy formatting, and `lastHitFlow` support all address useful product needs.
- `RuleStatsPlugin` reset handling is more robust than before: buffered stats and pending last-hit flows are cleared/filtered on reset.
- Moving some iptables operations from shell execution to `IptablesControl` is a maintainability improvement.
- Event queue cleanup before re-setup is a good defensive fix.

### ⚠️ Issues found

- **No linked issue found.**

- **High — block and route audit records are dropped before being written.**  
  `[sensor/ACLAuditLogPlugin.js:510]`  
  `added` is only assigned for `allow` and `disturb`. For `block` and `route`, it remains `undefined`, so `if (!added) return` exits before hostname enrichment, `writeBuffer(record)`, reverse-record handling, Redis publish, and block rule hit accounting. This likely breaks block audit logs and route audit behavior.

- **High — inactive host cutoff calculation is wrong when config is missing.**  
  `[net2/HostManager.js:1928]`  
  `Date.now()/1000 - fc.getConfig().timing['host.active'] || INACTIVE_TIME_SPAN` falls back to `INACTIVE_TIME_SPAN` itself, not “now minus default span”. If the config value is missing/invalid, the cutoff becomes a timestamp near 1970, so stale hosts may never be filtered correctly. Use `Date.now()/1000 - (configured || INACTIVE_TIME_SPAN)`.

- **High — remote port ipset is referenced without being created/populated.**  
  `[alarm/PolicyManager2.js:2735]`  
  `remotePortSet` and `remotePortNegate` are still set, but the previous `ipset.create(...)` and `Block.batchBlock(["443"], remotePortSet)` were removed. If the set does not already exist, rule application can fail; if it is stale/empty, matching behavior changes.

- **Medium — policy-list/init paths can now do expensive per-policy enrichment.**  
  `[controllers/netbot.js:1517]` / `[net2/HostManager.js:1668]`  
  `loadActivePoliciesForApp()` and `formatPoliciesForApp()` enrich `lastHitFlow` for every returned policy. On large policy lists this can add latency to common app init/list APIs. Consider making enrichment opt-in or bounded.

- **Medium — `getAppFromTarget()` can throw on invalid target input.**  
  `[alarm/AccessRequestManager.js]`  
  `getAppFromTarget(target)` calls `target.startsWith(...)` without guarding `target`. Several policy fields are optional; a malformed or partial policy can crash request matching.

### 💡 Suggestions

- Restrict duplicate suppression in `ACLAuditLogPlugin` to `allow` / `disturb`; do not early-return block/route logs.
- Fix inactive cutoff with explicit parentheses and validation.
- Restore remote port ipset creation/population or remove the `remotePortSet` dependency entirely.
- Add regression tests for:
  - block audit logs still being written/published
  - route audit records not being dropped
  - inactive host filtering with missing `host.active`
  - policy stats reset + `lastHitFlow`
  - domain/TLS block unenforcement remote-port behavior

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/8997  
Head SHA: `b955513fb8a6c691ea489c7cfa83408831703307`  
Checked at: 2026-04-30 12:27:10 CST

PR #9026, opened 2026-05-08
add dap case for exclude non-unicast IP addresses

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds regression coverage for excluding non-unicast IP addresses from DAP learned/block flows.
- Covers both learning and optimizing state transitions.
- Extends the rule-stats test stub with `recordLastHitFlow()`, matching the plugin interface.
- PR metadata was accessible.

### ⚠️ Issues found

- **No linked issue found.** PR title/body/comments do not link a GitHub issue.

- **High — test now hardcodes a machine-specific DAP binary path.**  
  `test/test_dap_analyze_device.js` uncomments `process.env.DAP_BIN = "/data/dapenv/dap";`. This will break CI/developer machines without that exact path and overrides callers who intentionally set `DAP_BIN`. Keep this commented, or set a default only when the file exists and `process.env.DAP_BIN` is unset.

- **Medium — test expectation does not actually validate the non-unicast learning flow exclusion.**  
  After adding pair `2`, the test expects `learnedCount` to increase from `2` to `4`. But pair `2` uses `0.0.0.0`, so if the goal is “exclude non-unicast IP addresses,” the count should not increase due to those flows. As written, this may pass while the exclusion is broken or is being masked by pre-existing valid flows.

- **Low — fixed 2s sleep can make the test flaky/slow.**  
  `await new Promise(resolve => setTimeout(resolve, 2000));` depends on timing. Prefer polling the buffered/audit side effect with a short timeout, or expose/await the flush completion path directly.

### 💡 Suggestions

- Restore environment portability by respecting externally provided `DAP_BIN`.
- Add explicit assertions that no learned/block-flow entry contains `0.0.0.0`, multicast, broadcast, or other non-unicast addresses.
- Run this test in isolation and as part of the full suite to catch shared-state leakage.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9026  
Head SHA: `04a104622c294be14e76173326094150cda670cb`  
Checked at: 2026-05-08 13:20:15 CST

jasonlyc (approved):

PR #9067, opened 2026-05-18
fix bug, bypass chain wasn't removed

caixingcheng:
bypass chain wasn't removed when timelimit disturb rule is deleted

### DIFF
--- a/alarm/PolicyDisturbManager.js
+++ b/alarm/PolicyDisturbManager.js
@@ -245,6 +245,20 @@ class PolicyDisturbManager {
     await DisturbDispatch.teardownDispatchForPolicy(state.policy);
   }
 
+  checkIfNeedDisableQuic(policy) {
+    const targets = this._getPolicyTargets(policy);
+    if (_.isEmpty(targets)) return false;
+
+    for (const target of targets) {
+      const appName = this._getAppNameFromTarget(target) || policy.app_name || "";
+      const appConf = this._appConfValue[appName];
+      if (appConf && appConf.disableQuic) {
+        return true;
+      }
+    }
+    return false;
+  }
+
 }
 
 module.exports = new PolicyDisturbManager();
--- a/alarm/PolicyManager2.js
+++ b/alarm/PolicyManager2.js
@@ -225,6 +225,23 @@ class PolicyManager2 {
             } else {
               await this.unenforce(policy)
             }
+
+            // check if policy is disturb or block policy with bypass reference, 
+            // if yes, check if policy is removed from redis, then remove related bypass chain
+            if ((policy.type == "category" || policy.type == "mac" || policy.type == "internet") && (policy.action == "block" || policy.action == "app_block" || policy.action == "disturb")) {
+              if (!(await this.policyExists(policy.pid))) {
+                const chainName = `FW_${policy.pid}_BYPASS`;
+                const tables = ["filter", "mangle"];
+                for (const table of tables) {
+                  if (!Bypass.isBypassChainExist(table, policy.pid)) {
+                    continue;
+                  }
+                  await Bypass.removeBypassChain(table, policy.pid).catch((err) => {
+                    log.error(`Failed to remove bypass chain for policy ${policy.pid} on table ${table}`, err.message);
+                  });
+                }
+              }
+            }
           } catch (err) {
             log.error("unenforce policy failed:", err, policy)
           } finally {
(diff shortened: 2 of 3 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Moves bypass-chain creation into a shared helper instead of duplicating `-N` logic.
- Attempts to clean both `filter` and `mangle` bypass chains after deleted block/disturb policies are unenforced.
- Keeps cleanup errors contained with logging so policy processing can continue.
- Adds `checkIfNeedDisableQuic()` helper for app targets with `disableQuic` config.
- GitHub API lookup succeeded for PR metadata/comments.
- No linked issue found.

### ⚠️ Issues found

- **High — bypass-chain cleanup now depends on in-memory state, so stale chains can survive restarts.**  
  `isBypassChainExist()` only checks `initializedBypassChain`, not actual iptables/ip6tables state. If FireMain restarts after a bypass chain was created, or if the chain was created before this module initialized, `initializedBypassChain` is empty and `removeBypassChain()` will skip cleanup. This directly affects the bug being fixed: deleted time-limit disturb rules can still leave `FW_<pid>_BYPASS` chains behind. Please check actual iptables/ip6tables chain existence, or make `removeBypassChain()` attempt `-F`/`-X` idempotently and ignore “not found” errors.

- **Medium — replacing the old delete-time cleanup with deferred in-memory cleanup may miss policies not processed through this path.**  
  The previous `deletePolicy()` path always attempted chain removal. Now cleanup only happens inside the unenforce queue and only if `policyExists(policy.pid)` is false. If deletion exits before this queued path runs, or the policy is already gone but in-memory chain tracking is missing, cleanup is skipped. Consider keeping idempotent cleanup in `deletePolicy()` as a safety net.

### 💡 Suggestions

- Rename `ensurceCreateBypassChain` to `ensureCreateBypassChain` before it spreads to more call sites.
- Add a regression test for: create time-limit disturb rule → restart FireMain/module state → delete policy → verify `FW_<pid>_BYPASS` is removed from both IPv4/IPv6 tables.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9067  
Head SHA: `c1669d58c4e19f99ae91a8e48d0d3917ecba4b2d`  
Checked at: 2026-05-18 10:16:31 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Fixes the typo by exposing `ensureCreateBypassChain()`.
- Extends tag/interface bypass setup for disturb policies that need QUIC blocking to create filter-table bypass chains as well as mangle-table chains.
- Centralizes bypass-chain creation/removal helpers.
- Attempts cleanup across both `filter` and `mangle` tables after deleted block/disturb policies are unenforced.
- GitHub API lookup succeeded for PR metadata/comments.
- No linked issue found.

### ⚠️ Issues found

- **High — bypass-chain cleanup still depends on process-local memory.**  
  `isBypassChainExist()` still only checks `initializedBypassChain`, and `removeBypassChain()` still returns early when the key is absent. Any chain created before a FireMain restart, by older code, or by another process will not be removed. This still misses the core stale-chain case. Cleanup should either query actual iptables/ip6tables state or idempotently attempt `-F`/`-X` and ignore “chain not found”.

- **Medium — create helper can skip recreating a chain that was deleted externally.**  
  `ensureCreateBypassChain()` returns early when `initializedBypassChain[key]` is set, without checking actual chain existence. If iptables state is reset/reloaded while the process remains alive, later rule insertion into `FW_<pid>_BYPASS` can fail because the helper thinks the chain exists. Prefer verifying actual chain existence or treating `-N` as idempotent with ignored “already exists” errors.

- **Medium — cleanup was removed from `deletePolicy()` and now only runs in the unenforce queue.**  
  If the queued unenforce path is skipped, fails before cleanup, or runs after in-memory state was lost, stale chains remain. Keeping an idempotent cleanup call in `deletePolicy()` would provide a safer fallback.

### 💡 Suggestions

- Add a regression test for create → process restart/state loss → delete → verify `FW_<pid>_BYPASS` is gone in IPv4/IPv6 `filter` and `mangle`.
- Consider moving the `PolicyDisturbManager` import to module scope if there is no circular dependency concern.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9067  
Head SHA: `27a40a8d5eebdfb113c030fdf65ebac6050b3fe4`  
Checked at: 2026-05-18 10:46:27 CST

jasonlyc (approved):

PR #9079, opened 2026-05-19
Fix radius compatibility

xinnige:
https://github.com/firewalla/firecommit/issues/7778

### DIFF
--- a/extension/freeradius/freeradius.js
+++ b/extension/freeradius/freeradius.js
@@ -805,19 +805,19 @@ class FreeRadius {
   async cleanupConfig(options = {}) {
     // cleanup certificates
     log.info("Cleaning up freeradius certificates...");
-    await exec(`sudo rm -rf ${certsDir}/*`).catch((e) => {
+    await exec(`sudo find ${certsDir} -mindepth 1 -maxdepth 1 -exec rm -rf {} +`).catch((e) => {
       log.warn("Failed to cleanup certificates,", e.message);
     });
 
     // cleanup config
     log.info("Cleaning up freeradius config...");
-    await exec(`sudo rm -rf ${configDir}/*`).catch((e) => {
+    await exec(`sudo find ${configDir} -mindepth 1 -maxdepth 1 -exec rm -rf {} +`).catch((e) => {
       log.warn("Failed to cleanup freeradius config,", e.message);
     });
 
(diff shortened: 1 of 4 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Updates FreeRADIUS container name handling for Docker Compose v2 hyphenated names.
- Replaces `rm -rf dir/*` with `find ... -mindepth 1`, which avoids glob expansion edge cases.
- Applies the expected container name consistently across diagnostics and update scripts.
- Adds cleanup visibility via `xargs -t`.
- Adds a guard before policy revert to avoid operating on non-serializable policy state.
- GitHub API lookup failed with `HTTPError HTTP Error 404: Not Found`; linked issue verification is limited to the provided diff.
- No linked issue found.

### ⚠️ Issues found

- **Medium — disabled-feature cleanup still appears to use an undefined variable.**  
  In `update_freeradius.sh`, the disabled branch loops over `tags`, but removes containers with `ancestor=$img`. I don’t see `img` defined in this script; this likely means stale FreeRADIUS containers are not removed when the feature is disabled. Use `ancestor=$tag` or the intended image variable.

- **Medium — container-name detection may miss existing v1 containers after Docker Compose v2 is installed.**  
  `get_expected_container_name()` chooses the v2 hyphenated name solely from `docker compose version`. If a box already has the old underscore-named container from earlier runs, diagnostics/update checks may not see it and may leave it running until cleanup catches it. Consider checking both names or resolving the actual existing container first.

- **Low — `checkPolicy()` only validates JSON serialization, not policy validity.**  
  `JSON.stringify(policy)` succeeds for most malformed-but-serializable policy objects, so `revertPolicy()` may skip revert even when policy content is invalid. If the intent is to avoid reverting only when the current policy is semantically valid, add stronger validation.

### 💡 Suggestions

- Quote path variables in shell commands defensively, even if current paths are fixed.
- Add a small shell test/manual QA note for Compose v1 container, Compose v2 container, and feature-disabled cleanup paths.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9079  
Head SHA: `f4f28cb21ea7a857c85a30d5d34308fba6383797`  
Checked at: 2026-05-19 12:09:52 CST

PR #9081, opened 2026-05-19
Fix radius compatibility

xinnige:
https://github.com/firewalla/firecommit/issues/7778

### DIFF
--- a/extension/freeradius/freeradius.js
+++ b/extension/freeradius/freeradius.js
@@ -805,19 +805,19 @@ class FreeRadius {
   async cleanupConfig(options = {}) {
     // cleanup certificates
     log.info("Cleaning up freeradius certificates...");
-    await exec(`sudo rm -rf ${certsDir}/*`).catch((e) => {
+    await exec(`sudo find ${certsDir} -mindepth 1 -maxdepth 1 -exec rm -rf {} +`).catch((e) => {
       log.warn("Failed to cleanup certificates,", e.message);
     });
 
     // cleanup config
     log.info("Cleaning up freeradius config...");
-    await exec(`sudo rm -rf ${configDir}/*`).catch((e) => {
+    await exec(`sudo find ${configDir} -mindepth 1 -maxdepth 1 -exec rm -rf {} +`).catch((e) => {
       log.warn("Failed to cleanup freeradius config,", e.message);
     });
 
     // cleanup docker compose folder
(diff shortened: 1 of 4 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Replaces `rm -rf <dir>/*` cleanup with `find ... -mindepth 1 -maxdepth 1`, avoiding glob expansion edge cases.
- Updates FreeRADIUS diagnostics to handle Docker Compose v2 container naming (`freeradius-freeradius-1`) as well as legacy compose naming.
- Adds expected-image validation in `diag_radius.sh`, making policy/compose/image mismatch easier to diagnose.
- Fixes the disabled-feature cleanup loop in `update_freeradius.sh` to use the actual image tag variable.
- Adds compose image update logic when the expected policy image changes.
- GitHub API lookup succeeded for PR metadata/comments.
- No linked issue found.

### ⚠️ Issues found

- **Medium — `revertPolicy()` now skips almost all real revert operations.**  
  `checkPolicy()` only verifies that `JSON.stringify(policy)` succeeds. Most semantically bad or stale FreeRADIUS policies are still JSON-serializable, so `revertPolicy()` will return early and never restore the target/system policy. If the intent is to skip revert only when policy is valid according to the RADIUS schema, this needs stronger validation; otherwise this can leave broken policy in place.

- **Low — shell commands still interpolate directory paths unquoted.**  
  `sudo find ${certsDir} ...` is safer than `rm -rf ${certsDir}/*`, but still relies on paths never containing whitespace/shell metacharacters. These are likely constants today, but quoting would make this robust.

### 💡 Suggestions

- Add a test or explicit scenario for `revertPolicy()` covering corrupted policy, valid policy, and semantically invalid-but-serializable policy.
- Quote shell path variables in cleanup commands: `sudo find '${certsDir}' ...` or use `execFile`/argument arrays if available.
- In `update_freeradius.sh`, consider logging when compose image is updated but no restart is triggered, so operators understand whether the running container changed.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9081  
Head SHA: `4b671aafb780bc4c0594aa187c4e5952f66b2f60`  
Checked at: 2026-05-19 16:16:34 CST

jasonlyc (approved):

PR #9090, opened 2026-05-20
add ntp local clock other than orphan mode

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Simplifies IPv4/IPv6 NTP chain creation with a small loop while preserving both `NTP_CHAIN` and `NTP_CHAIN_DNAT`.
- Keeps `ntpd` available for redirected clients even when upstream NTP peers are unavailable.
- Adds explicit local clock fallback via `server 127.127.1.0` and `fudge ... stratum 10`, which matches the goal of preventing NTP redirect suspension during upstream outages.
- Keeps the config updates idempotent with `grep -q` guards.
- GitHub PR metadata lookup succeeded: title `add ntp local clock other than orphan mode`, body `null`, state `open`; no issue/review comments.
- No linked issue found.

### ⚠️ Issues found

- No blocking issues found.

### 💡 Suggestions

- Consider logging failures from the `/etc/ntp.conf` update/restart block instead of silently swallowing them, so field debugging can distinguish config failure from later NTP health-check failure.
- If `Rule.fam(4)` has not been used elsewhere, verify it produces the same IPv4 behavior as the previous default-family rule creation.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9090  
Head SHA: `e25fc3d4276e1075ccf023a7fbb669a9c4fc4270`  
Checked at: 2026-05-20 11:31:29 CST

jasonlyc (approved):

PR #9107, opened 2026-05-22
fix bug, use single recycleCategoryJob to serve all category

caixingcheng:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Reviewed 1 changed file and generated 1 inline finding.
- The change removes the unbounded per-category `UpdateJob` map, which addresses the likely resource/coordination problem from having one scheduler per category.
- Linked issue coverage: firewalla/firecommit#7950 requires target-list DNS results to populate `dm_set` / `ip_set` reliably on first rule creation so QoS rate limiting applies. This PR is only partially meeting that requirement because queued recycle work may now be reported as complete before it has actually run.

### ⚠️ Issues found
[control/CategoryUpdater.js:167]  
High: `await this.recycleCategoryJob.exec(event.category)` no longer guarantees that this category has been recycled. `UpdateJob.exec()` returns immediately when another run is already active; with the new single shared job, a different category recycle in progress causes this call to enqueue `event.category` and resolve right away. The handler then marks `initializedCategories[event.category] = true` even though the ipset rebuild may still be pending, preserving a race where first-time target-list QoS can observe empty `dm_set` / `ip_set`. Please either make `UpdateJob.exec()` return a promise that resolves after the queued argument is processed, or avoid relying on `await exec(category)` before marking category initialization / downstream readiness.

### 💡 Suggestions
- Add a regression test that schedules two different categories back-to-back and asserts the second caller does not proceed until its `recycleIPSet(category)` has completed.
- Consider preserving per-category completion promises while still using a single worker queue, so callers get correct readiness semantics without reintroducing concurrent ipset rebuilds.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9107  
Head SHA: `54cf06b6b6406691994c44c69ded5d5edbe27862`  
Checked at: 2026-05-22 14:09:23 CST

jasonlyc (approved):

PR #9117, opened 2026-05-22
sync orange 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- `CategoryUpdater` cleanup removes repeated per-call `UpdateJob` construction and centralizes recycle scheduling through one job.
- `UpdateJob.exec(event.category)` preserves the category argument, so different categories are still processed with the intended parameter.
- Serializing `recycleIPSet()` calls is a reasonable direction because the method mutates shared ipset/domain state and performs expensive system updates.
- The WAN bandwidth change makes `single` WAN behavior consistent with failover by selecting the active/default WAN config instead of whichever `wanConfs` entry happens to appear first.
- No linked issue found.

### ⚠️ Issues found

No blocking issues found.

### 💡 Suggestions

- In `HostManager.js`, consider a fallback when `activeWanUUID` is missing or not found in `bandwidth.wanConfs`. With the new logic, single-WAN QoS bandwidth becomes `0/0` in that case, while the old behavior used the first configured WAN entry. If startup ordering can make `getDefaultWanInterface()` temporarily unavailable, this could accidentally clear QoS bandwidth.
- In `CategoryUpdater.js`, one shared `UpdateJob` means category recycle jobs now run strictly serially. That is probably intended, but it would be worth confirming this does not create long delays when many category updates arrive together, since each queued recycle incurs the configured interval.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9117  
Head SHA: `5c8bf29e7c41f9debdeb873a9f40abd5b5869fbd`  
Checked at: 2026-05-23 04:32:19 CST

PR #9118, opened 2026-05-25
fix bug, single job might may cause recycleIpset to be queued

caixingcheng:
A single recycle job may cause the RecycleIpset for multiple categories to be queued.

### DIFF
--- a/control/CategoryUpdater.js
+++ b/control/CategoryUpdater.js
@@ -89,7 +89,7 @@ class CategoryUpdater extends CategoryUpdaterBase {
       this.origBfCategoryMap = {};
       this.loadCategoryBfParts();
       this.flowSignatureConfig = {};
-      this.recycleCategoryJob = new scheduler.UpdateJob(this.recycleIPSet.bind(this), 1000);
+      this.recycleCategoryJobs = new Map();
       // key: category, value: map of sigId to map of hashkey string to sig detected server entry
       // {category: {sigId: {hashkey: sigEntry}, ...}, ...}
       this.effectiveCategorySigDtSrvs = new Map();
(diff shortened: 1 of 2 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Replaces the single global `UpdateJob` with per-category recycle jobs, which matches the PR goal: unrelated categories no longer serialize behind one shared queue.
- Existing recycle call sites now consistently go through `_getRecycleJob(category)`.
- No linked GitHub issue found in PR metadata, so issue requirement coverage is limited to the PR body: “A single recycle job may cause RecycleIpset for multiple categories to be queued.” The implementation addresses that core requirement.

### ⚠️ Issues found

[CategoryUpdateSensor.js:584]  
Deleting the job from `recycleCategoryJobs` does not cancel an already-running or queued `UpdateJob`. If a recycle for the deleted category is already scheduled, it can still execute after the category data has been flushed. Also, because the map entry is removed, a later event can create a second job for the same category while the old one is still active. Please either avoid deleting active jobs, add cancellation/drain support, or explicitly flush the related ipset after category deletion.

### 💡 Suggestions

- Consider encapsulating category deletion cleanup behind a `CategoryUpdater` method instead of directly mutating `categoryUpdater.recycleCategoryJobs` from the sensor. That keeps scheduler lifecycle ownership inside `CategoryUpdater`.
- Add a focused test or regression case for two categories receiving recycle requests close together, confirming category A does not delay category B unnecessarily.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9118  
Head SHA: `f12822a5e90a7c5b1867ca2963c525beb7e81cf1`  
Checked at: 2026-05-25 10:07:27 CST

jasonlyc (approved):

PR #9164, opened 2026-06-02
fix bug, correct blockset name

caixingcheng:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The fix correctly reuses the same `blockSet` and `port` values for both `syncDomainIPMapping()` and `domainUpdater.clearIPCacheForDomain()`.
- This addresses the linked issue `firewalla/firecommit#8084`: non-port category domains previously cleared cache using the domain-port ipset name, so the computed `DomainUpdater` mapping key did not match and stale `ipCache` entries could survive recycle.
- Port-based entries still use `getDomainPortIPSetName(...)`, while normal domain entries now use `getIPSetName(...)`, matching the registration/mapping path.
- The change is small, scoped, and consistent with `DomainIPTool.getDomainIPMappingKey()` behavior, where `blockSet` and `port` are part of the cache/mapping key.

### ⚠️ Issues found

No correctness, security, concurrency, or networking issues found in the supplied diff.

Issue requirement coverage:
- `Met`: after category ipset recycle rebuilds/swaps the ipset, `DomainUpdater` cache is cleared for the matching non-port domain registration.
- `Met`: port-specific domain mappings continue clearing against the domain-port block set.
- `Met`: wildcard exact-match behavior is preserved because `clearIPCacheForDomain()` still derives `exactMatch` internally from the domain pattern.

### 💡 Suggestions

- Add a focused regression test for a non-port category domain: populate `DomainUpdater` cache, simulate recycle, call this path, then verify the next DNS update re-adds the IP to the normal category ipset.
- Add a second assertion for a port-based domain to ensure the refactor keeps the old port-specific behavior intact.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9164  
Head SHA: `03aee0a092df315ba5e9b3eec49e8b11c7715543`  
Checked at: 2026-06-02 11:50:24 CST

jasonlyc (approved):

PR #9166, opened 2026-06-02
sync purple 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The change targets the right root cause: after category ipsets are flushed or rebuilt outside `DomainUpdater`, the in-memory `ipCache` can prevent known IPs from being re-added.
- The new cache key derivation in `_clearIPCacheForDomain()` matches `registerDefaultUpdate()` for both exact domains and `*.` wildcard domains.
- Adding a bounded default LRU lifetime is reasonable; it reduces the blast radius if cache invalidation is missed.

### ⚠️ Issues found

- **Medium: cache clear jobs can run behind already queued DNS update jobs.**  
  `control/DomainUpdater.js:332` enqueues `clearIPCacheForDomain()` as a standard job. Since `updateDomainMapping()` also uses the standard queue, any DNS update already queued when `recycleIPSet()` flushes or swaps the ipset can run before the cache is cleared. That update may still skip cached IPs, leaving the rebuilt ipset missing entries until another DNS update happens.  
  Consider enqueueing cache clears as priority jobs, or clearing synchronously before/at the point where the ipset is externally flushed/swapped.

### 💡 Suggestions

- Add a regression test for this ordering case: queue a domain mapping update with a cached IP, flush/recycle the ipset, then verify the cache clear happens before the queued DNS update can skip the address.
- Consider awaiting the cache clear calls or batching them with `Promise.all()` in `CategoryUpdater.js` so `recycleIPSet()` completion more accurately means the category state and DomainUpdater cache state are both consistent.
- No linked issue found in the PR body or GitHub closing references.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9166  
Checked at: 2026-06-02 13:06:20 CST

MelvinTo (approved):

PR #9167, opened 2026-06-02
sync purple se 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly identifies that DomainUpdater’s `ipCache` can become stale when CategoryUpdater flushes/swaps ipsets outside DomainUpdater.
- Adds targeted cache clearing instead of globally flushing all domain update state.
- Cache-key reconstruction appears aligned with `registerDefaultUpdate()` for the current callers, including wildcard domains and port-specific block sets.
- No linked issue found in PR title/body/closing references.

### ⚠️ Issues found

- **Medium:** `clearIPCacheForDomain()` enqueues the clear operation on the standard queue, same as DNS update jobs. If DNS update jobs are already queued after an ipset flush/swap, they can run before the cache clear and still skip re-adding cached IPs, leaving the rebuilt ipset incomplete until another DNS event arrives.

  `[control/DomainUpdater.js:331]`  
  Consider enqueueing this as a priority job, or otherwise guaranteeing cache clear runs before queued `updateDomainMapping()` jobs after CategoryUpdater rebuilds the ipset.

- **Low:** CategoryUpdater calls `domainUpdater.clearIPCacheForDomain(...)` without awaiting the returned promise. That makes recycle completion/logging independent from whether the cache clear actually ran.

  `[control/CategoryUpdater.js:1694]`  
  `[control/CategoryUpdater.js:1764]`  
  Prefer awaiting these clears, or batching them with `Promise.all`, so recycle completion reflects the cache invalidation work.

### 💡 Suggestions

- Add a regression test around this sequence: DNS mapping cached → CategoryUpdater flush/swap → DNS update for the same IP → verify the IP is re-added to the correct ipset.
- Consider documenting whether `clearIPCacheForDomain()` expects the original domain string (`*.example.com`) or the stripped suffix (`example.com`), since passing the stripped suffix for wildcard domains would produce an exact-match key and miss the cache.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9167  
Checked at: 2026-06-02 13:04:26 CST

PR #9168, opened 2026-06-02
sync gold se 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The fix targets the right mechanism: `DomainUpdater`’s `ipCache` can suppress re-adding IPs after an ipset is flushed or swapped outside `DomainUpdater`.
- The new cache key logic appears consistent with `registerDefaultUpdate()` for wildcard domains, since both use the stripped domain plus `exactMatch: false`.
- No linked issue found in the PR body or closing issue references.

### ⚠️ Issues found

- **Medium: cache is cleared before the final ipset swap**
  
  In `control/CategoryUpdater.js:1766`, `clearIPCacheForDomain()` is called before `filterIPSetByDomain()` and `swapIpset()`. The comment says the ipsets “were fully rebuilt via swap,” but the swap has not happened yet.

  If a DNS update runs after the cache reset but before `swapIpset()`, `DomainUpdater` can repopulate the cache and add the IP to the old active ipset. Then `swapIpset()` replaces that set, and future DNS updates may skip the IP because it is already cached. That can preserve the same missing-IP behavior this PR is trying to fix.

  Move the cache clearing until after `await this.swapIpset(category)` and clear all affected domains there.

- **Low/Medium: queued cache clear promises are not awaited**

  In `control/CategoryUpdater.js:1695` and `control/CategoryUpdater.js:1766`, `domainUpdater.clearIPCacheForDomain()` returns a promise from `_enqueueJob()`, but the caller does not await it.

  Since the correctness of the recycle depends on the cache actually being reset, the recycle path should wait for these jobs, preferably with `await` or collecting promises and `await Promise.all(...)`.

### 💡 Suggestions

- Consider adding debug logging when `_clearIPCacheForDomain()` misses a registration. A silent miss would make future regressions hard to diagnose because the function has no visible effect unless the exact mapping key exists.
- The new default `ipCache` max age of 10 minutes is reasonable, but it changes behavior for registrations without `ipttl`. A short comment explaining why `0` is unsafe here would help future readers.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9168  
Checked at: 2026-06-02 13:02:04 CST

PR #9169, opened 2026-06-02
sync gold pro 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The PR addresses a real stale-cache problem: when category ipsets are flushed or rebuilt outside `DomainUpdater`, cached IPs can otherwise prevent DNS updates from re-adding entries.
- The `blockSet` / `port` variables make the mapping key and cache-clear call use the same values in the rebuild path.
- Wildcard handling in `_clearIPCacheForDomain()` appears consistent with the existing registration behavior for `*.domain` versus exact domains.
- No linked issue found in the PR metadata.

### ⚠️ Issues found

- `control/CategoryUpdater.js:1695` and `control/CategoryUpdater.js:1767`

  `domainUpdater.clearIPCacheForDomain()` returns a Promise from `_enqueueJob()`, but both call sites fire it without `await`. That means recycle/update completion can be reported before the cache reset has actually run, and any future rejection becomes an unhandled async failure. Since this is part of the consistency fix, the caller should `await domainUpdater.clearIPCacheForDomain(...)` or collect the promises and `await Promise.all(...)`.

### 💡 Suggestions

- Add a small unit/regression test for the specific case this PR fixes: populate `ipCache`, rebuild or flush the category ipset, then verify the next DNS update re-adds the same IP.
- Consider logging when `_clearIPCacheForDomain()` cannot find a matching registration. Silent no-op is okay operationally, but a debug log would help catch option/key mismatches in future category-mode changes.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9169  
Checked at: 2026-06-02 13:00:10 CST

PR #9170, opened 2026-06-02
sync orange 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The fix targets the right failure mode: `recycleIPSet()` can flush/swap ipsets outside `DomainUpdater`, while `DomainUpdater.ipCache` may still suppress re-adding already-seen DNS IPs.
- `clearIPCacheForDomain()` mostly mirrors the wildcard normalization used by `registerDefaultUpdate()`, so exact and `*.` domain keys should resolve correctly.
- Port-aware categories are handled by passing the same `blockSet` and `port` used for `syncDomainIPMapping()`.

Linked issue coverage: No linked issue found.

### ⚠️ Issues found

- **Low - Potential unnecessary repeated ipset work**  
  `control/DomainUpdater.js:314` changes registrations without `options.ipttl` from non-expiring cache entries to a 10-minute max age. That means long-lived domain registrations may reprocess the same DNS IPs every 10 minutes even when no recycle/flush happened. Since this PR adds explicit cache clearing after flush/swap, consider keeping the old no-ttl behavior and relying on `clearIPCacheForDomain()` for the recycle case, unless the 10-minute fallback is intentional and measured.

- **Low - Cache-clear jobs are not awaited**  
  `control/CategoryUpdater.js:1695` and `control/CategoryUpdater.js:1766` enqueue cache clears but do not await them. The queue ordering likely makes this work for future DNS updates, but awaiting would make recycle completion deterministic and avoid subtle timing assumptions if an already-queued DNS update runs around the same time.

### 💡 Suggestions

- Add a focused unit/integration test for: populate `ipCache`, recycle/swap category ipset, then send the same DNS update and verify the IP is re-added to the correct ipset.
- If the 10-minute fallback is kept, add a comment explaining why no-`ipttl` registrations should expire periodically now.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9170  
Checked at: 2026-06-02 12:57:30 CST

PR #9171, opened 2026-06-02
recreate lease file on ENOENT error

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Extracting the watcher setup into `_watchLeaseFile()` makes the reload path easier to follow.
- Retrying `fs.watch()` after transient `ENOENT` is a reasonable direction for lease files that can be recreated by miniupnpd.
- The existing watcher close/reload flow is preserved for normal `rename` events.

### ⚠️ Issues found

**Medium - `_watchLeaseFile()` can block the entire reload loop indefinitely**

[sensor/UPNPSensor.js:122]
Because `scheduleReload()` now does `await this._watchLeaseFile(leaseFile)` for each interface, one lease file that keeps returning `ENOENT` will keep sleeping and recursing forever. That prevents watchers from being installed for later interfaces and prevents `this.scheduleCheckUPnPLeases()` from running at the end of the reload.

This is a behavior regression from the old code, where each failed lease file logged an error and the reload continued. Consider making the retry bounded/backgrounded, or letting `_watchLeaseFile()` schedule a delayed retry and return so the reload can continue.

**Medium - delayed recursive retries can create stale/duplicate watchers after reload**

[sensor/UPNPSensor.js:144]
The retry timer is uncancellable. If `_watchLeaseFile()` hits `ENOENT`, waits 5 seconds, and another `scheduleReload()` runs in the meantime, the old delayed retry can still wake up and add a watcher after the newer reload already closed/rebuilt `this.upnpLeaseFileWatchers`.

That can leave duplicate watchers or watchers for an old interface set. Add a reload generation token, cancellation guard, or track pending retry timers and clear them in `scheduleReload()` alongside closing active watchers.

### 💡 Suggestions

- Add a unit/integration test for the `ENOENT` path where one interface’s lease file stays missing while another interface is valid. Expected behavior should be that valid interfaces still get watched and lease checks are still scheduled.
- Consider avoiding recursive retry here. A loop with a max retry count, or a separate retry task keyed by lease file, would be easier to reason about and safer for lifecycle cleanup.
- No linked issue found in the PR title/body/comments, so issue requirement coverage could not be verified.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9171  
Head SHA: `c3e29a43d4b91b0971c443587340e5725c28dff7`  
Checked at: 2026-06-02 14:54:32 CST

jasonlyc (approved):

PR #9180, opened 2026-06-03
add iptables -S test

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds a focused `--iptables` sanity-check mode.
- Includes the check in the full/default sanity run.
- Captures stderr and prints the command output only on failure, which keeps normal output clean.
- CLI usage text is updated for the new option.

### ⚠️ Issues found

- Low: `sudo iptables -S` does not use the xtables lock wait option. Other parts of this script already use `sudo iptables -w ...`; without `-w`, this new check can fail transiently while firewall rules are being updated, producing a false “blocking & routing might not working correctly” warning. Consider `sudo iptables -w -S` or a bounded wait such as `sudo iptables -w 5 -S`.

### 💡 Suggestions

- If the goal is to validate broader routing/blocking health, consider checking relevant tables explicitly, for example `filter`, `nat`, and `mangle`, since plain `iptables -S` only dumps the default table.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9180  
Head SHA: `643f54bf6feeccf21fe21ddee326edba4489aca6`  
Checked at: 2026-06-03 17:15:43 CST

jasonlyc (approved):

PR #9215, opened 2026-06-09
Orange 1.983 Alpha Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Redis save policy is now platform-specific, which is a cleaner place for GSE tuning.
- Guardian interval suppression for one-shot info fetches is a good fix for accidental background timer buildup.
- LiveStatsPlugin now avoids treating missing switch samples as zero-throughput data.
- No linked issue found in the PR title/body/closing references.

### ⚠️ Issues found

- **Medium: `DomainUpdater.clearIPCacheForDomain()` is queued asynchronously, but callers do not await it.**  
  In `CategoryUpdater.js`, the new calls enqueue a standard DomainUpdater job and then continue. If an existing `updateDomainMapping()` job is already queued, it can run before the cache reset, still skip cached IPs, and then no later DNS update may arrive to re-add them. Since `_clearIPCacheForDomain()` only resets in-memory LRU state, this should probably be synchronous/direct, or at least awaited and prioritized so the cache is cleared before DNS update jobs can observe stale entries.

- **Low/Medium: `APCMsgSensor.js` calls `conntrack.setConnEntry()` without awaiting or catching failures.**  
  `setConnEntry()` is async and writes Redis. If Redis errors, this creates an unhandled rejection path while processing AP messages. Existing similar code in `ACLAuditLogPlugin` awaits this call. Either make the handler path async and `await`, or attach `.catch()` with logging.

### 💡 Suggestions

- Consider adding a focused regression test around the `DomainUpdater` cache clear ordering: flush/recycle an ipset, enqueue a DNS update with an address already present in `ipCache`, and verify the address is re-added after recycle.
- The new local-flow conntrack lookup in `BroDetect.js` expands from AP bridge flows to all local flows. That may be intentional for integrated AP, but it is worth watching Redis lookup volume on busy LAN-local traffic.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9215  
Head SHA: `3f701c9844d7e2341ee88b204e26c3eb50a75302`  
Checked at: 2026-06-09 17:35:29 CST

PR #9217, opened 2026-06-10
modify NoiseDomains bf reload method

chgchi:


jasonlyc (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The change addresses linked issue `firewalla/firecommit#8224`: when one process downloads new NoiseDB data, it now broadcasts `NOISE_DOMAIN_BF_UPDATED` so both `FireMain` and `FireApi` reload from the shared local cache file.
- `cloudcache.download()` writes cache content before invoking `onUpdateCallback`, so reloading from the local file after the event should see the updated data.
- Startup behavior remains covered because `enableCache(..., download(true))` still triggers the callback, and the new event path reloads local cache.
- Missing/empty local cache now clears `this.bloomfilter`, matching the previous invalid-cache behavior.

### ⚠️ Issues found

No blocking correctness, security, concurrency, or performance issues found in the reviewed diff.

Low: `init()` registers a new `sem.on(EVENT_NOISE_DOMAIN_BF_UPDATED, ...)` listener every time it runs. If this sensor can be initialized more than once in the same process during tests or lifecycle reloads, each event will trigger duplicate file reloads. Consider guarding listener registration or clearing the previous listener if repeated init is possible.

### 💡 Suggestions

- Remove the unused `data` parameter in the `enableCache` callback for clarity.
- Consider adding a focused unit/integration test that simulates two processes/listeners: one cloudcache update emits the event, both sensor instances reload the local cache.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9217  
Head SHA: `3944cef5df0ed70510215e40081b614c23899a34`  
Checked at: 2026-06-10 11:11:40 CST

PR #9219, opened 2026-06-10
sync purple 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Reviewed 1/1 changed file: `sensor/NoiseDomainsSensor.js`.
- The intent to decouple cloudcache updates from direct in-memory loading is reasonable.
- `reloadNoiseDomainBFFromLocal()` improves missing-file handling versus the prior test-only loader.
- No linked issue found.

### ⚠️ Issues found

**Medium** `[sensor/NoiseDomainsSensor.js:71]`

`reloadNoiseDomainBFFromLocal()` clears `this.bloomfilter` before the replacement data is successfully read and parsed. If `readFile()` fails with a transient error, or if `loadNoiseDomainBFData()` fails to parse malformed cache content, the sensor silently loses the last known-good bloom filter and `find()` returns no noise matches.

Prefer loading into a temporary filter first, then swap only after successful parse. Only clear the active filter for explicit no-data cases such as confirmed cache removal/`ENOENT`.

**Low** `[sensor/NoiseDomainsSensor.js:46]`

`init()` registers a new global `sem.on(EVENT_NOISE_DOMAIN_BF_UPDATED, ...)` listener every time it runs, with no guard or cleanup. If this sensor is initialized more than once in the same process, each event will trigger duplicate reloads and keep old sensor instances referenced.

Add an initialization guard, clear/register once, or bind the listener in a lifecycle path that cannot be repeated.

### 💡 Suggestions

- Consider keeping the cloudcache callback’s `data` path as a fallback, or at least avoid re-reading the local file when valid content is already provided by `enableCache()`.
- Add a regression test for “reload fails after a valid bloom filter is loaded” to ensure the previous filter remains active.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9219  
Checked at: 2026-06-10 11:54:09 CST

PR #9220, opened 2026-06-10
sync purple se 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Moves reload behavior toward a process-wide event model, so other Firewalla processes can refresh the noise-domain Bloom filter after cloudcache updates.
- Keeps the existing test helper API by forwarding `loadLocalNoiseDomainData4Test()` to the new reload path.
- Adds explicit handling for missing local cache files instead of letting `readFile()` reject through the test helper path.

### ⚠️ Issues found

**Medium - `sensor/NoiseDomainsSensor.js:70`**

`reloadNoiseDomainBFFromLocal()` clears `this.bloomfilter` before the local cache file is successfully read and parsed. If the read fails for a transient reason, or if the file is temporarily invalid/partially written, the sensor loses the last known-good Bloom filter and `find()` will return no noise-domain matches until a later successful reload.

This is a behavior regression from the previous callback path, where `loadNoiseDomainBFData(data)` only replaced the Bloom filter after successful parse/validation.

Suggested fix: read and parse into a temporary BloomFilter first, then swap `this.bloomfilter` only after success. Only clear the existing filter for deliberate invalidation cases such as empty content or `ENOENT`.

**Medium - `sensor/NoiseDomainsSensor.js:50`**

The cloudcache callback now ignores the `data` argument and always reloads from `${cacheFolder}/noise_domain`. When cloudcache calls the callback with `null` to indicate unsupported/invalid cache metadata, the new code may still load an orphaned local cache file if the data file exists without valid metadata. The old behavior deleted the in-memory filter in that case.

Suggested fix: preserve the invalidation semantics by checking `data` in the callback. If `data` is falsy, send an invalidation event or delete locally instead of blindly reloading from disk.

### 💡 Suggestions

- Consider passing the cache payload through the event when possible, or use separate event types for “updated” vs “invalidated”. That would make the expected behavior clearer and avoid relying entirely on local file state.
- If multiple `init()` calls are possible in the same process, guard the `sem.on()` registration to avoid duplicate reload listeners.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9220  
Checked at: 2026-06-10 11:52:37 CST

PR #9221, opened 2026-06-10
sync gold se 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The change correctly moves noise-domain BF refresh toward a local-file reload path, which should allow all processes receiving `NOISE_DOMAIN_BF_UPDATED` to refresh from the same cloudcache artifact.
- `ENOENT` is handled explicitly and clears the BF, matching the intended behavior when cloudcache removes/does not provide valid data.
- No linked issue found in PR title/body/comments/closing references.

### ⚠️ Issues found

- **Medium:** `reloadNoiseDomainBFFromLocal()` clears `this.bloomfilter` before the replacement file is successfully read and parsed. If `readFile()` fails with a transient non-`ENOENT` error, or if `loadNoiseDomainBFData()` catches invalid JSON / malformed BF data, the sensor is left with `null` BF and all `find()` calls return empty results. The previous implementation kept the existing BF when parsing failed because `loadNoiseDomainBFData()` did not clear it first.
  
  Suggested fix: read and parse into a temporary BF first, then swap `this.bloomfilter` only after successful validation. Only clear it for intentional no-data cases such as `ENOENT` or empty content.

### 💡 Suggestions

- Consider making the event listener registration idempotent if `init()` can be called more than once in the same process. Repeated `sem.on(EVENT_NOISE_DOMAIN_BF_UPDATED, ...)` registrations would cause duplicate reloads per event.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9221  
Checked at: 2026-06-10 11:51:13 CST

PR #9222, opened 2026-06-10
sync gold pro 1.983 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Moves noise-domain data loading onto `cloudcache`, removing the custom daily cloud fetch path.
- Handles missing local cache files explicitly and clears `this.bloomfilter` on missing/empty local data.
- Keeps `loadLocalNoiseDomainData4Test()` compatible by delegating to the new local reload path.
- No linked issue found.

### ⚠️ Issues found

- **Medium:** `init()` no longer waits for the bloom filter to be loaded.  
  In `sensor/NoiseDomainsSensor.js`, the `cc.enableCache()` callback now calls `sem.sendEventToAll()`, and the listener does `await this.reloadNoiseDomainBFFromLocal()`. However, `EventEmitter.emit()` does not await async listeners, so `init()` can return while `this.bloomfilter` is still `null`. Previously the data was loaded directly during initialization. This can cause early `find()` calls after startup to miss noise-domain matches.

  Suggested fix: load the local file directly in the `enableCache` callback for the current process, or make `init()` explicitly await `reloadNoiseDomainBFFromLocal()` after `cc.enableCache()`.

- **Low:** Repeated `init()` calls add duplicate event listeners.  
  `run()` and `apiRun()` both call `init()`, and `init()` now always registers `sem.on(EVENT_NOISE_DOMAIN_BF_UPDATED, ...)`. If this sensor is initialized more than once in the same process, each update event triggers multiple reloads. Consider guarding listener registration or using `sem.removeListener`/`sem.off` before adding.

### 💡 Suggestions

- The callback receives `data` from `cloudcache`; using it directly for the local process would avoid the startup race and the extra disk read, while still sending the event to other processes if cross-process reload is needed.
- Consider adding a small test that verifies `await init()` leaves the bloom filter ready when valid cache content exists.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9222  
Checked at: 2026-06-10 11:49:45 CST

PR #9226, opened 2026-06-10
Purple SE 1.983 Alpha Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- New access-request flow has dedicated manager/sensor tests and clear pending/archive state handling.
- Static `/time_limits` serving includes path segment decoding and traversal guards.
- The iptables queue clones `Rule` objects before storing, which avoids common mutable-rule queue bugs.

### ⚠️ Issues found
- **High** [api/routes/time_limits.js:330](https://github.com/firewalla/firewalla/pull/9226/files#diff-PLACEHOLDER) / `alarm/AccessRequestManager.js:455,522`  
  `leftQuota` is supplied by the requesting client, stored in Redis, then trusted during approval to compute effective extra quota and bypass policy quota. A client can submit an inflated `leftQuota`; even if the admin or auto-approve path only approves a small `approvedQuota`, `approveRequest()` adds the untrusted `leftQuota` into `totalQuotaLeft` and `appTimeUsage.quota`, granting more time than approved. Recompute remaining quota server-side from current app usage/policy state at approval time, or clamp/ignore client-provided `leftQuota`.

- **Medium** `scripts/apt-get.sh:20`  
  The previous default `: ${FIREWALLA_HOME:=/home/pi/firewalla}` was removed, but the script still immediately runs `source ${FIREWALLA_HOME}/platform/platform.sh`. Any direct caller without `FIREWALLA_HOME` set now tries to source `/platform/platform.sh` and loses platform detection / `RAMFS_ROOT_PARTITION` behavior. Restore the default before the `source`.

### 💡 Suggestions
- Add a regression test for the extra-time approval path where the request body has a bogus `leftQuota` much larger than the real remaining quota.
- Consider whether the new `timeout 60` around apt update/install is long enough for slower devices and overlayroot installs; previous cron paths used longer timeouts.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9226  
Head SHA: `425fc0a9449a8931110acb3fcb0e836ee47c442d`  
Checked at: 2026-06-10 17:33:27 CST

PR #9233, opened 2026-06-11
Orange 1.983 Alpha Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Moves reload through `SensorEventManager`, which should let other Firewalla processes refresh their in-memory noise-domain bloom filter after cloudcache updates the local cache file.
- `reloadNoiseDomainBFFromLocal()` now handles missing local files without throwing, and clears the in-memory bloom filter on `ENOENT`.
- Parsing remains isolated in `loadNoiseDomainBFData()`, and invalid JSON no longer leaves the previous bloom filter active because reload clears `this.bloomfilter` first.
- No linked issue found in the PR body or closing issue metadata.

### ⚠️ Issues found

- **Medium - `sensor/NoiseDomainsSensor.js:47`**
  The cloudcache callback now ignores the `data` argument and always broadcasts `NOISE_DOMAIN_BF_UPDATED`. When cloudcache calls the callback with `null`, that means the cache is unsupported, obsolete, or otherwise invalid. The previous implementation immediately cleared `this.bloomfilter`; the new flow tries to reload from `${cacheFolder}/noise_domain` instead. If an orphaned/stale local data file exists without valid metadata, this can reactivate stale noise-domain data that cloudcache intended to invalidate.

  Suggested fix: preserve the null-data semantics, for example by sending a separate delete/invalid event or including a validity flag in the event so listeners call `deleteNoiseDomainBFData()` instead of reloading from disk.

### 💡 Suggestions

- Consider making the event handler registration idempotent in `init()`. If `init()` can be called more than once in a process, `sem.on(EVENT_NOISE_DOMAIN_BF_UPDATED, ...)` will accumulate listeners and each update will trigger duplicate file reloads.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9233  
Head SHA: `c0a9e66f51e1c5389b2fb681aa9c042f320c4e67`  
Checked at: 2026-06-11 16:08:45 CST

PR #9234, opened 2026-06-11
Gold 1.983 Alpha Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Redis save policy is now platform-specific, which is cleaner than hard-coding all thresholds in `RuntimeConfigSensor`.
- GSE gets more conservative Redis RDB save intervals, which matches the likely intent for lower write pressure.
- `Guardian` now supports non-daemon instances, avoiding background interval jobs when `HostManager.getGuardians()` only needs read-only initialization data.
- Noise domain BF reload now goes through local cache and broadcasts an update event, which should keep multiple processes in sync after cloudcache refreshes.

### ⚠️ Issues found

No blocking correctness issues found in the reviewed diff.

No linked issue found, so there were no issue requirements to verify against.

### 💡 Suggestions

- `sensor/NoiseDomainsSensor.js`: `init()` now registers a `sem.on(EVENT_NOISE_DOMAIN_BF_UPDATED, ...)` listener every time it runs. If this sensor can be initialized more than once in the same process, reloads will duplicate. Consider guarding initialization or clearing/reusing the listener.
- `sensor/Guardian.js`: the third positional constructor argument works, but `new Guardian(alias, {}, false)` is easy to misread later. A named option like `{ daemon: false }` would be more maintainable if this constructor keeps growing.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9234  
Checked at: 2026-06-11 16:07:05 CST

PR #9235, opened 2026-06-11
Gold SE 1.983 Alpha Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Guardian constructor change is scoped and preserves default daemon behavior while avoiding interval leaks for read-only init-data lookups.
- Redis save config is centralized through `Platform.getRedisSaveConfig()` and GSE override, which is cleaner than embedding platform-specific logic in `RuntimeConfigSensor`.
- Noise domain reload now handles missing local cache files without throwing from the test helper path.

No linked issue found.

### ⚠️ Issues found

**Medium** - `sensor/NoiseDomainsSensor.js:48`

The cloudcache callback now ignores the `data` argument and always emits `NOISE_DOMAIN_BF_UPDATED`. When `cloudcache.download()` calls the callback with `null` because cloud metadata is unavailable/unsupported, the old behavior cleared the bloom filter. The new behavior reloads from the local cache file instead, which can resurrect stale local data if the cache content exists but metadata is missing/corrupt.

Suggested fix: preserve the null-data path:

```js
if (data) {
  sem.sendEventToAll({ type: EVENT_NOISE_DOMAIN_BF_UPDATED });
} else {
  this.deleteNoiseDomainBFData();
}
```

or include an explicit event payload that tells receivers to clear instead of reload.

**Low** - `sensor/NoiseDomainsSensor.js:67`

`reloadNoiseDomainBFFromLocal()` clears `this.bloomfilter` before the file read succeeds. For non-`ENOENT` read failures, the catch block logs the error but leaves the sensor with no bloom filter even though the previous filter may still be valid. That creates a transient IO/error path where noise-domain matching is disabled.

Suggested fix: read and parse into a local filter first, then swap `this.bloomfilter` only after success; explicitly clear only for empty/missing/invalid cache cases.

### 💡 Suggestions

- Consider adding a small unit test around `NoiseDomainsSensor.init()` covering `enableCache(... callback(null))` to lock in whether null cloudcache data should clear or reload from disk.
- Redis save tier changes look intentional, but it would help to add a focused test for the 60/120/240MB boundaries so future tuning does not accidentally invert intervals or thresholds.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9235  
Head SHA: `6e019474364016a620107490fcca83aef53289ec`  
Checked at: 2026-06-11 16:05:41 CST

PR #9236, opened 2026-06-11
Purple 1.983 Alpha Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Moves noise-domain bloom filter reload to an event-driven path so multiple processes can refresh from the local cloudcache artifact.
- Adds `ENOENT` handling for missing local cache data instead of letting `loadLocalNoiseDomainData4Test()` throw.
- Keeps the existing test helper API by forwarding `loadLocalNoiseDomainData4Test()` to the new reload method.
- No linked issue found in the PR title/body/comments/closing references.

### ⚠️ Issues found

- **Medium** `[sensor/NoiseDomainsSensor.js:70]`  
  `reloadNoiseDomainBFFromLocal()` clears `this.bloomfilter` before the new file is successfully read and parsed. If `readFile()` fails with a transient non-`ENOENT` error, or if the file content is corrupted/partially written/invalid JSON, the existing valid bloom filter is lost and noise-domain detection becomes disabled. Prefer loading into a temporary bloom filter and only swapping `this.bloomfilter` after successful validation; clear it only when the cache is explicitly absent/obsolete.

- **Low** `[sensor/NoiseDomainsSensor.js:45]`  
  `init()` registers a new `sem.on(EVENT_NOISE_DOMAIN_BF_UPDATED, ...)` listener every time it runs. Since both `run()` and `apiRun()` call `init()`, repeated initialization can create duplicate reload handlers on the same instance. Consider guarding registration or clearing/reusing the listener.

### 💡 Suggestions

- The `enableCache(hashKey, (data) => { ... })` callback now ignores `data` and reloads from disk. That is probably intentional for cross-process consistency, but a short comment would help future readers understand why the in-memory callback payload is not used directly.
- Consider making `loadNoiseDomainBFData()` return a boolean success/failure so `reloadNoiseDomainBFFromLocal()` can decide whether to preserve or clear the previous filter.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9236  
Head SHA: `4b917040e0e27fa8f696936fc2d1cc16f468a7ff`  
Checked at: 2026-06-11 16:03:51 CST

PR #9238, opened 2026-06-12
support rx-flow-hash config in apply_profile.sh

yijin-png:
**Support rx-flow-hash config in apply_profile.sh**

Adds a new rx_flow_hash profile key so RX flow hashing can be tuned per-NIC via ethtool.

Config format
  
```
  "rx_flow_hash": [
    [ "eth1", "udp4", "sdfn" ],
    [ "eth1", "udp6", "sdfn" ]
  ]
```
Fields: interface, flow-type, hash-opts

### DIFF
--- a/scripts/apply_profile.sh
+++ b/scripts/apply_profile.sh
@@ -61,6 +61,39 @@ set_nic_feature() {
     done
 }
 
+# examples
+#
+# "rx_flow_hash": [
+#     [ "eth1", "udp4", "sdfn" ],
+#     [ "eth1", "udp6", "sdfn" ]
+# ]
+#
+# fields: <interface> <flow-type> <hash-opts>
+#   flow-type: tcp4|udp4|tcp6|udp6|ah4|esp4|sctp4|...
+#   hash-opts: combination of m v t s d f n r (e.g. sd, sdfn)
+set_rx_flow_hash() {
+    while read nic flow_type hash_opts
+    do
+        local nics=
+        if [[ "$nic" == *"*"* ]]; then
+            shopt -s nullglob
+            for matched_nic in /sys/class/net/$nic; do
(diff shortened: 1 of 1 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds `rx_flow_hash` in the same profile-processing pattern as existing keys like `nic_feature`.
- Supports wildcard NIC selection consistently with the existing `set_nic_feature` behavior.
- `PROFILE_CHECK` mode uses read-only `ethtool -n`, while normal mode applies via `ethtool -N`.
- PR body clearly documents the new JSON shape.
- No linked issue found.

### ⚠️ Issues found

- **Medium - partial failures can be hidden**  
  `scripts/apply_profile.sh:87-92`  
  If multiple NICs or rows are processed, an `ethtool` failure on an earlier NIC can be masked by a later successful command, so `process_profile` may report success even though part of the profile was not applied. This is especially likely with wildcard NIC patterns where some matched interfaces may not support the requested `rx-flow-hash` type. Consider accumulating a local `rc` in `set_rx_flow_hash` and returning non-zero if any `ethtool` call fails.

### 💡 Suggestions

- Quote the simple command arguments where glob expansion is not intended, e.g. `"$n"`, `"$flow_type"`, and `"$hash_opts"`, to make malformed profile values fail predictably instead of being split into extra arguments.
- Consider adding a log line when a wildcard pattern matches no interfaces. Silent no-op behavior makes profile mistakes harder to diagnose.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9238  
Head SHA: `8d5e0d1d89cb5297641e993ed0e6092e9bb0d352`  
Checked at: 2026-06-12 15:49:46 CST

jasonlyc (approved):

PR #9241, opened 2026-06-15
Orange 1.983 Beta Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Redis save policy is now platform-specific, which is cleaner than keeping GSE thresholds inside `RuntimeConfigSensor`.
- `Guardian` now supports non-daemon use, avoiding extra cleanup intervals when `HostManager` only needs to read guardian info.
- `LiveStatsPlugin` now safely handles missing switch metric samples.
- Domain updater cache invalidation generally follows the right intent after direct ipset flush/swap.
- No linked issue found in the PR title/body/comments/closing references.

### ⚠️ Issues found
**Medium: `sensor/NoiseDomainsSensor.js:49`**

The cloudcache callback now ignores the `data` argument and always broadcasts `NOISE_DOMAIN_BF_UPDATED`. This changes the invalid-cache path: `cloudcache` calls the callback with `null` when cloud metadata is missing/unsupported or cache data should be cleared. Previously that immediately deleted the in-memory bloom filter. With this change, the sensor reloads from `${cacheFolder}/noise_domain` instead, so a stale local cache file can be reloaded even though cloudcache signaled “no valid data”.

Please preserve the `null` semantics, for example by sending an explicit clear event or by deleting locally when `data` is falsy instead of blindly reloading from disk.

### 💡 Suggestions
- Consider awaiting `domainUpdater.clearIPCacheForDomain(...)` in `CategoryUpdater` or intentionally collecting promises. It currently enqueues work and drops the promise; if a future change makes cache clearing throw, this can become an unhandled rejection.
- Consider awaiting `conntrack.setConnEntry(...)` in `APCMsgSensor`, or explicitly catching/logging failures, since Redis write failures would otherwise be detached from the message processing path.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9241  
Head SHA: `34d2c083a3b83be9a69514dc37eb2d0f0d911d63`  
Checked at: 2026-06-15 16:34:04 CST

PR #9246, opened 2026-06-16
Gold SE 1.983 Beta Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- `Guardian` now supports non-daemon read-only instances, and `HostManager.getGuardians()` uses that path to avoid creating background cleanup/truncate intervals during API reads.
- Redis save tuning is centralized through `getRedisSaveConfig()` overrides and keeps platform-specific behavior separated between base and GSE platforms.
- Noise domain loading now uses `cloudcache`, which is a cleaner fit than direct cloud polling and legacy Redis storage.

### ⚠️ Issues found

- **Medium: `NoiseDomainsSensor` can load stale local data when cloudcache reports no valid data.**

  In `sensor/NoiseDomainsSensor.js:50`, the `enableCache()` callback ignores the `data` argument and always emits a reload event. When `cloudcache` calls the callback with `null` for unsupported/missing cloud metadata, the new code re-reads `${cacheFolder}/noise_domain`. If an orphaned or stale cache file exists without valid metadata, the sensor will load it. The previous implementation explicitly deleted the bloom filter when `data` was falsy.

  Suggested fix: preserve the `data` check in the callback, or include validity in the event:

  ```js
  await cc.enableCache(hashKey, (data) => {
    if (data) {
      sem.sendEventToAll({ type: EVENT_NOISE_DOMAIN_BF_UPDATED });
    } else {
      this.deleteNoiseDomainBFData();
      sem.sendEventToAll({ type: EVENT_NOISE_DOMAIN_BF_UPDATED, valid: false });
    }
  });
  ```

  Even better, the local process can load from `data` directly and only use the event to notify other processes.

### 💡 Suggestions

- Add a small regression test for `NoiseDomainsSensor` covering `enableCache()` callback with `null` data while a stale local cache file exists.
- Consider documenting the `daemon` constructor parameter in `Guardian`, since it changes lifecycle behavior and future callers may otherwise miss why intervals are suppressed.
- No linked issue found.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9246  
Head SHA: `ea33133f5622449b53b8e2c8231811a3aaa10edd`  
Checked at: 2026-06-16 17:15:00 CST

PR #9253, opened 2026-06-18
reduce redis redundant writes

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The PR targets a real Redis write-amplification problem and keeps the changes narrowly scoped.
- Conntrack TTL throttling uses a deferred drain, which is the right general shape for avoiding permanent TTL loss during normal operation.
- `deviceLastFlowTs` throttling is reasonable because the key has no TTL and the reader appears to use coarse time windows.
- Reviewed 6 changed files. No linked issue found in the PR metadata/body.

### ⚠️ Issues found

- **Medium: deferred TTL refreshes are dropped on Redis failure.**  
  In `net2/Conntrack.js` and `net2/DNSTool.js`, `_drainConnTTL()` / `_drainDnsTTL()` replace the pending map before `multi.execAsync()`, then update the last-refresh cache before knowing whether Redis accepted the `EXPIRE`s. If Redis errors, those pending refreshes are lost and the throttle cache still says they were refreshed. This can leave keys with stale TTLs until a later write happens, and contradicts the comment that throttled refreshes are “deferred, not dropped.”

- **Medium: metrics `EXPIREAT` cache is updated before the batch succeeds.**  
  In `util/redis_manager.js`, `execBatch()` calls `rememberExpireAt()` while building the batch, before `batch.execAsync()` succeeds. If the batch fails, future writes for the same time-series key may skip `EXPIREAT` because the local cache believes it was already applied. That can create metrics keys without expiration after the next successful `HINCRBY`.

### 💡 Suggestions

- Requeue pending TTL refreshes on drain failure, and only update `connExpireTs` / `dnsExpireTs` after Redis confirms success.
- In `execBatch()`, collect keys whose `EXPIREAT` should be remembered, execute the batch, then update `mclientExpireAtCache` only after successful execution.
- Consider adding small unit tests around Redis failure paths for the TTL throttles, since these are exactly the paths where local cache and Redis state can diverge.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9253  
Head SHA: `e17548fcb1b3476378aa88e5c14a1b984a002468`  
Checked at: 2026-06-18 10:16:07 CST

jasonlyc (approved):

PR #9256, opened 2026-06-18
skip link-local IPv6 resolvers in fallback DNAT

melo-crypto:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Correctly addresses the `ipset` limitation by removing the IPv6 zone identifier before adding DNS6 entries to the gateway IPv6 ipset.
- Scope is minimal and localized to the `dns6` gateway set population path.
- PR metadata has no linked issue; no linked issue found.

### ⚠️ Issues found
- No blocking correctness, security, performance, or concurrency issues found in the changed diff.

### 💡 Suggestions
- Consider using `dns6.split('%', 1)[0]` or a tiny helper if scoped IPv6 normalization is needed elsewhere later, but the current logic is fine for this path.
- A small unit/regression test with `fe80::1%eth0` would help prevent this from coming back.

### Verdict
APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9256  
Head SHA: `fd23b5e97eb9415b9a1d4ad97e33c5446132e05c`  
Checked at: 2026-06-18 16:51:10 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly avoids generating IPv6 fallback DNAT rules toward link-local resolvers, which are not routable destinations for this DNAT path.
- Keeps the DNAT target normalization by stripping the zone suffix before emitting the iptables destination.
- PR is small and localized to IPv6 resolver handling.
- No linked issue found in PR metadata/body.

### ⚠️ Issues found

- **Low:** `extension/dnsmasq/dnsmasq.js:1622` now calls `resolver6.filter(...)` before the previous `resolver6 && resolver6.length > 0` guard. `sysManager.myResolver6()` currently returns an array, so this is not an immediate runtime bug, but it weakens the defensive behavior of this path. Safer pattern: compute `routableResolver6` only after confirming `resolver6` is an array/non-empty, or use `(resolver6 || []).filter(...)`.

- **Low:** `net2/NetworkProfile.js:590` skips link-local DNS6 entries from the gateway IPv6 ipset, but the nearby comment still says DNS6 is added because IPv6 gateways typically use link-local addresses. If this behavior change is intentional, update the comment. If the original intent was only to avoid scoped `%iface` values breaking ipset, consider stripping the zone suffix instead of skipping all link-local DNS6 entries.

### 💡 Suggestions

- Add a regression test or fixture covering `fe80::1%eth0` and a global IPv6 resolver, verifying that fallback DNAT skips the link-local resolver and still emits rules for the global resolver.
- Consider a small helper for scoped IPv6 normalization/filtering if this logic appears in both dnsmasq and network profile paths.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9256  
Head SHA: `0593957a57ca7a57808e44c0fbab18a46e65283f`  
Checked at: 2026-06-18 18:09:02 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly skips link-local IPv6 resolvers before generating fallback DNAT rules, avoiding unusable DNAT targets like `fe80::/10`.
- Keeps scoped-address normalization for the actual DNAT destination with `.split('%')[0]`.
- Uses `(resolver6 || [])`, so the fallback rule path remains defensive if resolver discovery returns null/undefined.
- `ip-address@6.2.0` treats `fe80::1%eth0` as valid and link-local, so the scoped link-local case is handled by the filter.
- No linked issue found in PR metadata/body.

### ⚠️ Issues found

- No blocking correctness, security, performance, concurrency, or networking issues found in the current diff.

### 💡 Suggestions

- `net2/NetworkProfile.js:587`: The comment says “global unicast DNS6”, but the code only filters link-local addresses. ULA, loopback, multicast, or other non-link-local IPv6 values would still pass if present. Consider wording this as “non-link-local DNS6” unless global-only behavior is intended.
- Add a small regression test/fixture covering mixed resolvers such as `fe80::1%eth0` and `2001:4860:4860::8888`, verifying that fallback DNAT skips the link-local resolver and still emits the global resolver rule.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9256  
Head SHA: `6ba8b8c61196fd2614eab67a72624515667cc4af`  
Checked at: 2026-06-18 18:50:55 CST

jasonlyc (approved):

PR #9260, opened 2026-06-22
Get bonjour config from assets, implement service priority 

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The VPN interface skip fix in `BonjourSensor` is correct; the previous nested `continue` only skipped the inner prefix loop.
- Moving Bonjour config into an asset-backed helper is a good direction, with local config fallback for missing lists.
- The per-field Bonjour priority filter is scoped well and avoids persisting priority numbers, which keeps future asset tuning flexible.
- No linked issue found.

### ⚠️ Issues found

- **Medium - `sensor/BonjourSensor.js:335`**  
  `ignoreNames` changed from substring matching to exact list membership. The old code ignored a Bonjour name when any configured pattern was contained in the name; the new `isServiceInList(name, 'ignoreNames')` only ignores exact full-name matches. This will allow previously ignored noisy names to be written as `host.bname`. Keep substring semantics for `ignoreNames`, while using exact matching for service-type lists.

- **Medium - `sensor/BonjourSensor.js:236`**  
  `_printer` now always emits `detect.type = 'printer'`, removing the previous `!detected.type` guard. The new priority filter only compares existing Bonjour sources, so it does not protect existing non-Bonjour or flattened detections from being overwritten by a low-confidence printer service. Either restore the guard for existing detected type, or extend the priority logic to account for existing non-Bonjour detection precedence.

### 💡 Suggestions

- Consider catching/logging errors around the async Bonjour browser callback, since `bonjourParse()` now awaits asset lookups and the callback at `sensor/BonjourSensor.js:87` does not await or handle promise rejections.
- Add focused tests or fixtures for:
  - `ignoreNames` substring behavior.
  - lower-priority `_printer` not overwriting higher-confidence device type.
  - higher-priority Bonjour service replacing lower-priority pending/persisted Bonjour fields.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9260  
Head SHA: `e059bcaf2dc5b4c4d776febe037c25bbf380f74e`  
Checked at: 2026-06-22 18:04:09 CST

jasonlyc (approved):

PR #9261, opened 2026-06-22
enable tls and udp_tls block on ad-block

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds TLS and UDP-TLS rule creation for global, device, tag/group, network, and identity adblock scopes.
- Uses existing `Block.setup*Rules()` helpers and `tlsHostSet` support, so the iptables path is consistent with existing policy enforcement.
- Enables strict-mode TLS category handling via `CategoryUpdater` and reserves a synthetic PID to avoid future user-policy allocation collisions.

### ⚠️ Issues found

- **Medium - Default-mode TLS hostset can remain empty on first scope enable.**  
  `sensor/AdblockPlugin.js:565` activates the default TLS set after creating a rule, but it does not repopulate the hostset from `adblock_list:*`. `_syncDefaultTlsHostSet()` runs before `_reapplyTlsRules()` during filter updates (`sensor/AdblockPlugin.js:255-258`), so if no adblock TLS rule existed yet, the kernel `/proc` hostset may not exist and the writes can be dropped/logged. Later enabling adblock for a device/group/network/global scope creates the rule but only activates the set, leaving TLS/UDP-TLS blocking ineffective until the next filter refresh.  
  Fix: after creating/activating the default-mode rule, call `_syncDefaultTlsHostSet()` or otherwise ensure the hostset is populated after the `/proc` file exists.

- **Low - Existing policy PID collision is not handled.**  
  `alarm/PolicyManager2.js:418` prevents future allocation of `65500`, but deployed boxes may already have `policy:65500` because `POLICY_MAX_ID` is `65535`. The synthetic adblock TLS rules would then share the same mark/comment namespace as an existing user policy.  
  Fix: add a migration/guard for existing `policy:65500`, or choose a reservation strategy that can prove no existing policy owns the PID.

### 💡 Suggestions

- Add tests or a small integration check for these flows:
  - adblock lists already loaded, no active scope, then enable device/group/network/global adblock
  - switch default ↔ strict mode while scopes are active
  - TCP-only, UDP-only, and TCP+UDP TLS platform support
- Consider making `_addTlsRules()` stop early if `platform.installTLSModules()` fails, instead of logging and continuing into rule setup attempts.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9261  
Head SHA: `678fa2ff31d35c8d2fff5c391ea09484b4d67b18`  
Checked at: 2026-06-23 12:28:52 CST

PR #9266, opened 2026-06-25
add mac type constraint

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds a reasonable central hook for MAC-vendor-based type constraints via `getMacVendorAllowedTypes()`.
- Applies vendor validation to `ua`, `bonjour`, and `nameBased` sources, which matches the main failure mode in the linked issue: Bonjour/UA/name guesses overriding more reliable vendor evidence.
- The delayed sweep in `_mergeAndSave()` is a good idea because `macVendor` may arrive after an earlier detection update.

### ⚠️ Issues found

- **High** `[extension/detect/common.js:42]`  
  `getMacVendorAllowedTypes()` loads `detect/common/macTypeConstraint.json`, but that file is not included in this PR. The PR only changes three files, and the referenced asset path returns 404 at commit `e9e2af9d...`. As written, the new constraint system has no rule data and will either no-op or fail depending on `assetsManager.get()` behavior. This means the linked issue is not actually fixed.

- **Medium** `[sensor/BonjourSensor.js:340-341]`  
  The new `ignoreNames` check bypasses the existing `isServiceInList()` helper and loses its fallback to `this.config.ignoreNames`. Existing deployments/configs that rely on sensor config instead of the asset list will stop ignoring those names, causing a regression in Bonjour naming behavior. Keep the helper fallback or extend the helper to support substring matching.

### 💡 Suggestions

- **Issue requirement coverage:**  
  - Met: PR introduces vendor/type consistency checks for detected device type.
  - Partially met: The constraint data file is missing, so no vendor constraints can actually be applied.
  - Partially met: Devices with `macVendor: Unknown` in the linked issue still cannot benefit from this approach.
  - Partially met: The check only runs when a source has `sub.type`; suspicious brand/model/os fields without a type can still survive.

- Consider sorting MAC vendor constraint keys by descending length before substring matching. The comment says “first match wins,” but JSON insertion order makes overlapping keys fragile if both a broad and specific vendor key exist.

- Consider caching the parsed MAC vendor constraint map if `assetsManager.get()` is not already cached internally, since this can run during frequent detection updates.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9266  
Head SHA: `e9e2af9d0b6e627979157f329f268c4f2c1d2e12`  
Checked at: 2026-06-25 15:57:10 CST

jasonlyc (approved):

PR #9271, opened 2026-06-26
sync gold 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Good direction on `api/routes/sensor.js`: the sensor-name whitelist closes a direct dynamic `require()` traversal risk.
- The PR adds useful test coverage around access request/time limit, bypass, DNS service resilience, SSH, VPN client, and dynamic reload flows.
- Several async ipset/iptables calls are now awaited, which should reduce ordering races in firewall setup/teardown.
- PR metadata has no body and no `closingIssuesReferences`, so: **No linked issue found.**

### ⚠️ Issues found

- **High** `[control/IpsetControl.js:113]`  
  `_execOne()` executes `sudo ipset -! ${line}` through a shell. Many `line` values are assembled from ipset targets/comments, including domain/category-derived comments. If any untrusted or cloud/user-controlled value contains shell metacharacters, autonomous-mode ipset operations become command-injection capable. Use `spawn/execFile` with argv, or route all autonomous single-command operations through `ipset restore` stdin/file without shell interpolation.

- **Medium** `[control/Bypass.js:145]`  
  Non-category DNS bypass unenforce does not mirror enforce. Enforce adds DNS bypass entries for each affected policy using `[policy.target]` and `aPid`, but unenforce calls `removePolicyFilterEntry(targets, ...)`, where `targets` are from the bypass policy, not the affected policies. This can leave stale dnsmasq bypass entries after a bypass policy is removed. Iterate `affectedPids` on unenforce and remove the exact affected-policy targets/options that were added.

- **Medium** `[alarm/PolicyDisturbManager.js:103]`  
  `_getAppNameFromTarget()` no longer strips the legacy `TLX-fw-` prefix. Existing targets with that prefix will resolve as `TLX-fw-<app>` instead of `<app>`, so app-specific disturb config and `disableQuic` lookup can silently stop applying. Unless all stored/frontend targets have migrated, keep both `TLX-fw-` and `TLX-dt-` supported.

- **Low** `[alarm/Policy.js:456]`  
  The new protocol guard only rejects when both `this.protocol` and `alarm['p.protocol']` exist. For a port-based policy with a protocol set, alarms missing `p.protocol` still match, which contradicts the comment and can over-match protocol-specific port rules. Treat missing alarm protocol as non-match when the policy specifies one, or document why protocol-less alarms should still match.

### 💡 Suggestions

- Add focused regression tests for bypass DNS unenforce, legacy `TLX-fw-` disturb targets, and autonomous ipset execution with comments containing spaces/special characters.
- Consider making the bypass chain existence/removal logic resilient across process restarts instead of relying only on the in-memory `initializedBypassChain` map.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9271  
Head SHA: `0427d015e0cdd2cc89efc76fea83d92ff5c2ec8e`  
Checked at: 2026-06-26 11:31:40 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The PR adds broad regression coverage around bypass rules, DNS service resilience, access requests, SSH, VPN client, OpenSSL, and policy behavior.
- `api/routes/sensor.js` now validates sensor names before dynamic `require()`, which closes a path traversal class of bug.
- Several firewall/ipset/iptables paths were converted to `await`, reducing ordering races during rule setup and teardown.
- No linked issue found in the PR body or GitHub `closingIssuesReferences`.

### ⚠️ Issues found

- **High** `[control/IpsetControl.js:113]`  
  `_execOne()` runs `sudo ipset -! ${line}` through a shell during autonomous mode. Many ipset lines are assembled from domain/category/policy-derived values, so a malicious or malformed value containing shell metacharacters could become command injection. Prefer `spawn/execFile` with argv, or always apply ipset operations through `ipset restore` via stdin/temp file.

- **Medium** `[control/Bypass.js:145]`  
  DNS bypass unenforce does not mirror enforce for non-category rules. Enforce adds entries per affected policy using `[policy.target]` plus `aPid`, but unenforce removes using the bypass policy `targets` and no per-affected-policy `aPid`. This can leave stale dnsmasq bypass entries after deleting a bypass policy. Iterate `affectedPids` and remove the exact target/options that were added.

- **Medium** `[alarm/PolicyDisturbManager.js:101]`  
  `_getAppNameFromTarget()` now strips only `TLX-dt-`; the old implementation also supported `TLX-fw-`. Existing stored/frontend targets with the legacy prefix will resolve to `TLX-fw-<app>` and miss app-specific disturb config, including `disableQuic`. Keep backward compatibility unless all persisted targets are migrated.

- **Low** `[alarm/Policy.js:456]`  
  The new protocol guard only checks when both the policy protocol and alarm `p.protocol` are present. A protocol-specific port rule can still match alarms missing `p.protocol`, which contradicts the comment and may over-match. If the policy specifies a protocol, missing alarm protocol should probably be treated as non-match.

### 💡 Suggestions

- Add regression tests for DNS bypass unenforce with multiple affected policies and different targets.
- Add compatibility coverage for legacy `TLX-fw-` disturb targets.
- Add an autonomous-mode ipset test using comments/targets with shell metacharacters to verify command execution is not shell-interpreted.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9271  
Head SHA: `ec857f1696d1d0d6bb34e8703233a758c2ce1c10`  
Checked at: 2026-06-26 12:02:30 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Good security tightening in `api/routes/sensor.js`, `_removeSingleUPnP()`, common-name validation, and `pkill -x` usage.
- The PBR/VPN route separation looks directionally correct: explicit route rules now use PBR ipsets and skip the Direct-mode gate.
- Several correctness fixes are solid, including tag removal writing back to `tag`, protocol-aware alarm policy matching, and safer TLS set command parsing.
- No linked issue found. The PR page has no description, so I could not verify against issue-level acceptance criteria.

### ⚠️ Issues found

**High: `extension/dnscrypt/dnscrypt.js`**
The new fallback path only logs when no selected DNSCrypt servers are found, but it still writes an empty server list into the config. That means DNSCrypt can start with no usable resolver despite the warning saying it is falling back to all servers. Set `serverList = allServerNames` or the default server list before replacing `%DNSCRYPT_SERVER_LIST%`.

**Medium: `control/Bypass.js`**
`bypassDNSRules()` enforces category bypass entries from the affected policies’ categories, but unenforce removes entries using `targets` from the bypass policy. If those are not the same set, dnsmasq bypass entries can remain after the bypass policy is removed. The unenforce path should reconstruct the same category/category-BF/seq keys used during enforce, or persist enough metadata to remove the exact generated entries.

**Medium: `control/IpsetControl.js`**
`restore(ops, true)` writes to the long-lived interactive `ipset` process and returns immediately without confirming that ipset accepted the commands. Since stderr is only logged asynchronously, callers can observe success while rules were rejected or the process died after the write. For critical enforcement paths, this should either wait for command completion/health or be limited to explicitly best-effort paths.

**Low: `api/lib/getGid.js`**
When `ui.conf` is unreadable or missing `gid`, the self-heal write replaces the whole file with `{ gid }`. If `ui.conf` can contain other settings, this silently drops them. Prefer merging into the existing parsed object when possible, and only create a minimal file when the file truly does not exist or cannot be parsed.

### 💡 Suggestions

- Add focused tests for DNSCrypt config generation when selected servers are stale or empty.
- Add bypass enforce/unenforce regression coverage for category policies, BF category names, and mixed `scope`/`tags`/`intfs` cases.
- Consider documenting which ipset writes are allowed to be deferred, since the new interactive mode changes execution semantics.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9271  
Head SHA: `9873983fb782c82dbca43ef9575b6d618e7859e7`  
Checked at: 2026-06-26 12:37:11 CST

PR #9273, opened 2026-06-29
use in-mem queue for ip_set_to_be_processed

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Replacing the Redis polling loop with a bounded-concurrency in-process queue removes the fixed 500 ms polling delay and avoids JSON round-tripping for normal `DestIPFound` flow objects.
- `processIP()` keeps backward compatibility for string inputs, so direct callers and legacy-style JSON/string payloads still work.
- SysInfo was updated to read the new diagnostic queue metric instead of the old Redis zset count.
- No linked issue found.

### ⚠️ Issues found

- **Medium - queued intel work is now lost across FireMain restart/crash**  
  `hook/DestIPFoundHook.js:80`, `hook/DestIPFoundHook.js:496`  
  The old Redis zset preserved pending work across process restarts. The new `LimitedQueue` keeps all pending flows in memory, and startup explicitly deletes the legacy Redis key. If FireMain restarts while the queue has backlog, those flows are dropped and `MSG_FLOW_ENRICHED` may never be emitted for them. If this is intentional, the PR should call it out; otherwise keep a durable handoff or drain/migrate the Redis backlog before deleting it.

- **Low - queue size/backpressure undercounts active work**  
  `util/asyncNative.js:85`, `hook/DestIPFoundHook.js:507`  
  `LimitedQueue.size` only reports items waiting in `_queue`, not the items currently being processed by up to 100 workers. That means SysInfo can show `0` while 100 intel lookups are active, and pause/resume logic only reacts to queued backlog, not total in-flight load. Consider tracking `_active` and exposing `pending + active`, or naming the metric explicitly as pending-only.

- **Low - `addip` can generate invalid JSON for unescaped input**  
  `scripts/alias.sh:106`  
  The shell function interpolates `$1` directly into a JSON string. An IP-like value is fine, but quotes/backslashes or accidental extra input will produce malformed JSON. Use `jq -n --arg ip "$1" ...` or validate the argument before publishing.

### 💡 Suggestions

- Add a small unit test for `LimitedQueue` covering concurrency limit, handler errors, and size/active-count semantics.
- Consider publishing the queue metric immediately on enqueue/dequeue completion, or accept/document the current 10-second diagnostic lag.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9273  
Head SHA: `30a31317a4ad18ae87433839f4ab8e34d5f7dbf7`  
Checked at: 2026-06-29 13:56:08 CST

jasonlyc (approved):

PR #9278, opened 2026-07-01
Record switch ACL accounting events as local flows

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Switch ACL accounting is now routed through the normal `processConnData` local-flow pipeline, which should populate `flow:local` history instead of only coarse traffic totals.
- `fd='lo'` matches the linked issue expectation: switch pseudo flows should not be treated as directional `in`/`out` flows.
- Connection-count timeseries are skipped for switch accounting, which is correct because these records are byte aggregates, not real connections.
- Linked issue coverage: **Partially met** for `firewalla/firecommit#8573`. The PR creates 30-second pseudo local flows with no port/protocol and `fd=lo`, but there is one coverage caveat below.

### ⚠️ Issues found

[vpn/VpnManager.js:944] **High:** `ovpngen.sh` no longer runs through `sudo`.

The PR removes both `sudo` and `-E` from the `execFileAsync('flock', ...)` path:

```js
['-n', vpnLockFile, './ovpngen.sh', ...]
```

The surrounding changes preserve `sudo` while removing `-E`, so this looks accidental. If `ovpngen.sh` requires elevated permissions like before, VPN profile generation will fail. This should likely be:

```js
['-n', vpnLockFile, 'sudo', './ovpngen.sh', ...]
```

[sensor/APCMsgSensor.js:733] **Medium:** switch pseudo flows are dropped whenever either host lacks a known IPv4 address.

The linked issue says switch accounting cannot provide L3+ data, but this implementation requires both `host1.o.ipv4Addr` and `host2.o.ipv4Addr` before generating any local flow. That means MAC-only known devices, IPv6-only devices, or stale host records will still have no minute-level history even though switch ACL accounting has byte data for them. If this is an intentional product constraint, it should be called out; otherwise consider a MAC-only local-flow path or a fallback placeholder that preserves the accounting record.

[sensor/APCMsgSensor.js:733] **Low:** `tx_bytes + rx_bytes > 0` does not normalize missing values before the check.

The generated flow defaults each side with `|| 0`, but the guard does not. If one field is missing, `undefined + value` becomes `NaN` and the record is skipped even when the other side has traffic. Safer:

```js
const tx = Number(tx_bytes) || 0;
const rx = Number(rx_bytes) || 0;
if (ip1 && ip2 && tx + rx > 0) { ... }
```

### 💡 Suggestions

- Add a regression test or small fixture for switch ACL accounting covering:
  - bidirectional bytes produce correct upload/download for both peers
  - one-way bytes are still recorded
  - `fd` is `lo`
  - port/protocol are absent
- In `FlowAggregationSensor.processEnrichedFlow`, consider explicitly rejecting `fd === 'lo' && !local` so malformed non-local events cannot enter the local-style aggregation path.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9278  
Head SHA: `1e1cc7ab2daffdc0cfeeb03e6e774171742f59ad`  
Checked at: 2026-07-01 17:25:08 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Routes switch ACL accounting through the existing local-flow pipeline instead of maintaining a separate aggregation path.
- Uses `fd = "lo"` for switch pseudo flows, matching the issue expectation that these flows are neither inbound nor outbound.
- Skips connection-count timeseries for switch accounting, which is correct because these are 30-second byte aggregates, not real connections.
- Removes the dedicated `MSG_FLOW_SWITCH_ACCOUNTING` path and simplifies `FlowAggregationSensor`.

### ⚠️ Issues found

[sensor/APCMsgSensor.js:733] **Medium:** Switch accounting records are dropped unless both devices have known IPv4 addresses.

Issue `firewalla/firecommit#8573` says switch accounting cannot provide L3+ information, but still asks to generate pseudo local flows from each switch accounting batch. This implementation requires both `host1.o.ipv4Addr` and `host2.o.ipv4Addr`; MAC-only, stale-host, or IPv6-only devices will silently lose the accounting record and still have no minute-level local history. Consider a MAC-preserving fallback, or explicitly document this as an accepted limitation.

[sensor/APCMsgSensor.js:733] **Low:** The traffic guard uses `tx_bytes + rx_bytes > 0` before normalizing missing values.

If one side is undefined, `undefined + value` becomes `NaN`, and valid one-way accounting can be skipped. Normalize first:

```js
const tx = Number(tx_bytes) || 0;
const rx = Number(rx_bytes) || 0;
if (ip1 && ip2 && tx + rx > 0) {
```

### 💡 Suggestions

- Add regression coverage for switch ACL accounting:
  - one-way bytes still generate a local flow
  - both peer perspectives have correct upload/download
  - generated flows have `fd: "lo"`
  - generated flows have no port/protocol
- In `FlowAggregationSensor.processEnrichedFlow`, consider explicitly requiring `local` when `fd === "lo"` so malformed non-local events do not enter this local-flow aggregation path.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9278  
Head SHA: `e59691be81db688020356581c2dd54a32f0140c8`  
Checked at: 2026-07-02 13:22:30 CST

MelvinTo (approved):

PR #9279, opened 2026-07-01
cache gid and harden ui.conf self-heal with atomic write

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Centralizing `gid` caching inside `api/lib/getGid.js` directly addresses firecommit #8567: after the first successful Redis fallback, repeated local API calls no longer re-read `ui.conf`, retry the failing write, or flood logs.
- Removing the dashboard-level `gid` cache is a good cleanup and satisfies firecommit #8570’s consistency goal: callers now rely on one shared `getGid()` cache.
- The atomic write pattern in both `getGid.js` and `sys/kickstart.js` is a meaningful hardening improvement over truncate-in-place writes.
- Requirement coverage:
  - #8567 log/write amplification: **Met**
  - #8570 unified caller caching: **Met**
  - #8570 atomic self-heal and pairing-time writes: **Met**

### ⚠️ Issues found

No blocking correctness, security, or runtime issues found in the reviewed diff.

One minor robustness concern: `writeUiConfAtomic()` uses a fixed temp path, `/home/pi/.firewalla/ui.conf.tmp`. If several cold-start requests call `getGid()` concurrently before `cachedGid` is populated and `ui.conf` is missing/corrupt, they can race on the same temp file. Functional impact is low because all writers use the same `gid` and failures are caught, but it can produce a spurious self-heal failure log during a burst.

### 💡 Suggestions

- Consider caching the in-flight `getGid()` promise, or using a unique temp path such as `${UI_CONF_PATH}.${process.pid}.${Date.now()}.tmp`, to avoid same-temp-path races.
- The atomic write comments say a crash/power loss “can never” leave a truncated `ui.conf`. The rename avoids torn target writes, but without fsyncing the parent directory, the durability guarantee is not absolute across power loss. Consider softening the wording or fsyncing the directory if that level of guarantee matters.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9279  
Head SHA: `efda77ce3d2cc636183cbaafa10c2184f972ef0c`  
Checked at: 2026-07-01 19:37:33 CST

jasonlyc (approved):

PR #9289, opened 2026-07-03
deduplicate local flows from box and ap by comparing stp port on two hosts

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The core `_dropLocalSameNetworkFlow()` policy is simple and well-tested for same, different, and unknown `stpPort` cases.
- The late BroDetect gate re-checks after host resolution, which is the right place to make the authoritative decision.
- `bridge monitor fdb` uses `spawn()` with argv, so it avoids shell injection risks.
- No linked issue found in PR body or GitHub `closingIssuesReferences`.

### ⚠️ Issues found

- **Medium: identity same-network local flows can bypass the previous drop behavior.**  
  `net2/BroDetect.js:1271` adds `!isIdentityIntf && !isDstIdentityIntf` around the authoritative same-network gate. The old code dropped non-bridge same-network flows regardless of identity state. Now, if either side is an identity interface and the early fast-path could not resolve both `stpPort`s, the flow is kept. Since `_dropLocalSameNetworkFlow(false, null, null)` already preserves the old behavior, this exclusion looks unnecessary and can reintroduce duplicate local-flow accounting for VPN/identity same-interface cases.

- **Low: bridge monitor restart leaves the readline interface unmanaged.**  
  `sensor/DeviceSTPSensor.js:71` creates a `readline.Interface`, but the restart path at `sensor/DeviceSTPSensor.js:78` does not close it or remove listeners. If `bridge monitor fdb` repeatedly exits or errors, old interfaces/listeners may linger until GC. Close the reader in `restart()` and consider consuming/logging `stderr` as well.

### 💡 Suggestions

- Add an integration-style BroDetect test for same-network local flow with unknown `stpPort` after host resolution, including an identity/non-Host case if practical.
- Consider validating MAC format in `processFdbLine()` before calling `getHostFastByMAC()`, mainly to ignore unexpected monitor output cleanly.
- Consider adding a small unit test for `DeviceSTPSensor.processFdbLine()` covering monitored bridge, non-monitored bridge, permanent entries, and VLAN suffix stripping.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9289  
Head SHA: `631e8f316b7d0396d4b5257a88b52fee46cce650`  
Checked at: 2026-07-03 14:10:36 CST

MelvinTo (approved):

PR #9290, opened 2026-07-03
Refresh ept token before/on expiry so checkin self-heals

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly fixes the original async bug in `rrWithEptRelogin()` by awaiting `rrWithErrHandling()`, allowing local `catch` to handle 401s.
- Adds both proactive refresh before JWT expiry and reactive self-healing on checkin 401/403.
- Persists refreshed `eid`/`token` back to `sys:ept`, which addresses the root issue from `firewalla/firecommit#8501`.
- Keeps FireKick as the primary writer while documenting the new runtime refresh writers.
- Linked issue coverage:
  - Met: long-uptime boxes can refresh without reboot/FireKick restart.
  - Met: near-expiry JWTs are proactively refreshed.
  - Met: rejected checkin can refresh and retry once.
  - Met: refreshed token is persisted to Redis for Bone/checkin reuse.

### ⚠️ Issues found

No blocking correctness issues found in the supplied diff.

One minor risk: concurrent/manual checkins could trigger duplicate `refreshToken()` calls if they happen while the token is near expiry or rejected. That should be harmless if `eptLogin` is idempotent for the existing endpoint identity, but it may create noisy logs or extra cloud calls.

### 💡 Suggestions

- Add a small test or manual validation note for:
  - JWT exp within refresh window refreshes before checkin.
  - malformed/expired token causes checkin 401/403, refreshes, calls `Bone.checkCloud()`, then retries successfully.
  - missing `gid` fails gracefully and original checkin path still reports the real failure.
- Consider guarding refresh with a short in-process promise lock if duplicate checkin triggers are common.
- Consider validating `this.config.eptTokenRefreshAdvance` as a number; `|| DEFAULT` is fine for current usage but prevents intentional `0`.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9290  
Head SHA: `5430db6089768f8da9a091e705a1ce6912c33189`  
Checked at: 2026-07-03 16:06:27 CST

jasonlyc (approved):

PR #9291, opened 2026-07-03
Fix to filter out fwapc/dap get

xinnige:
https://github.com/firewalla/firecommit/issues/8584

### DIFF
--- a/lib/ControllerBot.js
+++ b/lib/ControllerBot.js
@@ -194,6 +194,32 @@ module.exports = class {
         });
     }
 
+    ignoreRecordTracelog(data) {
+      if (data && (data.item === "fwapc" || data.item === "dap") && data.value) {
+        const method = (data.value.method || "GET").toUpperCase();
+        if (method === "GET") {
+          // no need to record fwapc/dap GET/NoMethod
+          return true;
+        }
+
+        // read-only style POSTs (queries/validation/ping) that do not change config
+        const path = data.value.path ? String(data.value.path).split("?")[0] : "";
+        const skipPostPaths = [
+          /^\/?(?:v1\/)?station_history\/[^/]+\/stats$/,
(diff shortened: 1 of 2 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The implementation matches linked issue `firewalla/firecommit#8584`: fwapc/dap GET or missing-method requests are skipped, and the listed read-only POST endpoints are filtered.
- Good coverage for fwapc and dap, lower-case methods, missing method, query strings, optional `/v1` prefix, and negative config-changing POST cases.
- The helper keeps the existing `recordTracelog` flow simple and reduces risk compared with adding more inline conditions.

### ⚠️ Issues found

No blocking issues found.

Issue requirement coverage:
- Filter `item: fwapc` / `dap`: Met.
- Filter `value.method` missing or `GET`: Met.
- Filter listed read-only POST APIs: Met.
- Keep config-changing POSTs recorded: Met, with targeted negative tests.

### 💡 Suggestions

- Consider moving `skipPostPaths` to a module-level constant or static class field so the regex array is not rebuilt on every tracelog call. This is not urgent, but `recordTracelog` can be hot-path-ish.
- Consider being explicit about missing method normalization:
  - Current code treats any falsy method (`""`, `0`, `false`, `null`, `undefined`) as `GET`.
  - That is mostly compatible with the old behavior, but if malformed clients send an empty string for a mutating action, it would now be filtered for both fwapc and dap.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9291  
Head SHA: `5a55ea71e33c41cd3ecc772ea24d1507cd8d5d77`  
Checked at: 2026-07-03 17:25:20 CST

jasonlyc (approved):

PR #9292, opened 2026-07-06
check and disable udp_tls module when kernel crash

caixingcheng:
1. check and disable udp_tls module when kernel crash
2. update tls modules


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds a dedicated `KernelCrashMonitor` module instead of spreading pstore/Redis logic through startup and platform code.
- Archives pstore contents before clearing them, and includes archived pstore logs in techsupport.
- Covers the main monitor behavior with focused unit tests: no pstore files, panic detection, duplicate crash handling, version-change re-enable, Redis failure, and archive cleanup.
- No linked issue found in the PR body/title metadata available from GitHub.

### ⚠️ Issues found

**High**

[platform/platform.sh:242]  
The shell implementation permanently skips `xt_udp_tls` whenever Redis has `shouldDisableUdpTls=true`, but it does not mirror the JS re-enable logic that compares the current module version against `udpModuleVersion`. If `platform.sh installTLSModule xt_udp_tls` is used after an upgrade before `net2/main.js` runs the monitor, the module can remain disabled even though the version changed and should be retried. Please add the same version/srcversion comparison here, or route this path through the shared monitor logic before returning.

**Medium**

[net2/KernelCrashMonitor.js:130]  
`isUdpTlsCrash` searches `Modules linked in:.*xt_udp_tls` across every dmesg file in pstore once any file contains `Kernel panic`. That can classify an unrelated panic as UDP-TLS-related if another pstore fragment/archive residue contains `xt_udp_tls`. Since this disables module loading, the match should be scoped to the same crash set as the panic, or grouped by pstore record/timestamp before deciding.

**Low**

[net2/KernelCrashMonitor.js:25]  
`getModuleVersion()` builds a shell command with `koPathOrName` directly. The current callers pass mostly internal values, but this is still brittle in systems code. Prefer `execFile`/argument-based execution, or at least shell-quote the argument consistently.

### 💡 Suggestions

- Add a regression test for the shell path or an integration-style test that proves version-change re-enable works when `platform.sh` is the caller.
- Consider persisting `monitorStartedAt` even when no crash info changed, or remove it if it is not meant to be durable; currently it is set only in memory unless another update triggers `saveCrashInfo()`.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9292  
Head SHA: `6a62ac72e6b0dcb9a638ac9a3f8bd263d39bb9cb`  
Checked at: 2026-07-06 13:16:11 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds a focused crash monitor for `xt_udp_tls` with Redis-backed state and startup checks before module loading.
- Archives pstore contents before clearing them, which preserves useful techsupport evidence.
- Adds techsupport collection for archived pstore logs.
- Covers several important scenarios in the new unit tests: panic vs non-panic, UDP TLS related crashes, version changes, archive cleanup, and Redis failure handling.
- No linked issue found in the PR metadata or body.

### ⚠️ Issues found

[net2/KernelCrashMonitor.js:136]  
Medium: `quotedPaths` is built by interpolating filesystem paths directly into shell commands. Double quotes do not fully protect against shell interpretation such as command substitution, and this pattern is repeated in `grep`/`cat`. Even if pstore filenames are usually kernel-controlled, this is privileged code running through `sudo`; use `execFile`/`spawn` with argv, or strictly validate filenames before shelling out.

[net2/KernelCrashMonitor.js:65]  
Medium: archive cleanup removes directories based on raw `ls` output and interpolates each entry into `sudo rm -rf`. This can remove unexpected paths if a malformed archive name appears. Since archives are expected to be timestamp directories, filter entries with something like `/^\d+$/` before deleting.

[test/test_KernelCrashMonitor.js:25]  
High: the Redis fake does not implement the new lock behavior. `acquireLock()` expects `setAsync(..., 'NX', 'EX', ttl)` to return `'OK'`, but this stub returns `undefined`, so most `checkPstoreAndUpdateRedis()` tests will skip the scan entirely and fail their archive/crash expectations. Please update the fake Redis to emulate `SET NX EX` and `evalAsync` lock release.

[api/bin/www:62]  
Low: `checkPstore()` is fire-and-forget. Most internal errors are caught, but a require-time failure or unexpected rejection would become an unhandled promise rejection. Prefer `checkPstore().catch(err => log.error(...))` so FireApi startup remains explicit and observable.

### 💡 Suggestions

- Consider extracting shell execution helpers for pstore scanning/archive operations so command construction is centralized and easier to harden.
- Add one test for the FireApi/FireMain double-start case to verify the Redis lock prevents duplicate archive/delete behavior.
- Add a regression test for malformed archive directory names to make sure cleanup never deletes anything outside the expected timestamp archives.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9292  
Head SHA: `19ed2cf34e769e4c89af64a6d60fecbac2657317`  
Checked at: 2026-07-07 10:42:15 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The feature direction is reasonable: detect pstore kernel panic logs, persist crash state in Redis, and skip loading `xt_udp_tls` after a suspected crash.
- The Redis lock is the right shape for coordinating FireMain and FireApi startup.
- Archiving pstore before clearing it is important for support/debugging.
- Platform JS and shell module-install paths were both updated, which avoids split-brain behavior.
- No linked issue found in the PR title/body/closing references.

### ⚠️ Issues found
[test/test_KernelCrashMonitor.js:34]  
`FakeRedis.setAsync()` does not emulate Redis `SET NX EX`; it returns `undefined`, so `acquireLock()` always returns false. That means every `checkPstoreAndUpdateRedis()` test exits early with “Another process is already checking pstore” and the crash-detection assertions will fail. Please make the fake return `'OK'` for successful lock acquisition and reject/return null for locked cases.

[test/test_KernelCrashMonitor.js:223]  
This expectation conflicts with the implementation: `checkPstoreAndUpdateRedis()` writes `monitorStartedAt` even when no dmesg files exist. Once the lock stub is fixed, this test will no longer return `{}`. Either remove the `monitorStartedAt` write for no-crash scans, or update the tests to assert the intended metadata.

[api/bin/www:62]  
`checkPstore()` is started fire-and-forget before `run0()`. The first app init can read stale `kernelCrashInfo` before the pstore scan finishes, and setup failures outside `checkPstoreAndUpdateRedis()` can become unhandled rejections. Since `HostManager` now exposes this value during init, this should be awaited or explicitly caught/logged before continuing.

### 💡 Suggestions
[net2/KernelCrashMonitor.js:56]  
Prefer `execFile`/argument arrays for `modinfo` and other commands that interpolate paths. Most inputs here are internal, but pstore/archive paths and module paths are still safer and easier to reason about when not shell-expanded.

[net2/KernelCrashMonitor.js:155]  
`isUdpTlsCrash` checks all pstore dmesg files, not just files associated with the selected panic timestamp. If pstore ever contains mixed crash fragments, this can attribute an unrelated panic to `xt_udp_tls`. Consider narrowing the module scan to the panic files or grouping by pstore record/crash timestamp.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9292  
Head SHA: `0ee06c1230a84cd0686e07ee2a6ba5a7c0445d4e`  
Checked at: 2026-07-07 12:53:42 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds a focused crash monitor around `xt_udp_tls` and uses Redis to persist disable/version state across restarts.
- Uses `execFile` in Node instead of shell interpolation for most command execution.
- Adds unit coverage for the new monitor behavior, including Redis failures, repeated crashes, version changes, and archive failures.
- Startup ordering in `net2/main.js` checks pstore before normal TLS module loading, which matches the intended mitigation path.

### ⚠️ Issues found

**Medium** `[net2/KernelCrashMonitor.js:202]`  
UDP TLS crash attribution can produce false positives. After finding any `Kernel panic`, the code concatenates **all** discovered `dmesg-*` files and checks for `Modules linked in:.*xt_udp_tls`. If pstore contains multiple crash records or unrelated dmesg fragments, one file can provide the panic while another unrelated file mentions `xt_udp_tls`, causing UDP TLS to be disabled incorrectly. Please correlate module matching to the same panic record/group, or only scan files belonging to the newest panic set.

**Medium** `[net2/KernelCrashMonitor.js:92]`  
Archive cleanup trusts directory names from `/log/system/pstore` and passes them into `sudo rm -rf ${PSTORE_ARCHIVE_PATH}/${dir}`. `execFile` avoids shell injection, but a malicious or accidental entry like `../...` could still expand the deletion outside the archive directory. Since this runs with sudo, validate archive names as expected timestamps before deleting, or resolve and verify the path stays under `PSTORE_ARCHIVE_PATH`.

**Low** `[net2/HostManager.js:1778]`  
`kernelCrashInfoForInit()` lets a Redis read failure reject the full init data build. Kernel crash metadata is diagnostic/status information, so a transient Redis parse/read issue should probably degrade to `{}` or `null` instead of breaking the app init response.

### 💡 Suggestions

- Add a test case where one pstore file has `Kernel panic` and another separate file has `Modules linked in: xt_udp_tls`; this would lock down the intended correlation behavior.
- In `platform/platform.sh`, consider handling missing/invalid `jq` output explicitly so the shell path cannot silently skip the disable gate if Redis contains bad JSON.
- No linked issue found in the PR body/page, so I could not verify against issue acceptance criteria.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9292  
Head SHA: `b443e4c182aa9706b08e609e37375bc347151c47`  
Checked at: 2026-07-07 14:52:56 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds a focused `KernelCrashMonitor` module with Redis-backed state for disabling `xt_udp_tls` after suspected kernel panics.
- Checks pstore early in `net2/main.js` before normal TLS module loading, which matches the mitigation intent.
- Archives pstore contents before clearing them and includes archived pstore logs in techsupport output.
- Uses `execFile` for the Node command paths, avoiding shell interpolation for the privileged pstore operations.
- Adds broad unit coverage for crash detection, duplicate crash handling, version-change re-enable, Redis failures, archive cleanup, and module-load state updates.
- Issue requirement coverage: No linked issue found in the PR title/body/closing references.

### ⚠️ Issues found

**Medium** `[net2/KernelCrashMonitor.js:202]`  
UDP TLS crash attribution can produce false positives. Once any pstore file contains `Kernel panic`, the code concatenates all discovered `dmesg-*` files and checks for `Modules linked in:.*xt_udp_tls`. If pstore contains multiple crash records or unrelated fragments, one file can provide the panic while another unrelated file mentions `xt_udp_tls`, causing the module to be disabled incorrectly. Please correlate the module match to the same panic record/group, or only scan files belonging to the newest panic set.

**Medium** `[platform/platform.sh:242]`  
The shell installer path skips `xt_udp_tls` whenever Redis has `shouldDisableUdpTls=true`, but it does not mirror the JS version-change re-enable logic. If this shell path runs after a module upgrade before the JS monitor has refreshed Redis, `xt_udp_tls` can remain disabled even though the new module version should be retried. Please add the same `version/srcversion` comparison here or ensure this path cannot run before the monitor updates the disable state.

**Low** `[net2/HostManager.js:1778]`  
`kernelCrashInfoForInit()` allows a Redis read failure to reject the whole init data build. This crash metadata is diagnostic/status data, so it should probably degrade to `{}` or `null` and log a warning rather than breaking the app init response.

### 💡 Suggestions

- Add a regression test where one pstore file has `Kernel panic` and a separate unrelated pstore file has `Modules linked in: xt_udp_tls`; that would lock down the intended crash-correlation behavior.
- Add coverage for the `platform.sh` disable/re-enable path, especially the “disabled in Redis, module version changed” case.
- Consider making invalid/missing `jq` handling in `platform.sh` explicit so the shell path does not silently bypass the disable gate on bad Redis JSON or missing tooling.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9292  
Head SHA: `a42d0220db7f25bd4388cde4228fa210c334c066`  
Checked at: 2026-07-07 16:11:35 CST

jasonlyc (approved):

PR #9296, opened 2026-07-06
Remove stale radius certificate when migrate

xinnige:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Fixes the log message for custom cert migration failure.
- Expands stale file cleanup beyond the previous fixed certificate filename list, which matches the PR intent.
- Keeps certificate verification before touching the destination.

### ⚠️ Issues found

- `extension/freeradius/freeradius.js:426`: **High** - `saveCerts()` now deletes all custom cert files whenever `customCerts` is absent. The method explicitly allows payloads with only `certs`, and the command handler passes client data through directly, so a default-cert-only save can unintentionally wipe existing custom certs. Deletion should require an explicit signal, such as `customCerts: null` or a dedicated remove flag, instead of treating field absence as removal.

- `extension/freeradius/freeradius.js:369` and `extension/freeradius/freeradius.js:428`: **Medium** - The new destructive `sudo find ${path} -mindepth 1 -delete` commands interpolate paths directly into shell strings. These paths are probably internal, but this is still fragile for spaces/shell metacharacters and risky because the command is recursive and destructive. Prefer `fs.rm`/`fs.readdir` with path APIs, or at minimum quote/escape arguments through a shared helper.

- `extension/freeradius/freeradius.js:369`: **Low/Medium** - `_saveCerts()` logs and continues if stale deletion fails, then still moves new certs into the target directory. That means the operation can return success while stale files remain, which is the main behavior this PR is trying to fix. If stale cleanup is required for correctness, the failure should make `_saveCerts()` return false before moving.

### 💡 Suggestions

- Add tests for `saveCerts({ certs })`, `saveCerts({ customCerts })`, and an explicit custom-cert removal case. The current test coverage exercises `_saveCerts()` only, so the new `else` branch is not protected.
- Consider creating the target directory before cleanup, then delete children using filesystem APIs instead of `sudo find`.
- Issue requirement coverage: No linked issue found in PR title/body/comments/closing references.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9296  
Head SHA: `dabc91bf3f70e6ba64684de6fa41dfe7791feb40`  
Checked at: 2026-07-06 19:18:21 CST

xinnige:
@j-sallyjin this is intentional, by default remove customCerts

extension/freeradius/freeradius.js:426: High - saveCerts() now deletes all custom cert files whenever customCerts is absent. The method explicitly allows payloads with only certs, and the command handler passes client data through directly, so a default-cert-only save can unintentionally wipe existing custom certs. Deletion should require an explicit signal, such as customCerts: null or a dedicated remove flag, instead of treating field absence as removal.

jasonlyc (approved):

PR #9298, opened 2026-07-06
encipher: per-request random IV via extensible { iv, message } envelope

MelvinTo:
## What

Adds an opt-in **per-request random IV** to the box's netbot request/response crypto (app via `api/lib/Encryption.js`, MSP via `sensor/Guardian.js`, primitives in `encipher/lib/encipherio.js`), fixing the deterministic-ciphertext weakness of the fixed all-zero IV while staying fully backward compatible.

Tracking issue: firewalla/firecommit#8632 (this consolidates the earlier #9300, which was closed and folded in).

## Wire contract (current — please read; the design changed during review)

The IV is **embedded in the message value as a JSON envelope**, not a separate top-level field. This keeps the format extensible (a future auth tag is just another field) and means the box mirrors whatever scheme the request used.

- **Random IV** — the `message` value is a JSON string `{"iv":"<base64>","message":"<base64 ciphertext>"}`:
  ```json
  { "message": "{\"iv\":\"<base64 16B IV>\",\"message\":\"<base64 ciphertext>\"}", "rkeyts": ... }
  ```
- **Legacy** — the `message` value is bare base64 (zero IV), exactly as today:
  ```json
  { "message": "<base64 ciphertext>", "rkeyts": ... }
  ```
- **Reply** mirrors the request: an envelope request gets an envelope reply (fresh IV inside `message`); a legacy request gets a legacy reply. **There is no top-level `iv` field on the request or the reply** — it lives inside the envelope.

Parsing (`encipherio.js` `_parseEnvelope`, mirror on the client):
- raw base64 (not JSON) → legacy, zero IV
- JSON string `"<base64>"` → legacy, zero IV
- object `{ message, iv }` → use `iv`
- object `{ message }` (no iv) → legacy fallback, zero IV
- object without `message` → invalid (rejected)

The base64 alphabet has no `{`/`"`, so a legacy bare-base64 message never collides with the JSON forms.

> Note for reviewers: earlier commits on this branch used a **separate top-level `iv` field**; that was superseded by the embedded envelope (see commit `a5b924a79`). The description, tests, and `scripts/test_encipher_iv.js` all reflect the envelope contract. This is why `Encryption.js` reads the IV from the message value rather than `req.body.iv`, and replies with the envelope inside `message`.

## Scope: no group-level marker (app decides per request)

An earlier iteration added a group `info.iv` capability marker plus `enableIv`/`disableIv`/`getIv` netbot cmds. **That was removed** — with the envelope, the app simply chooses per request whether to send it, and the box mirrors it. So this PR intentionally:
- has **no** `enableIv`/`disableIv`/`getIv` cmds, and
- **removes** the `iv:1` stamping from group metadata in `sys/kickstart.js` / `net2/FWCloudWrapper.js`.

How a client decides it's safe to send the envelope (e.g. gating on box firmware version) is the client's choice; an old box that can't parse it returns HTTP 400/412.

## Backward compatibility
- Box-first and zero-risk: the box uses a random IV only in reply to a request that used one. Old app/MSP are untouched (legacy bare base64, zero IV).
- Out of scope (still zero IV): streaming responses, group broadcast messages, file uploads (`encryptBinary`).
- Still AES-256-CBC (no authentication); AEAD/HMAC is a tracked follow-up.

## Review hardening already applied
- **Strict IV validation** (`_normalizeIV`): canonical base64 of exactly 16 bytes + byte-exact round-trip; malformed IVs → decrypt error → 412.
- **MSP error frames** never carry an IV.
- **Single parse:** `decryptRequest` parses the envelope once and returns `usedIv`, so the reply mirrors it without a second `_parseEnvelope`.

## Tests
`test/test_encipher_iv.js` covers `_normalizeIV`, `_parseEnvelope` (all forms), and `encrypt`/`decrypt` round-trips (legacy bare/JSON-string/`{message}`-no-iv, envelope, non-determinism, invalid→null, bad-iv→null). Verified against the real class on a box (all assertions pass) and end-to-end on a testbed via `scripts/test_encipher_iv.js` (`--iv` sends the envelope; default legacy).

## Files
`encipher/lib/encipherio.js`, `api/lib/Encryption.js`, `sensor/Guardian.js`, `sys/kickstart.js`, `net2/FWCloudWrapper.js`, `scripts/test_encipher_iv.js`, `test/test_encipher_iv.js`.


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Backward compatibility is preserved: absence of `iv` keeps the legacy zero-IV behavior.
- The opt-in negotiation model is simple and stateless.
- IV validation enforces the AES-CBC 16-byte requirement before use.
- API and MSP paths both generate a fresh response IV only when the request provided one.
- Streaming paths intentionally remain legacy-compatible.

### ⚠️ Issues found
No blocking correctness issues found.

One minor concern:
- `encipher/lib/encipherio.js:_normalizeIV` relies on Node’s permissive base64 decoder. Invalid base64 characters can be ignored by `Buffer.from(iv, 'base64')`. Since IVs are not secret this is not severe, but stricter validation would make malformed client input behavior clearer.

### 💡 Suggestions
- Add focused unit tests around `_normalizeIV`, `decryptRequest`, and `encryptResponse`, especially malformed base64, wrong IV length, legacy zero-IV compatibility, and random-IV response behavior.
- Consider documenting the streaming exception clearly for app/MSP clients: request IV normally means response IV, except streaming responses remain zero-IV and omit `iv`.
- Longer term, AES-CBC still lacks message authentication. Random IV fixes deterministic ciphertext leakage, but authenticated encryption or an HMAC remains the stronger protocol direction.

### Verdict
APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9298  
Head SHA: `61616c7711bb595a267c8cf992ffdd038e22c596`  
Checked at: 2026-07-06 19:42:13 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Backward-compatible negotiation is clean: no `iv` keeps legacy zero-IV behavior, while opt-in clients get random response IVs.
- API and MSP paths are updated consistently for normal request/response traffic.
- `_normalizeIV()` enforces AES-CBC’s 16-byte IV length before crypto use.
- Streaming/group/file paths are intentionally left unchanged and documented as out of scope.
- Reviewed 4 changed files. No linked issue found.

### ⚠️ Issues found
- Low: `[encipher/lib/encipherio.js:724]` `Buffer.from(iv, 'base64')` is permissive and can ignore invalid base64 characters. Since this is protocol input, consider strict base64 validation before accepting the IV.
- Low: `[sensor/Guardian.js:726]` MSP error responses can include an `iv` even when `encryptedResponse` was never produced. If clients treat `iv` as “decrypt `message` with this IV,” malformed/decrypt-error requests may cause confusing client behavior.

### 💡 Suggestions
- Add focused tests for malformed base64 IVs, wrong IV length, legacy no-IV compatibility, and negotiated random-IV replies.
- In `Guardian.js`, only include `iv` when `code === 200` and `encryptedResponse` exists, unless the MSP protocol explicitly expects IVs on error frames.
- Longer term, random IV fixes deterministic ciphertext leakage, but AES-CBC still lacks authentication; adding HMAC or moving to AEAD should remain a follow-up.

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9298  
Head SHA: `eab3e7fbd5b7da7845f708cd1ed36ab7b7c213c6`  
Checked at: 2026-07-07 11:04:31 CST

MelvinTo:
Addressed the review feedback in `edaefea8d`:

**1. `_normalizeIV` permissive base64 (both review rounds).** Node's `Buffer.from(iv,'base64')` silently ignores invalid characters. Now strictly validated: a string IV must match canonical base64 of exactly 16 bytes (`/^[A-Za-z0-9+/]{22}==$/`) **and** round-trip byte-exactly (`ivBuf.toString('base64') === iv`); Buffers must be 16 bytes. Malformed client IVs are rejected up front (on the request path this surfaces as `decrypt` → null → HTTP 412) instead of being quietly reinterpreted.

**2. MSP error frames carrying an `iv` (`sensor/Guardian.js`).** The `iv` is now included only when the response actually succeeded and produced ciphertext — `replyIVBuf && code === 200 && encryptedResponse`. Error frames (412/500, no `encryptedResponse`) no longer advertise an `iv` the client would try to decrypt with.

**3. Focused unit tests.** Added `test/test_encipher_iv.js` (mocha/chai) covering `_normalizeIV` (null, 16-byte Buffer, valid base64, wrong-length Buffer, wrong-length/malformed/non-canonical base64) and `encrypt`/`decrypt` round-trips (legacy zero-IV, random-IV Buffer-encrypt→base64-decrypt, random-IV non-determinism, zero-IV determinism, random-IV-not-readable-under-zero-IV, malformed-IV → `decrypt` returns null, `encrypt` throws).

Verification: mocha/chai aren't installed on the target box (production deps only), so the suite runs in CI. To confirm the assertions hold against the real class, I ran the equivalent checks against the branch's `encipherio.js` on a box — **all 15 assertions passed** (`_normalizeIV` ×8, encrypt/decrypt ×7).

**Streaming doc (suggestion).** The out-of-scope note is documented in the app/MSP integration guide (firecommit#8632): a request IV normally yields a response IV, except streaming responses stay zero-IV and omit `iv`.

**AEAD/HMAC (suggestion).** Acknowledged as the stronger long-term direction; tracked as a separate follow-up beyond this PR (random IV here only fixes deterministic-ciphertext leakage; CBC still lacks authentication).

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds opt-in per-request IV handling while preserving legacy zero-IV behavior when `iv` is absent.
- Validates IV shape strictly instead of relying on Node’s permissive base64 decoder.
- Keeps streaming and broader group/file encryption paths out of scope, which is a reasonable compatibility boundary.
- Adds useful unit coverage for IV normalization, legacy round-trip, random-IV round-trip, and malformed-IV handling.
- Guardian error responses correctly avoid sending an `iv` when no encrypted response was produced.

### ⚠️ Issues found

- **Medium: unauthenticated IV makes first-block CBC malleability explicit.**  
  The PR sends the IV in clear, which is normal only when ciphertext integrity is enforced. Since this protocol still uses AES-CBC without MAC/GCM, any party able to modify the request can now flip bits in the IV to predictably alter the first plaintext block before JSON parsing. CBC ciphertext was already unauthenticated, but exposing caller-controlled IV expands the tamper surface to block 0. This should be called out as a security limitation and ideally paired with an HMAC or AEAD follow-up before clients broadly rely on this as a security improvement.

  `[encipher/lib/encipherio.js:786]`  
  `decryptRequest()` accepts the caller-provided IV and decrypts without authenticating either the IV or ciphertext. Consider adding integrity protection, or at minimum rejecting random-IV mode unless the request also carries a verified MAC over `{iv, message}`.

### 💡 Suggestions

- Add integration-style tests for `decryptRequest()` / `encryptResponse()` with a stubbed `getKey()`, not only direct `encrypt()` / `decrypt()`. That would cover the exact API middleware and Guardian code paths introduced here.
- Consider documenting accepted IV encoding as “canonical padded standard base64 only” because `_normalizeIV()` intentionally rejects base64url and unpadded forms. That is fine, but clients need to match it exactly.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9298  
Head SHA: `edaefea8d9914e92ed63aeb2000a02b02b10164c`  
Checked at: 2026-07-07 11:32:56 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The opt-in IV negotiation preserves legacy zero-IV behavior for old clients.
- API and MSP paths are consistent: negotiated requests get fresh response IVs, streaming remains legacy.
- IV validation is now strict and covered by focused unit tests.
- MSP error responses correctly avoid advertising an IV when no encrypted response exists.

### ⚠️ Issues found

- **Medium:** `encipher/lib/encipherio.js:399` silently treats undecryptable or malformed existing `group.info` as `{}`. In `upgradeGroupInfoIV()`, that object is then written back with only `iv`, which can erase existing group metadata if `group.info` decrypts to `null` or parses badly. Please fail closed here: if `group.info` exists but cannot decrypt or parse to a JSON object, throw and avoid `_writeGroupInfo()`.

### 💡 Suggestions

- Add a unit test for `_getGroupInfoForUpdate()` / `upgradeGroupInfoIV()` where existing `group.info` is malformed or decrypts to `null`, verifying it does not overwrite metadata.
- Document that this PR fixes deterministic ciphertext only; CBC remains unauthenticated, and caller-supplied IV mode should eventually be paired with HMAC or AEAD.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9298  
Head SHA: `dbef081c88e1e6a1f36b0bde7d872dfb21dbdfb4`  
Checked at: 2026-07-07 11:55:16 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Backward compatibility is preserved: no `iv` keeps legacy zero-IV request/response behavior.
- API and MSP paths consistently generate a fresh response IV only when the request provided one.
- IV validation is now strict and covered by focused unit tests.
- Streaming/group/file paths are intentionally left out of scope and remain legacy-compatible.

### ⚠️ Issues found
- **Medium / REQUEST_CHANGES** `[encipher/lib/encipherio.js:397]`  
  `_getGroupInfoForUpdate()` silently converts an existing but undecryptable, malformed, or non-object `group.info` into `{}`. `enableGroupInfoIV()` then writes that object back with only `iv`, which can erase existing group metadata. If `_parseJsonSafe()` returns a primitive or array, the update can also behave incorrectly or fail unexpectedly. Since these commands mutate persisted group info, this should fail closed when `group.info` exists but cannot decrypt/parse to a plain object.

### 💡 Suggestions
- Add tests for `_getGroupInfoForUpdate()` / `enableGroupInfoIV()` where `group.info` decrypts to invalid JSON, `null`, an array, and a primitive string, verifying no update is posted.
- Consider documenting that accepted IV strings must be canonical padded standard base64, not base64url or unpadded base64.
- Longer term, CBC remains unauthenticated; random IV fixes deterministic ciphertext leakage but should eventually be paired with HMAC or AEAD.

### Verdict
REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9298  
Head SHA: `44d2e4e9153665eb2a777d09ef4c0f9148328404`  
Checked at: 2026-07-07 12:14:24 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The request/response IV negotiation is backward compatible: no request `iv` keeps legacy zero-IV behavior, while clients that send `iv` get a fresh response IV.
- IV validation is strict enough to avoid Node’s permissive base64 decoding surprises.
- Streaming and broadcast paths are intentionally left on legacy behavior, which matches the stated scope.
- Unit tests cover the core crypto behavior, malformed IV handling, and legacy determinism.

Issue requirement coverage: no auto-closing linked issue found. The PR references `firecommit#8590`, but that issue is about key generation entropy and explicitly lists fixed zero-IV CBC as out of scope / separate follow-up. This PR addresses that follow-up scope, not the keygen issue itself.

### ⚠️ Issues found

**Medium** `[encipher/lib/encipherio.js:394]`  
`_getGroupInfoForUpdate()` treats an existing but undecryptable or unparsable `group.info` as `{}`. For `enableGroupInfoIV()`, that can overwrite all existing group info with only `{ iv: ... }`, losing fields like `type`, `member`, or `model` if decrypt/parse fails due to key mismatch, format drift, or corruption. If `group.info` exists, decrypt/parse failure should reject instead of silently defaulting to an empty object.

**Medium** `[controllers/netbot.js:2292]`  
`const ivVersion = (value && value.iv) || 1` accepts arbitrary truthy values. A caller can persist `"1"`, `"foo"`, `{}`, etc., while the test tool and likely client gating check `info.iv === 1`. That can leave the group marked in a way that is reported as enabled but not actually honored. Validate the command input and only allow the supported marker value, or coerce strictly to integer `1`.

### 💡 Suggestions

- Add a unit test for preserving existing `group.info` fields during `enableGroupInfoIV()` and for rejecting malformed existing `group.info`.
- Add command-level tests for `enableIv` with invalid `value.iv` inputs.
- Consider documenting that this only fixes CBC equality leakage; it still does not authenticate ciphertext.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9298  
Head SHA: `9b07edbe27c135bf61f6ff6453086ad74b36ab31`  
Checked at: 2026-07-07 12:35:48 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Backward compatibility is mostly preserved: legacy bare ciphertext still decrypts with the zero IV, and responses only use a fresh IV when the request envelope carries one.
- `_normalizeIV()` now strictly validates canonical 16-byte base64 IVs.
- The API and MSP paths consistently moved to the embedded `{ iv, message }` envelope model.
- Unit coverage for IV parsing, malformed IVs, legacy behavior, and random-IV round trips is useful.
- Issue coverage: PR partially addresses `firecommit#8632` by adding opt-in random IV for app/MSP request/response paths, while leaving streaming, group broadcast, file encryption, and authentication out of scope as expected.

### ⚠️ Issues found

- **Medium** `[encipher/lib/encipherio.js:397]`  
  `_getGroupInfoForUpdate()` silently converts an existing but undecryptable, malformed, or non-object `group.info` into `{}`. `enableGroupInfoIV()` then writes that back with only `iv`, which can erase existing group metadata. If `group.info` exists, decrypt/parse failure should fail closed and avoid `_writeGroupInfo()`.

- **Medium** `[controllers/netbot.js:2294]`  
  `const ivVersion = (value && value.iv) || 1` accepts arbitrary truthy values like `"1"`, `"foo"`, or `{}`. The client/test-tool gating checks `info.iv === 1`, so this can persist a marker that looks enabled in group info but is not honored by strict clients. Validate or coerce this to the supported integer marker.

- **Medium** `[api/lib/Encryption.js:116]`  
  The implementation now embeds IV inside `message`, but `firecommit#8632` still documents top-level request/reply `iv` fields for app/MSP integration. That mismatch can break client rollout if clients implement the linked issue guidance. Update the issue/client guide or keep the documented wire contract.

### 💡 Suggestions

- Add tests for `_getGroupInfoForUpdate()` with malformed JSON, decrypt failure, arrays, `null`, and primitive JSON values, verifying no update is posted.
- Add command tests for invalid `enableIv` inputs.
- Document the final wire format clearly: negotiated messages are JSON envelope strings inside the outer `message` field.
- Longer term, CBC remains unauthenticated; random IV fixes equality leakage but should eventually be paired with HMAC or AEAD.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9298  
Head SHA: `a5b924a799c6d1dbb532ddc2154895f17b2a3b48`  
Checked at: 2026-07-07 18:39:24 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds focused crypto unit coverage for legacy zero-IV behavior, random-IV envelopes, malformed IV handling, and backward-compatible decrypt paths.
- Keeps legacy callers working when no IV is provided.
- Uses `crypto.randomBytes(16)` for negotiated replies, which is the right primitive for CBC IV generation.
- Avoids applying the new IV response path to streaming responses, which preserves the current SSE behavior.

### ⚠️ Issues found

- **High: Request/response IV wire format does not match the PR contract.**  
  `api/lib/Encryption.js:57` now only reads the IV if it is embedded inside `req.body.message` as a JSON envelope. The PR body says clients send `{ message, iv }` with `iv` as a top-level field, and responses return `{ message, iv }`. A client following that contract will have its IV ignored and the ciphertext decrypted with the zero IV, causing 412/decrypt failures.

- **High: Response IV is hidden inside `message`, not returned as a sibling field.**  
  `api/lib/Encryption.js:103` returns `{ message: encryptedResponse }`, where `encryptedResponse` may itself be a JSON string containing `{ iv, message }`. That changes the expected API shape from top-level `{ message, iv }` to nested/escaped JSON. App/MSP callers expecting `reply.iv` will not detect the IV.

- **Medium: Group IV marker scope is inconsistent with the PR title/body.**  
  `net2/FWCloudWrapper.js:53` and `sys/kickstart.js:194` remove `iv: 1` from group metadata, while the PR title/body mentions “group info iv marker cmds” and says new groups get the marker. The diff also does not include `upgradeIv` / `downgradeIv` command implementation. Either the PR description is stale, or the implementation is missing part of the intended behavior.

- **Medium: Tests lock in the nested envelope behavior but do not cover the stated top-level API contract.**  
  `test/test_encipher_iv.js` validates `message = JSON.stringify({ iv, message })`, but there is no coverage for `{ message: ciphertext, iv }` on request or `{ message, iv }` on response. That means the current tests would pass while the documented app/MSP integration fails.

- **Issue requirement coverage:** No linked issue found. The PR references closed PR `#9300`, not a linked issue.

### 💡 Suggestions

- Decide on one wire contract and make code, tests, script, and PR body agree. If the intended contract is top-level `{ message, iv }`, pass `req.body.iv` into decrypt and return `res.json({ message: ciphertext, iv })`.
- If the intended contract is the nested message envelope, update the PR body/title and app/MSP integration notes explicitly, because this is a breaking expectation change from the current description.
- Either restore/add the group info marker command changes, or remove that scope from the PR title/body.
- Add route-level tests around `api/lib/Encryption.js` and `sensor/Guardian.js` to verify legacy request, IV request, malformed IV, and reply shape.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9298  
Head SHA: `2ef989fdadde19588feb0fbb19ed8fcf8d439445`  
Checked at: 2026-07-07 18:56:54 CST

MelvinTo:
Thanks — every flagged item was **stale PR description**, not a code defect: the design moved from a separate top-level `iv` field to an embedded `{ iv, message }` envelope during review, and the body/title hadn't caught up. Updated title + description to the current contract. Point by point:

1. **"Wire format doesn't match contract" / "response IV hidden inside message"** — intended. The contract is now the **embedded envelope**: request `message` is `{"iv","message"}` (or bare base64 for legacy), and the reply mirrors it — no top-level `iv` on request or reply. `Encryption.js` reading the IV from the message value and returning the envelope inside `message` is correct for this contract. Description rewritten to state this explicitly.

2. **"Group IV marker scope inconsistent; upgradeIv/downgradeIv missing"** — intended. The `info.iv` marker and `enableIv`/`disableIv`/`getIv` cmds were **removed**; the `iv:1` stamping revert in `kickstart.js`/`FWCloudWrapper.js` is deliberate. With the envelope the app chooses per request, so there's no group marker to manage. Removed that scope from the title/body.

3. **"Tests lock nested envelope but not top-level contract"** — the nested envelope **is** the contract, so the tests are correct. Added `_parseEnvelope` coverage for all forms (raw base64, JSON-string, `{message,iv}`, `{message}`-no-iv, invalid). A top-level-`iv` test would assert a contract that no longer exists.

4. **"No linked issue; references closed #9300"** — linked the tracking issue firewalla/firecommit#8632 in the body; #9300 was closed and folded into this PR.

Also already addressed from the prior round: strict `_normalizeIV` base64 validation, MSP error frames never carrying an IV, and the envelope is now parsed once per request (`decryptRequest` returns `usedIv`).

Head is `2ef989fda`. Happy to add route-level tests around `Encryption.js`/`Guardian.js` if you'd like belt-and-suspenders coverage of the legacy/envelope/malformed request shapes.

jasonlyc (approved):

PR #9299, opened 2026-07-06
sync purple 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- `api/routes/sensor.js` adds a strict sensor-name whitelist before dynamic `require`, which closes an obvious traversal/injection path.
- `api/lib/getGid.js` centralizes GID recovery and avoids repeated direct reads of `ui.conf`.
- VPN/common-name validation in `controllers/netbot.js` is improved by reusing `isValidCommonName`.
- PR metadata has no linked issue: No linked issue found.

### ⚠️ Issues found

- **High** `[extension/cloudcache/cloudcache.js:39]`  
  The Redis download lock expires after only 5 seconds, but cloud downloads can easily exceed that. Another process can acquire the same lock while the first download is still running, and `tryAcquireDownloadLock()` even force-deletes the lock after about 5 seconds of retries. This reintroduces the concurrent writer contention the lock is supposed to prevent. Use a TTL longer than the worst expected download time, renew the lease while downloading, or avoid force-deleting unless ownership/staleness can be proven.

- **High** `[control/IpsetControl.js:44]`  
  The spawned command is `sudo ipset - -!`, but the existing bulk path uses `ipset restore -!`. If the intent is a stdin-fed restore process, this should spawn `ipset restore -!` and write restore commands to stdin. As written, the interactive process may exit immediately or reject input, while `_batchWrite()` still reports success.

- **Medium** `[control/IpsetControl.js:51]`  
  `_batchWrite()` does not handle process `exit`/`close`, `stdin` backpressure, or command completion. Callers return as soon as data is written to stdin, so later policy logic can assume ipsets are updated even if the process died, the pipe is full, or ipset rejected a command asynchronously. This path needs lifecycle handling and a fallback/retry that confirms the batch was accepted.

### 💡 Suggestions

- Add an integration/unit test around deferred ipset execution that kills the child process and verifies the next batch is not silently lost.
- Add a cloudcache contention test with a mocked slow download over 5 seconds to prove only one writer can update cache content and metadata.
- Consider making the PR smaller next time; this one spans many unrelated systems, which makes regression review much harder.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9299  
Head SHA: `e1ba78b47eadf54de9985255dfc68f7228cd1e99`  
Checked at: 2026-07-06 20:02:13 CST

PR #9300, opened 2026-07-07
netbot: add upgradeIv / downgradeIv cmds for group info iv marker

MelvinTo:
## What

Adds a netbot cmd (`item: "upgradeIv"`) that back-fills the `iv` marker into an existing group's encrypted `info`, using the cloud update-group API. No-op if the group already has `iv` configured.

Companion to the group `info.iv` marker (firewalla/firewalla#9297) and the per-request IV work (firecommit#8632 / firewalla/firewalla#9298). New groups get `iv` at creation; this command back-fills groups created before that.

## How

`encipherio.upgradeGroupInfoIV(gid, ivVersion = 1)`:
1. `GET {endpoint}/group/{appId}/{gid}` for the raw group record (not `groupFind()`, which rewrites `group.name` to the decrypted value).
2. Decrypt `group.info` with the **base** group symmetric key (`privateDecrypt` of this eid's `symmetricKeys[].key`) — info rides the base key + legacy zero IV, not the rotated rkey.
3. If `info.iv` is already set → return `{ updated: false, iv }` (no-op).
4. Otherwise set `info.iv = ivVersion`, re-encrypt with the base key, and `POST {endpoint}/group/{appId}/{gid}` with `{ name, xname, info }` — `name`/`xname` are passed back unchanged (both are required by the update-group API).

`netbot.cmdHandler`: `case "upgradeIv"` calls it; `ivVersion` can be overridden via `msg.data.value.iv`, defaulting to `1`.

## Verification (testbed, dev cloud)

Ran the equivalent logic against a real group on box 192.168.201.20 (group info had no `iv`):

```
BEFORE info: {"type":"fb","member":"direct","model":"goldpro"} | iv present: false
POSTing update with name/xname unchanged + new info(iv=1)...
AFTER  info: {"type":"fb","member":"direct","model":"goldpro"} | iv present: false
RESULT: cloud returned 200 but DID NOT persist info
```

The client flow (read → detect missing iv → update with name/xname/info) is correct. Persistence did not take effect because the **dev cloud (dv0) does not have the cloud-side update-group fix deployed** — the documented behavior is "server accepts the request and returns 200 but silently drops the info change." So end-to-end persistence is gated on the cloud fix; the box-side code here is complete and correct.

## Notes / open question

- Default `ivVersion` is `1`. Under the current `info.iv` convention this needs to line up with what enables the per-request random IV on the app side — worth confirming the intended value alongside firecommit#8632.
- Info stays on the base key + zero IV (unchanged storage scheme); only the plaintext gains an `iv` field.

---

**Update:** also added `downgradeIv` (cmd item `downgradeIv`) for debugging — removes the `iv` field from the group info via the same update-group API (no-op if unset). The shared GET-group / decrypt-info / write-info logic is factored into `_getGroupInfoForUpdate()` and `_writeGroupInfo()`, used by both `upgradeGroupInfoIV` and `downgradeGroupInfoIV`.

### DIFF
--- a/controllers/netbot.js
+++ b/controllers/netbot.js
@@ -2288,6 +2288,12 @@ class netBot extends ControllerBot {
       case "rekey":
         await this.eptcloud.reKeyForAll(gid);
         return
+      case "upgradeIv": {
+        // Back-fill the "iv" marker into the group info via the cloud
+        // update-group API. No-op if the group already has iv configured.
+        const ivVersion = (value && value.iv) || 1;
+        return await this.eptcloud.upgradeGroupInfoIV(gid, ivVersion);
+      }
       case "syncLegacyKeyToNewKey":
         await this.eptcloud.syncLegacyKeyToNewKey(gid);
         return
--- a/controllers/netbot.js
+++ b/controllers/netbot.js
@@ -2294,6 +2294,10 @@ class netBot extends ControllerBot {
         const ivVersion = (value && value.iv) || 1;
         return await this.eptcloud.upgradeGroupInfoIV(gid, ivVersion);
       }
+      case "downgradeIv": {
+        // Debugging: remove the "iv" marker from the group info. No-op if unset.
+        return await this.eptcloud.downgradeGroupInfoIV(gid);
+      }
       case "syncLegacyKeyToNewKey":
         await this.eptcloud.syncLegacyKeyToNewKey(gid);
         return
--- a/encipher/lib/encipherio.js
+++ b/encipher/lib/encipherio.js
@@ -367,6 +367,64 @@ let legoEptCloud = class {
     return name
   }
 
+  // Back-fill the "iv" marker into the group's encrypted info via the cloud
+  // update-group API. No-op if the group info already carries an "iv" field.
+  // Returns { updated, iv }.
+  async upgradeGroupInfoIV(gid, ivVersion = 1) {
+    if (this.appId === undefined || gid === undefined) {
+      throw new Error("parameter error");
+    }
+
+    // Fetch the raw group record (pristine name/xname/info); do not go through
+    // groupFind() since parseGroup() rewrites group.name to the decrypted value.
+    const resp = await this.rrWithEptRelogin({
+      uri: `${this.endpoint}/group/${this.appId}/${gid}`,
+      family: 4,
+      method: 'GET',
+      json: true,
+      maxAttempts: 2
+    });
+    const group = resp && resp.body;
+    if (!group || _.isString(group) || !_.isArray(group.symmetricKeys)) {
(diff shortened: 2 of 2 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The helper correctly avoids `groupFind()` so it can preserve raw `name`/`xname` values for the cloud update API.
- It uses the base symmetric key for `group.info`, matching the linked `firecommit#8632` guidance.
- `upgradeIv` is idempotent when `info.iv` already exists, and `downgradeIv` is useful for controlled debugging.
- Reviewed 2 changed files and found 3 actionable issues.

### ⚠️ Issues found

**Issue requirement coverage:** The PR is related to `firewalla#9297`, `firewalla#9298`, and `firecommit#8632`. It partially satisfies the backfill requirement for existing groups, but gaps remain around validation and persistence confirmation.

[encipher/lib/encipherio.js:397]  
**High:** If `group.info` exists but fails to decrypt or parse, this code falls back to `{}` and later writes that back with only the new `iv` field. That can silently erase existing group metadata such as `type`, `member`, or `model`. For an update/migration command, malformed or undecryptable existing info should fail closed instead of being replaced.

[controllers/netbot.js:2293]  
**Medium:** `ivVersion` accepts any truthy `value.iv`, including strings, objects, negative numbers, or unsupported versions. The linked app/MSP contract gates random-IV behavior on `info.iv === 1`, so writing `"1"` or another value may leave clients in legacy mode or create inconsistent state. Validate this as the supported numeric value, or explicitly reject unsupported versions.

[encipher/lib/encipherio.js:430]  
**Medium:** `upgradeGroupInfoIV()` returns `{ updated: true }` immediately after the POST, but the PR notes that some cloud deployments return 200 while silently dropping the `info` change. Since this command is specifically a backfill/migration tool, it should re-fetch and verify `info.iv` was persisted before reporting success, or return a distinct “accepted but not verified” result.

### 💡 Suggestions

- Add a focused test for existing `group.info` with valid metadata to ensure upgrade preserves all existing fields.
- Add tests for malformed/decrypt-failing `group.info` to confirm the command does not overwrite metadata.
- Consider documenting whether `downgradeIv` should be production-accessible or guarded as a debug-only command.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9300  
Head SHA: `ff9481c4a25d85eb9c1dee16f6fc985f965553f4`  
Checked at: 2026-07-07 11:31:12 CST

MelvinTo:
Folded into #9298 to keep the IV work as a single PR — the two commits (upgradeIv, downgradeIv) were cherry-picked onto the #9298 branch (commits a39837815, dbef081c8). Closing this in favor of #9298.

PR #9305, opened 2026-07-08
listen link_wan_switched in VirtWanGroup

chgchi:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The implementation matches the linked issue’s intended design: VWG now listens for `link_wan_switched` and resets the affected member’s `connState.ready` to `false` without immediately refreshing routes.
- This works with the existing `processLinkStateEvent` logic: a later `link_established` with `routeUpdated === false` can still trigger `refreshRT()` because `ready` is no longer `true`.
- Listener cleanup was added in `destroyEnv()`, avoiding a stale event listener after the VWG is removed.
- Issue requirement coverage:
  - Met: Handle Dual WAN failover for VPN groups containing Docker-based VPN clients.
  - Met: Do not immediately refresh routes on WAN switch while the VPN client may still be restarting.
  - Mostly met: Allow later `link_established` to rebuild the VWG route table even if the initial startup event was missed.

### ⚠️ Issues found

[net2/VirtWanGroup.js:432]  
Medium: `processWanSwitchedEvent()` updates `this.connState[profileId].ready` in memory but does not persist the changed `connState` back to Redis. `processLinkStateEvent()` explicitly saves `connState` after state transitions, but this new path does not. If firemain restarts after `link_wan_switched` and before the later `link_established`, Redis can still show `ready: true`; after reload, the later periodic `link_established` with `routeUpdated === false` may not trigger `refreshRT()`, which is the exact recovery path this PR depends on. Consider persisting `connState` here the same way `processLinkStateEvent()` does, guarded by the existing VWG Redis key check.

### 💡 Suggestions

- Add a focused unit/regression test for the linked issue flow:
  - initial VWG member is `ready: true`
  - emit `link_wan_switched`
  - verify `ready` becomes `false` and is persisted
  - emit later `link_established` with `routeUpdated: false`
  - verify `refreshRT()` is scheduled/called
- Consider logging at debug/info level when a VWG member is marked not ready due to WAN switch. This would make future route-table recovery investigations easier without changing behavior.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9305  
Head SHA: `a78417ea3b5e239e848bc19e943f03b85639aeb9`  
Checked at: 2026-07-08 19:28:19 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly hooks `link_wan_switched` for `VirtWanGroup` only in the main process.
- Filters events to VWG member `profileId`s before mutating group state.
- Marks the affected member `connState.ready = false` without immediately calling `refreshRT()`, matching the linked issue design.
- Persists updated `connState` to Redis, so recovery still works across process restart.
- Removes the new listener in `destroyEnv()`, avoiding stale event handlers.
- Linked issue coverage:
  - Met: Handle Dual WAN failover for VPN groups containing Docker-based VPN clients.
  - Met: Avoid immediate route refresh while the VPN client may still be restarting.
  - Met: Allow later periodic `link_established` with `routeUpdated === false` to trigger VWG route rebuild because `ready` is no longer true.

### ⚠️ Issues found

No blocking issues found.

### 💡 Suggestions

- Add a focused regression test for the issue flow:
  - VWG member starts with `ready: true`
  - emit `link_wan_switched`
  - verify `ready` becomes `false` and Redis `connState` is updated
  - emit later `link_established` with `routeUpdated: false`
  - verify `refreshRT()` is triggered

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9305  
Head SHA: `1b8e6aa5abc5aa18d706463f836ce750f31e1372`  
Checked at: 2026-07-08 19:46:19 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The PR targets the right recovery path: a later `link_established` can now trigger VWG route refresh even when `connState.ready` was already `true`.
- Initializing VPN connectivity counters to `INTERNET_ON_OFF_THRESHOLD - 1` helps the first post-start connectivity check converge faster.
- The route-table command uses a VWG table name derived from a validated UUID, so command injection risk is low here.

### ⚠️ Issues found

[net2/VirtWanGroup.js:451]  
Medium: The 5-minute `_lastStartTime` cutoff can still leave the linked issue unresolved if a Docker-based VPN takes longer than 5 minutes to stabilize after Dual WAN failover. The issue explicitly says Docker VPN failover can be slow/unstable and relies on later periodic `_checkConnectivity` events. After this cutoff, a valid later `link_established` will no longer repair a VWG table stuck at `unreachable default`. Consider clearing the recovery window only after a valid VWG default route is observed, or otherwise making the retry condition tied to route state instead of a fixed short time window.

[net2/VirtWanGroup.js:462]  
Medium: `_memberRoutesMissing()` returns `false` as soon as either IPv4 or IPv6 has a default route for the member interface. For dual-stack clients, this can miss a partially broken VWG table where IPv4 exists but IPv6 is still only `unreachable default` or vice versa. Since `refreshRT()` adds per-family routes independently, the recovery check should verify each expected address family and refresh when any expected default route is missing.

### 💡 Suggestions

- Add regression coverage for the linked issue flow: VWG table contains only `unreachable default`, VPN client emits a later `link_established` with `routeUpdated: false`, and `refreshRT()` is triggered.
- Add route parser tests for single-path defaults, multipath defaults, unreachable-only output, and partial IPv4/IPv6 route presence.
- Consider tokenizing `ip route` output instead of substring matching ` dev ${intf}` to avoid accidental prefix matches between similar interface names.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9305  
Head SHA: `162c64eba12f156c1759f89cd9cd9d07f887da3c`  
Checked at: 2026-07-09 13:06:04 CST

jasonlyc (approved):

PR #9313, opened 2026-07-10
add category to flow data

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The intent is solid: baking a compact flow category snapshot at write time can reduce repeated `intel:ip` lookups and improve query-time performance.
- `LimitedQueue.pushAndWait()` is covered with useful concurrency and error-path tests.
- Removing `af.ip` reduces per-flow storage size and fits the broader goal of keeping persisted flow records smaller.
- No linked issue found in the PR title/body/comments or GitHub closing issue metadata.

### ⚠️ Issues found

- **High — `HostManager` now sorts `undefined` entries.**  
  `[net2/HostManager.js:2854]`  
  The `asyncNative.mapLimit()` callback mutates `flow` but does not return it, so `enriched` becomes an array of `undefined` values. The following `.sort((a, b) => b.count - a.count)` will throw when this code path runs. Return `flow` from the callback, or use `eachLimit` and sort the original `traffic` array.

- **High — inline-intel readiness skips lookups for flows without a baked category.**  
  `[net2/LogQuery.js:464]`  
  `const getIntel = inlineIntelReady || f.category != null` means once the migration is considered ready, all query-time intel lookups are skipped, including callers without a baked snapshot. That contradicts the comment and breaks aggregated/top-sum flows that only provide `{ ip, appHosts }`. This should likely be `const hasSnapshot = f.category != null; if (!hasSnapshot) ...`, or otherwise preserve lookup behavior for records that do not carry `c/category`.

- **Medium — URL intel enrichment can dereference `conn.intel` when it is missing.**  
  `[net2/DNSManager.js:142]`  
  `enrichHttpFlow()` can now reach `conn.intel[key] = ...` when `conn.intel` is undefined, especially after `enrichDestIP()` returns early in inline-ready mode with only `flow.category` set. That turns URL-based intel override into a caught error and loses the enrichment. Initialize `conn.intel = conn.intel || {}` before copying URL fields.

- **Medium — data-usage top flows are now capped at 5000 per direction.**  
  `[sensor/DataUsageSensor.js:286]`  
  The previous implementation queried all raw flows in the time range via `queryFlows()`. The replacement uses `getDeviceLogs(... count: 5000, asc: true)` for each direction, which can truncate monthly windows and return the earliest 5000 records rather than all/top records. That can undercount or miss large destinations for heavy devices. This needs pagination or an aggregation API equivalent to the old full-range behavior.

### 💡 Suggestions

- Add focused regression tests for `HostManager.getStats()` or the specific top-sum path so the missing `return flow` failure is caught.
- Add a test case where `intelTool.isInlineIntelReady()` is true but a flow has only `ip/appHosts` and no `category`, to lock down the intended fallback lookup behavior.
- Add an `enrichHttpFlow()` unit test for `{ urls: [{ category: 'intel', ... }], c: 'porn' }` without an existing `intel` object.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9313  
Head SHA: `b0d1ae554a130036654e8a83e3296c9a1914989e`  
Checked at: 2026-07-10 12:41:42 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Inline flow intel snapshots (`c`) are a good direction to reduce query-time `intel:ip` lookups and preserve category data after intel expiry.
- The migration/backfill logic considers sumflow member splitting, which is an important compatibility detail.
- `LimitedQueue.pushAndWait()` is covered with focused tests for ordering, concurrency, errors, and worker wakeups.
- PR metadata has no body and `closingIssuesReferences` is empty, so: **No linked issue found.**

### ⚠️ Issues found

**High — category enum allocation is not safe across processes**

[net2/IntelTool.js:175]  
`categoryToNumber()` allocates new numeric category IDs from the process-local `this.categoryEnum` under an in-process `AsyncLock`. This does not protect multiple Firewalla processes from assigning the same next number concurrently. Two processes can both see max `7`, assign different categories to `8`, and persist/publish conflicting mappings. Since flow records store only the numeric code, this can permanently corrupt category decoding for historical flows.

Use a Redis-atomic allocation strategy, such as a Lua script or `HINCRBY`/`HSETNX` with a dedicated counter, and make the Redis state authoritative during allocation.

**Medium — data usage top-flow collection can silently drop flows**

[sensor/DataUsageSensor.js:286]  
`getSumFlows()` now reads device logs with `count: 5000, asc: true` per direction. The previous `queryFlows()` path read all matching raw flows in the time window. On busy devices or longer billing windows, only the first 5000 flows are included, so later high-volume destinations can be missed and the generated data-usage alarm context can be wrong.

This should paginate through the full time range or use an aggregation API that preserves the previous “all flows in window” behavior.

### 💡 Suggestions

- Add a concurrency test or integration test for category enum allocation across simulated independent `IntelTool` instances.
- Add regression coverage for `DataUsageSensor.getSumFlows()` with more than 5000 flows where the largest destination appears after the first page.
- Consider documenting whether `flow_inline_intel` backfill intentionally covers only the last 24 hourly buckets, especially if any relevant retention can exceed 24 hours.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9313  
Head SHA: `9209ffad4311d0985cae6cf105c33a54a42c9aa1`  
Checked at: 2026-07-13 13:57:52 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The overall direction is good: storing compact category snapshots on flow/audit records should reduce repeated `intel:ip` lookups and preserve category data after intel cache expiry.
- `LimitedQueue.pushAndWait()` has solid coverage for return values, concurrency limits, ordering, worker wakeups, and error handling.
- The latest revision fixed earlier hazards around `HostManager` returning `undefined` from `mapLimit()` and `DNSManager.enrichHttpFlow()` writing into a missing `conn.intel`.
- No linked issue found.

### ⚠️ Issues found

**High — category enum allocation is not safe across processes**

[net2/IntelTool.js:175]  
`categoryToNumber()` assigns new numeric category IDs from the process-local `this.categoryEnum` under an in-process `AsyncLock`. Firewalla has multiple processes that can call this path, so two processes can concurrently observe the same max code and assign different categories to the same next number. Since persisted flows store only the numeric code, this can permanently corrupt historical category decoding.

Use a Redis-atomic allocator: for example a Lua script, `HSETNX` plus a dedicated counter, or another single Redis transaction that makes “category already exists / allocate next code / persist mapping” authoritative in Redis.

**Medium — data usage top-flow context can silently omit most flows**

[sensor/DataUsageSensor.js:286]  
`getSumFlows()` now reads `getDeviceLogs()` with `count: 5000, asc: true` per direction. The removed `queryFlows()` path read all raw flows in the requested time range. For busy devices or monthly windows, this only includes the earliest 5000 inbound and outbound records, so later high-volume destinations can be missed or undercounted in data-usage alarm context.

This should paginate through the full window, or switch to an aggregation source that preserves the old “all flows in range” behavior.

### 💡 Suggestions

- Add a regression test that simulates two independent `IntelTool` instances allocating new categories concurrently.
- Add coverage for `DataUsageSensor.getSumFlows()` with more than 5000 flows where the largest destination appears after the first page.
- Consider documenting whether `flow_inline_intel` backfill intentionally covers only the last 24 hourly buckets, especially if any top-flow retention/configuration can exceed that window.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9313  
Head SHA: `c663d25c3f53ae9999a9d897c50a88d31ee422cd`  
Checked at: 2026-07-13 14:15:36 CST

jasonlyc (approved):

PR #9321, opened 2026-07-10
sync gold pro 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Good direction on removing shell-heavy dnsmasq hashing and replacing it with structured filesystem traversal.
- The per-request IV encryption path keeps legacy zero-IV behavior for old clients while allowing upgraded clients to opt in.
- `KernelCrashMonitor` has reasonable locking and tests around pstore crash detection, Redis persistence, and module-version changes.
- The STP-port flow deduplication logic is isolated and covered with focused unit tests.
- QUIC log parsing now handles malformed/concatenated log lines more defensively.

### ⚠️ Issues found

[net2/VirtWanGroup.js:437]  
`_hasValidDefaultForDev()` can return a false positive when a non-target default route is followed by any non-indented route containing `dev ${intf}`. The code checks `trimmed.includes(dev)` before ending the previous default-route block, so a normal connected route can be mistaken as part of the default route. This can make `_memberRoutesMissing()` return `false` and skip the route-table refresh that this change is trying to trigger. End the default block on non-indented lines before checking `dev`, or only check continuation lines that are actually indented multipath entries.

[extension/freeradius/freeradius.js:422]  
The new `else` branch deletes all custom cert files whenever `data.customCerts` is absent or not an object. If `saveConfig()` is ever called with a partial config update that does not include `customCerts`, existing custom certificates will be removed unintentionally. Please gate deletion on an explicit “clear custom certs” signal or confirm all callers always send the full desired cert state.

[net2/KernelCrashMonitor.js:182]  
The UDP TLS crash detection only matches `Modules linked in:.*xt_udp_tls` on one line. Kernel module lists can wrap across lines in panic output, which would cause this safety path to miss an `xt_udp_tls` panic and leave the module enabled. Consider scanning the panic files for `xt_udp_tls` in the module-list section more robustly, or at least checking panic files for both `Kernel panic` and `xt_udp_tls` without requiring the same line.

### 💡 Suggestions

- Issue requirement coverage: the target PR body has no linked issue description. GitHub shows commit/PR references including `Fixes #8590`, but `#8590` appears to be an old merged PR titled “add source info for bonjour identifications,” not a requirement source for this staging sync. Treat as no actionable linked issue for this review.
- Consider adding a unit test for `_hasValidDefaultForDev()` where a default route for `eth0` is followed by a separate `tun0` route.
- For FreeRadius cert migration, add a regression test for “update config without `customCerts` keeps existing custom cert files” if partial updates are supported.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9321  
Head SHA: `f74786ff4009498c04b243f10d14a6bbab09b096`  
Checked at: 2026-07-10 15:06:19 CST

PR #9324, opened 2026-07-10
fix bug, only reload tls/udp_tls module when needed

caixingcheng:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Reviewed 1 changed file: `control/install_iptables_setup.sh`.
- The direction is good: avoiding unnecessary TLS module reloads should reduce disruptive `iptables-save` / `iptables-restore` cycles.
- The helper correctly checks platform support before deciding whether `xt_tls` or `xt_udp_tls` should be considered for update.
- The `.ko` `srcversion` check and `.so` checksum check are reasonable signals for deciding whether the bundled artifacts differ from the installed/runtime state.

### ⚠️ Issues found

- **Medium** `[control/install_iptables_setup.sh:319]`

  `tlsModuleNeedsUpdate` returns “needs update” immediately when `xt_udp_tls` is not loaded, before honoring the existing crash-disable guard in `installTLSModule`.

  If `kernel_crash_info.shouldDisableUdpTls` is true, every run will still add `udp_tls` to `modules_to_update`, strip TLS/UDP-TLS rules from the active ruleset, call `installTLSModule`, and then have `installTLSModule` skip the install. That defeats the stated goal of avoiding the disruptive restore cycle when no install should happen, and it risks restore failures if the saved rules contain `-m udp_tls` while the module remains intentionally unloaded.

  Consider adding the same `xt_udp_tls` crash-disable check to `tlsModuleNeedsUpdate` before returning `0` for the “not loaded yet” case.

### 💡 Suggestions

- `[control/install_iptables_setup.sh:374]` Fix the typo in the comment: `resotre` → `restore`.
- `[control/install_iptables_setup.sh:381]` Use `grep -wq` here to avoid printing `lsmod` output during normal setup runs.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9324  
Head SHA: `e61a9e2ab96924c1c8fdcd0e01829aca288efa09`  
Checked at: 2026-07-10 16:57:59 CST

jasonlyc (approved):

PR #9325, opened 2026-07-10
fix bug, only reload tls/udp_tls module when needed

caixingcheng:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The change correctly avoids the disruptive TLS rule removal / module reload path when neither `xt_tls` nor `xt_udp_tls` needs an update.
- Platform support checks are preserved before attempting install/update.
- Comparing bundled `.ko` `srcversion` and xtables `.so` checksum is a reasonable lightweight way to decide whether reload is necessary.
- No linked issue found.

### ⚠️ Issues found

- **Medium** `control/install_iptables_setup.sh:369-383`  
  If either temporary `iptables-restore` fails, the script still proceeds to `rmmod` / `installTLSModule` and finally restores the saved rules. That can leave the update path operating under the false assumption that TLS rules were removed, causing `rmmod` to fail or producing a partial update. Since this PR is specifically reducing disruption around module reloads, this block should check restore failures and abort the module update path while restoring the original rules.

- **Low** `control/install_iptables_setup.sh:322-329`  
  If `get_tls_ko_path` points to a bundled `.ko` but `modinfo` fails or produces no `srcversion`, the code silently treats the loaded module as current. That can skip a needed kernel module update. Consider treating a present bundled `.ko` with unreadable/missing `srcversion` as needing update, or log clearly why the version check was skipped.

### 💡 Suggestions

- Fix typo in the nearby comment: `resotre` → `restore`.
- Consider logging why each module is selected or skipped. That would make field debugging easier when the reload is intentionally avoided.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9325  
Head SHA: `698a86ecb208d140e0f488d2143a74fa7bfb28a8`  
Checked at: 2026-07-10 18:00:27 CST

PR #9326, opened 2026-07-10
Fix action history for msp query

xinnige:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- `appInfo.operator` is carried into trace/action history, which satisfies the operator attribution requirement.
- `operator.id === 'bg_job'` is skipped at record time and also filtered defensively at read time.
- `target` is now preserved in `RecordAction`, making target filtering usable for newly recorded actions.
- Network config changes now attempt to record prior state and diff data.

### ⚠️ Issues found

- **High: `autoTriggered` is checked at the wrong path.**  
  Linked issue `firewalla/firecommit#8626` defines the contract as `action.data.autoTriggered`, but [sensor/ActionPlugin.js:30] checks only `action.autoTriggered`. Auto side-effect records from MSP will still appear in action history unless producers flatten the field, which the issue explicitly says they do not.

- **Medium: the test fixture encodes the wrong contract.**  
  [test/test_action_sensor.js:31] creates `autoTriggered` directly under `action`, so the new test passes while missing the real MSP payload shape. Add coverage for `{"action":{"data":{"autoTriggered":true}}}`.

- **Medium: `difference()` now changes global semantics.**  
  [util/util.js:319] treats `false`, `null`, `undefined`, and `""` as equivalent defaults for every caller of `difference()`. That may be useful for action-history noise reduction, but it can hide legitimate changes elsewhere, such as explicit boolean/string resets. Prefer a scoped helper or an option for action-history diffing.

**Issue requirement coverage:**

- Operator in appInfo: **Met**
- Recognize/add `autoTriggered`: **Partially met**, wrong field path
- Skip bg_job action history records: **Met**
- Default filter out auto-triggered and bg_job records: **Partially met**, bg_job works; auto-triggered likely does not
- Better origin/diff for policy changes: **Partially met**, but via global `difference()` behavior change
- NetworkConfig diff in action history: **Met**

### 💡 Suggestions

- Update `hiddenAction()` to check both `action.autoTriggered` and `action.data.autoTriggered` if backward compatibility is needed.
- Add tests for `appInfo.operator.id === 'bg_job'` at `recordAction()` and `getActionHistory()` levels.
- Consider trimming whitespace in comma-separated platform filters, e.g. `"ios, android"`.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9326  
Head SHA: `60acdba477745f6ba97329d54de42f84e65b8e17`  
Checked at: 2026-07-10 18:38:38 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Carries `appInfo.operator` into trace/action history, which supports operator attribution.
- Skips `bg_job` records both defensively in history reads and at record time.
- Preserves `target` in `RecordAction`, so target filtering can work for newly recorded actions.
- Adds test coverage for hidden actions and platform filtering.

### ⚠️ Issues found

- **High:** `autoTriggered` is checked at the wrong path. The linked issue contract appears to use `action.data.autoTriggered`, but `sensor/ActionPlugin.js` only checks `action.autoTriggered`. MSP auto side-effect records will still show up if producers send the nested shape.

  `[sensor/ActionPlugin.js:30]`
  Check `action.data.autoTriggered` as well, or move producers/tests to one explicit payload contract.

- **Medium:** The new test fixture validates the wrong payload shape. `test/test_action_sensor.js` puts `autoTriggered` directly under `action`, so the test can pass while missing the real MSP payload.

  `[test/test_action_sensor.js:31]`
  Add coverage for `{"action":{"data":{"autoTriggered":true}}}`.

- **Medium:** `difference()` now changes global semantics by treating `false`, `null`, `undefined`, and `""` as equivalent defaults. That can hide legitimate updates elsewhere, especially explicit boolean disables or string clears.

  `[util/util.js:319]`
  Prefer a scoped action-history diff helper or an options flag instead of changing the shared utility for every caller.

- **Issue requirement coverage:** Partially met. Operator attribution, `bg_job` skipping, target preservation, and network config recording are covered. Auto-trigger filtering is only partially met because the field path likely does not match the linked issue requirement.

### 💡 Suggestions

- Trim comma-separated platform filters so `"ios, android"` works as expected.
- Add tests for `appInfo.operator.id === "bg_job"` at both `recordAction()` and `getActionHistory()` levels.
- Consider documenting the expected action history payload shape, especially `action.data` vs flattened `action` fields.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9326  
Head SHA: `42d1b37ead3218fa1c8f65f407e6eddc3afcf4d2`  
Checked at: 2026-07-10 19:20:08 CST

jasonlyc (approved):

PR #9330, opened 2026-07-13
scripts: add update_member_name.js to preview encrypted member name

MelvinTo:
## What

Adds `scripts/update_member_name.js`, a small read-only helper that, given a group member `eid` and a new cleartext account email, prints the AES-encrypted value as it would be stored in the group's `symmetricKeys[].name` in the Encipher cloud.

```
cd /home/pi/firewalla
OPENSSL_CONF=/dev/null bin/node scripts/update_member_name.js <eid> "<new email>"
```

Output shape:
```
gid           : <gid>
eid           : <eid>
old name      : <current email>
new name      : <new email>
new encrypted : <base64 ciphertext>
```

## How

Reuses `EptCloud`'s own crypto (`loadKeys` + `reloadGroupInfoFromRedis`) and the group info cached in redis (`sys:ept:me`). It:

- loads the box RSA key and the cached group,
- recovers the decrypted group symmetric key,
- decrypts the member's current `name` for reference,
- AES-256-CBC encrypts the new email with the group key (same path as `EptCloud.encrypt`).

## Notes

- **Read-only and offline** — never contacts the cloud and writes nothing (no commit flag, no redis writes). Purely a preview/verification tool.
- Verified on a live box: re-encrypting an existing member's email reproduced the exact ciphertext stored in the cloud group record (deterministic because the legacy `encrypt` path uses a fixed IV).
- `OPENSSL_CONF=/dev/null` is needed on the box because node 12 trips over the system OpenSSL engine config.

### DIFF
--- a/scripts/update_member_name.js
+++ b/scripts/update_member_name.js
@@ -1,0 +1,55 @@
+#!/usr/bin/env node
+/*
+ * Show the encrypted "name" (account email) for a group member entry.
+ *
+ * Given a member eid and a new cleartext email, print the AES-encrypted
+ * value as it would be stored in symmetricKeys[].name in the Encipher
+ * cloud. Read-only / offline: it reuses the box's own EptCloud crypto and
+ * the group info cached in redis (sys:ept:me), never contacting the cloud.
+ *
+ * Run on the box, from the firewalla home:
+ *   cd /home/pi/firewalla
+ *   OPENSSL_CONF=/dev/null bin/node scripts/update_member_name.js <eid> "<new email>"
+ */
+'use strict';
+
+const cloud = require('../encipher');
+
+async function main() {
+  const eid = process.argv[2];
+  const newName = process.argv[3];
+
+  if (!eid || !newName) {
+    console.error('usage: node scripts/update_member_name.js <eid> "<new email>"');
+    process.exit(1);
+  }
+
+  const eptcloud = new cloud('netbot');
+  await eptcloud.loadKeys();
+  await eptcloud.reloadGroupInfoFromRedis();
+
+  const gid = Object.keys(eptcloud.groupCache)[0];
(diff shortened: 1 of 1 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Reviewed 1 of 1 changed files.
- The helper is intentionally read-only and reuses the existing `EptCloud.encrypt/decrypt` path, which is the right way to avoid crypto drift.
- The usage comments are clear and include the `OPENSSL_CONF=/dev/null` runtime detail.

### ⚠️ Issues found

- Medium `[scripts/update_member_name.js:27]`  
  `reloadGroupInfoFromRedis()` depends on `parseGroup()`, which requires `eptcloud.eid`. The constructor sets `eid` asynchronously via `rclient.hgetAsync(...).then(...)`, but this script never awaits that read. On a fresh short-lived process, `parseGroup()` can run before `eid` is populated, return `null`, and the script reports `no group info cached` even though Redis has valid data. Please explicitly load/await the local eid before reloading group info, or add an `EptCloud` helper that guarantees identity state is ready.

### 💡 Suggestions

- Consider validating that `newName` looks like a reasonable account email before printing replacement ciphertext, since this script is operationally used to generate a cloud-stored member name value.
- The script name says `update_member_name`, while the implementation only previews/prints the encrypted value. A name like `encrypt_member_name.js` or `preview_member_name_update.js` would reduce operator confusion.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9330  
Head SHA: `3bfbda453970168bfbaffec418b3419e4bb938d6`  
Checked at: 2026-07-13 19:19:20 CST

jasonlyc (approved):

PR #9331, opened 2026-07-13
ept: allow overriding member name (email) via sys:ept:memberEmails

MelvinTo:
## What

Adds a redis hash `sys:ept:memberEmails` (field = member `eid`, value = email) that overrides the cloud-snapshotted member name returned for group members. When an eid has an entry in this hash, that email is used instead of the `name` stored in `sys:ept:members`.

## Why

The member `name` in the Encipher group record is the account email, AES-encrypted and snapshotted at invite/re-key time, so it is effectively immutable without re-inviting the eid. This adds a local, box-side override so the email shown for a member can be updated without touching the cloud group record.

## Changes

- **net2/Constants.js** — add `REDIS_KEY_EPT_MEMBER_EMAILS: "sys:ept:memberEmails"`.
- **net2/HostManager.js** (`encipherMembersForInit`) — overlay the email onto `m.name` when the eid has an entry, so the existing `eptGroup` get API returns the overridden value. Applied the same way as the existing `memberNames`/`lastvisit` overlays.
- **controllers/netbot.js** (`setHandler`) — add an `eptMemberEmail` command taking `{ eid, email }` and writing it to the hash; rejects missing eid/email with a 400.
- **extension/ept/eptcloud.js** and **sensor/EncipherPlugin.js** — remove the eid's email entry on unpair/revoke, consistent with the existing cleanup of `memberNames` and `member:lastvisit` (and include the new hash's keys in the stale-eid sweep).

## API

Set the email for a member (local API, port 8834):

```
curl -s -X POST -H 'Content-Type: application/json' \
  -d '{"eid":"<EID>","email":"<new email>"}' \
  'http://localhost:8834/v1/encipher/simple?command=set&item=eptMemberEmail'
```

Read back (existing command; now reflects the override in each member's `name`):

```
curl -s 'http://localhost:8834/v1/encipher/simple?command=get&item=eptGroup'
```

## Verification

- `node --check` passes on all five changed files.
- Overlay logic unit-checked in isolation: an eid with an entry gets its `name` overridden, members without an entry are untouched, and `dName` (from `memberNames`) is still applied independently.
- Note: end-to-end verification on a live box (deploy + fireapi restart) was intentionally not performed here; the change mirrors the adjacent, already-proven `memberNames` overlay and set-command patterns.

### DIFF
--- a/controllers/netbot.js
+++ b/controllers/netbot.js
@@ -1152,6 +1152,13 @@ class netBot extends ControllerBot {
           throw new Error("rename failed")
         }
       }
+      case "eptMemberEmail": {
+        const { eid, email } = value;
+        if (!eid || !email)
+          throw { code: 400, msg: "both eid and email are required" };
+        await rclient.hsetAsync(Constants.REDIS_KEY_EPT_MEMBER_EMAILS, eid, email);
+        return
+      }
       case "intelAdvice": {
         const { target, ip, intel } = value
         intel.localIntel = await intelTool.getIntel(ip)
--- a/extension/ept/eptcloud.js
+++ b/extension/ept/eptcloud.js
@@ -51,8 +51,9 @@ class EptCloudExtension {
     //clear members info which is not in the group
     const names = (await rclient.hgetallAsync("sys:ept:memberNames")) || {};
     const lastVisits = (await rclient.hgetallAsync("sys:ept:member:lastvisit")) || {};
+    const emails = (await rclient.hgetallAsync(Constants.REDIS_KEY_EPT_MEMBER_EMAILS)) || {};
 
-    const allEids = new Set([...Object.keys(names), ...Object.keys(lastVisits)])
+    const allEids = new Set([...Object.keys(names), ...Object.keys(lastVisits), ...Object.keys(emails)])
     const clientsEidList = clients.map(client => client.eid);
     const eidsToDelete = Array.from(allEids).filter(eid => !clientsEidList.includes(eid));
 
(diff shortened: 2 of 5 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The change is well scoped and follows the existing Redis overlay pattern for EPT member metadata.
- Cleanup is covered in both stale group sweep and explicit revoke/unpair flow.
- `HostManager.encipherMembersForInit()` applies the override before returning `eMembers`, matching the PR goal.
- No linked issue found.

### ⚠️ Issues found

[controllers/netbot.js:1156]  
Malformed requests with missing or null `value` will throw a destructuring `TypeError` and likely return a 500 instead of the intended 400. Use `const { eid, email } = value || {}` and validate both fields explicitly.

[controllers/netbot.js:1159]  
This creates a persistent override but does not provide a way to clear or revert it except unpair/revoke or manual Redis edits. Since `HostManager` always prefers this hash when present, a typo or obsolete email can remain stuck. Consider supporting `email: null` / empty string as `hdel`, or adding an explicit delete command.

### 💡 Suggestions

- Validate `email` as a trimmed string with a reasonable max length and basic email shape before storing it.
- Consider checking that `eid` belongs to the current group before accepting the override, to avoid stale or invalid entries living until the next cleanup sweep.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9331  
Head SHA: `1e50bb5a45c0452506523aadc529b28757968ad4`  
Checked at: 2026-07-13 20:02:47 CST

MelvinTo:
Tracking issue: firewalla/firecommit#8707

jasonlyc (approved):

PR #9335, opened 2026-07-15
key guardian live transports by item+streaming.id and refresh subscription on re-register

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Reviewed 2/2 changed files: `sensor/Guardian.js` and `sensor/LiveTransport.js`.
- The implementation addresses linked issue `firewalla/firecommit#8702` by separating cached live transports with `item + streaming.id`, so concurrent same-item streams no longer share one `LiveTransport`.
- Keepalive and close messages now resolve the same cache key before acting on the transport, so lifecycle actions are scoped per stream when `streaming.id` is present.
- Re-registering an existing transport refreshes `message`, `replyid`, socket, delay, and expiry, which fixes the stale first-request behavior described in the issue.

### ⚠️ Issues found

No blocking issues found.

Issue requirement coverage:
- Met: cache key includes a per-query discriminator via `streaming.id`.
- Met: concurrent same-item streams can now maintain separate push loops.
- Met: re-subscribe can update the stored request and `replyid`.
- Met with compatibility fallback: requests without `streaming.id` still fall back to `item`, preserving old behavior but also preserving old collision behavior for legacy callers without IDs.

### 💡 Suggestions

- Consider adding a focused unit/integration test for two concurrent `liveStats` subscriptions with different `streaming.id` values to prevent regression.
- Consider documenting that `streaming.id` is required for collision-free concurrent streaming; without it, the fallback intentionally behaves like the old alias-only cache.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9335  
Head SHA: `efd8d47135c8e1529a93cb6f040cf49c3badc96d`  
Checked at: 2026-07-15 11:36:53 CST

jasonlyc (approved):

PR #9337, opened 2026-07-15
fix bug, error during auto delete expired bypass rule

caixingcheng:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The `PolicyManager2` change fixes the expired auto-delete bypass flow by avoiding `_disablePolicy()` after `unenforce()` may already delete the policy.
- `removeBypassChainForPolicy()` is still called after deletion, which matches its expectation that the policy has already been removed from Redis.
- Keeping `enforcedPolicies[pid]` as an empty object in `AppTimeUsageManager` is consistent with later code that indexes `this.enforcedPolicies[pid][uid]`.

### ⚠️ Issues found

No blocking issues found.

No linked issue found.

### 💡 Suggestions

- Consider adding a regression test for an expiring bypass policy with `autoDeleteWhenExpires == "1"` to verify it does not throw `Policy not exist`.
- The new comment says `deletePolicy` is idempotent. That is effectively true for the Redis operations, but `deletePolicy()` currently calls async `policyExists()` without `await`, so the guard is not doing what it appears to do. Worth cleaning up separately.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9337  
Head SHA: `481f379474abe75a08beed53224b3dff546a5d7b`  
Checked at: 2026-07-15 15:19:20 CST

jasonlyc (approved):

PR #9339, opened 2026-07-16
sync gold 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Good direction on replacing shell-string `exec` with `execFile` in DNS/WireGuard/NIC paths.
- The GCM envelope work adds authenticated encryption and validates nonce/tag lengths.
- UDP TLS crash-disable handling is better centralized through platform support checks.
- Tests were added for `LimitedQueue`, action history, kernel crash handling, and encipher IV/GCM paths.

### ⚠️ Issues found

[net2/IntelTool.js:188] **High**  
`categoryToNumber()` uses an in-process `AsyncLock`, but this enum is shared by Redis and updated by multiple Firewalla processes. Two processes can see the same current max, assign the same numeric code to different new categories, then `SET` different full maps. Persisted flow records store only the numeric code, so a collision or lost update corrupts category decoding permanently. This needs a Redis-side atomic allocation/update path, such as Lua, `WATCH/MULTI`, or a Redis lock plus re-read inside the distributed lock.

[sensor/DataUsageSensor.js:286] **Medium**  
`getSumFlows()` now reads only `count: 5000` logs per direction for the whole interval. The previous `queryFlows()` path read all raw flows in the requested range before grouping. For monthly data-usage alarms, busy devices can easily exceed 5,000 flows, so later or larger destinations may be omitted from the top-flow list. Please paginate until `ets` or switch to an aggregate/top-flow source that covers the entire interval.

[net2/BroDetect.js:1533] **Medium**  
The conn ingest path now awaits intel enrichment before persisting every non-local flow. `processIP()` can perform Redis/DNS/cloud intel work, and under intel latency this turns flow logging into a blocking path instead of the previous async post-write enrichment. That can back up Bro/Zeek processing and risks delayed or dropped flow writes. Consider writing the flow first and updating/aggregating intel asynchronously, or add a small timeout/fallback that persists without `c` when enrichment is slow.

[util/asyncNative.js:84] **Medium**  
`LimitedQueue.size` reports only queued items, not in-flight work. `DestIPFoundHook.appendNewFlow()` uses this as the overload guard, so the system can still accept `QUEUE_SIZE_PAUSE + limit` outstanding items and, more importantly, monitoring publishes only queued backlog while up to 100 active intel jobs are hidden. If this is used as backpressure, track `pending = queued + running`.

[encipher/lib/encipherio.js:99] **Low**  
`this.gid` is loaded asynchronously in the constructor, while GCM decrypt/encrypt throws if `_localGid()` runs before that Redis read completes. A GCM request during startup can fail with `decrypt_error` even though the key is valid. Load the local gid through the same awaited initialization path as group/key loading, or have `_localGid()` lazily fetch/cache it before throwing.

### 💡 Suggestions

- Add a concurrency test for category enum allocation from separate processes or simulated stale enum snapshots.
- Add a data-usage regression test where more than 5,000 flows exist in the interval and the largest destination appears after the first page.
- Add an ingest-path test or benchmark proving conn saves are not delayed by slow intel lookups.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9339  
Head SHA: `47b4b01a62491ee08bcb8af09083e002a952821e`  
Checked at: 2026-07-16 14:06:58 CST

PR #9341, opened 2026-07-16
sync purple se 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

Reviewed 47 changed files. No linked issue found.

- Good move from shell-built `dig` commands to `execFile`, reducing command injection risk.
- The UDP TLS crash-disable flow is better centralized through `KernelCrashMonitor`.
- Added useful tests for `LimitedQueue`, GCM encryption, action filtering, and kernel crash behavior.
- Several changes reduce expensive query-time intel lookups by baking compact category snapshots into flow/audit records.

### ⚠️ Issues found

[net2/IntelTool.js:179]  
**High:** `categoryToNumber()` uses only a process-local `AsyncLock` while persisting a shared Redis enum. FireMain, FireApi, FireMon, or migration code can concurrently assign the same `next` code to different newly-seen categories, then overwrite `intel:enum:category` and publish conflicting deltas. Since flow records persist numeric codes, this can permanently misclassify historical flows. This needs a Redis-atomic allocation path, e.g. Lua / `WATCH` transaction / Redis hash plus monotonic counter / distributed lock.

[net2/BroDetect.js:1533]  
**Medium:** flow persistence now awaits inline intel enrichment for every non-local flow before writing to Redis. `processIP()` can hit Redis/cloud/DNS-style work and the caller blocks until the queue worker completes. Under backlog or cloud slowness, this can delay Zeek flow ingestion and increase flow loss risk. Consider persisting first with best-effort async enrichment, or enforce a short timeout/fallback before write.

[encipher/lib/encipherio.js:99]  
**Medium:** GCM depends on `this.gid`, but it is populated by an unawaited Redis read in the constructor and failures are swallowed. A GCM request or response before that read completes, or after a transient Redis failure, will fail with `local gid not available for GCM AAD`. Make gid loading explicit before accepting GCM, retry on failure, or derive the trusted local gid from already-loaded group state.

[net2/HostManager.js:2874]  
**Low:** this path stopped calling `flowTool.enrichWithIntel()` and now only fills country. `getTopSumFlowByKeyAndDestination()` also does not carry/decode the new `c` category field, so `systemFlows.upload/download/ipB/dnsB` can lose category/app enrichment even though other top-flow paths still enrich. If this API feeds clients, preserve category decoding here.

### 💡 Suggestions

- Add a concurrency test for category enum allocation across two simulated processes or Redis clients.
- Add a startup/GCM test where `this.gid` is initially unset to catch the async initialization failure mode.
- For inline intel, add timeout metrics/logging around `appendNewFlow(..., true)` so production can show whether ingestion is being held up.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9341  
Checked at: 2026-07-16 18:50:00 CST

PR #9342, opened 2026-07-16
sync gold se 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Good security cleanup replacing shell-built `dig`/`wg` commands with `execFile`.
- AES-GCM envelope support is a meaningful improvement over unauthenticated CBC, with tests covering tamper/tag/AAD validation.
- UDP TLS crash-disable logic is being pushed into a central platform capability check, which is the right direction.
- Several new tests were added around `LimitedQueue`, action history filtering, encryption, and kernel crash cache behavior.

### ⚠️ Issues found

**High** `[net2/IntelTool.js:153]` / `[net2/IntelTool.js:175]`  
The category enum allocator is not safe across Firewalla’s multi-process runtime. `AsyncLock` only serializes inside one Node process, and `loadCategoryEnum()` is started without being awaited. A process can start with only defaults, assign `next` for a new category, and overwrite Redis while another process has already assigned a different category to the same number. Since persisted flow records store only numeric `c`, this can permanently misclassify historical flow/category data. Use a Redis-backed lock or optimistic transaction, and ensure the Redis enum is loaded before allocating new codes.

**Medium** `[sensor/DataUsageSensor.js:286]`  
`getSumFlows()` now reads at most 5000 logs per direction via `getDeviceLogs()`, while the removed `queryFlows()` path read all raw flows in the requested window. For busy devices, abnormal bandwidth alarms can now build top destinations from only the first 5000 chronological records and miss the actual heavy destination later in the window. Either page through the whole range, query aggregated data, or make the limit explicit and large enough for the alarm semantics.

**Low** `[control/iptables_common.sh:684]`  
`local should_disable` is used at top level, outside a function. Bash emits `local: can only be used in a function`; depending on caller options this can become noisy or fatal. Drop `local` here.

### 💡 Suggestions

- Add a test or integration check for concurrent category allocation from separate processes, not only same-process calls.
- Add a regression test for `DataUsageSensor.getSumFlows()` with more than 5000 flows to verify top destination accuracy.
- Consider documenting that inline intel category codes are durable storage format and must never be renumbered.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9342  
Checked at: 2026-07-16 18:46:55 CST

PR #9343, opened 2026-07-16
sync gold pro 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Good security hardening in DNS checks by replacing shell-built `dig` commands with `execFile`.
- AES-GCM request/response scheme is integrated with tests for round-trip and tamper cases.
- `LimitedQueue.pushAndWait()` is well covered with concurrency and failure tests.
- TLS module reload now avoids unnecessary disruptive `rmmod` / restore cycles.

### ⚠️ Issues found

- **High** `[control/iptables_common.sh:681]`  
  `local should_disable` appears at top-level script scope. In bash, `local` is only valid inside a function and can abort initialization if `set -e` is active. This can break iptables generation exactly in the UDP TLS crash-disable path this PR is trying to protect. Use plain `should_disable=...` unless this block is inside a function.

- **High** `[net2/IntelTool.js:151]` / `[net2/IntelTool.js:173]`  
  `initCategoryEnum()` starts `loadCategoryEnum()` asynchronously but `categoryToNumber()` can run before Redis state is loaded. If Redis already contains an authoritative enum with non-default/new category codes, an early unseen category can be assigned from the default map and overwrite/collide with existing codes. Since stored flow records persist numeric category codes, this can permanently mis-decode historical flow/category data. Please await enum initialization before encoding, or make `categoryToNumber()` first await the initial load under the same lock.

- **Medium** `[sensor/DataUsageSensor.js:282]`  
  Monthly data usage now reads `getDeviceLogs(... count: 5000)` instead of scanning all raw flows in the time window. For busy devices/monthly windows, this can silently exclude older flows and undercount top destinations, which may suppress or skew large data usage alarms. Either page through all logs for `[begin, end]` or use an aggregation source that preserves complete byte totals.

- **Medium** `[net2/HostManager.js:2871]`  
  Top-flow enrichment was reduced from `flowTool.enrichWithIntel()` to country-only enrichment. Unless `getTopSumFlowByKeyAndDestination()` already decodes `c` and resolves host/category/app elsewhere, API consumers will lose category/app enrichment for top flows, especially during the migration window.

### 💡 Suggestions

- Add a startup/unit test for existing Redis `intel:enum:category` containing categories beyond defaults, then call `categoryToNumber()` immediately after constructing `IntelTool`.
- Add a shellcheck-style validation for `control/*.sh`; it would catch the top-level `local`.
- For the inline intel migration, add an integration test that verifies old and new sumflow members merge under the same destination after backfill.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9343  
Checked at: 2026-07-16 18:43:36 CST

PR #9345, opened 2026-07-16
do not disable udp_tls module when crash is older than the module

caixingcheng:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The stale-crash guard is directionally sound: comparing the pstore crash timestamp against the installed module file mtime prevents a first-run post-upgrade stale `xt_udp_tls` crash from disabling the newly installed module.
- The implementation keeps the previous fallback behavior when `stat` cannot determine the module mtime, so it does not silently weaken crash protection.
- Tests cover the core new behavior:
  - stale crash older than current module does not disable UDP TLS
  - crash newer than current module still disables UDP TLS

### ⚠️ Issues found

[test/test_KernelCrashMonitor.js:74]
The fake `grep` parser is now stale after adding `grep -l -e 'Kernel panic' -e 'Oops' ...paths`. `args.slice(3)` no longer points only at paths, and the fake still only checks for `'Kernel panic'`. That means the new production behavior for detecting `Oops` is not actually testable with this helper, and a future Oops-specific test would incorrectly fail or need to work around the mock. Please update the fake to parse `-e` patterns and remaining path args, then add a test proving an `Oops` pstore entry is treated as a crash.

[net2/KernelCrashMonitor.js:244]
The log message still says `Kernel panic detected` even though this branch now also handles `Oops`. This is minor, but it can mislead debugging when the actual pstore signal was an Oops. Consider changing it to `Kernel crash detected`.

### 💡 Suggestions

- Add one regression test for an `Oops` entry with `Modules linked in: xt_udp_tls` to verify it disables UDP TLS.
- Consider adding a boundary test for `latestCrashTSSec === koMtimeSec` so the intended same-second behavior is explicit.
- No linked issue found in the PR title/body/closing issue metadata, so there were no external issue requirements to verify.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9345  
Head SHA: `924339b0df8a356d78144bf1e82630eda5f6b400`  
Checked at: 2026-07-16 19:14:32 CST

jasonlyc (approved):

PR #9349, opened 2026-07-17
only do full reload from firerouter on essential network config change

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- The change directly targets the linked issue: `apc`-only / switch VLAN updates no longer trigger a full FireRouter reload.
- Essential config categories are explicit and match the issue test notes: `interface`, `routing`, `dns`, `dhcp`, `hostapd`.
- Backward compatibility for the old empty `MSG_NETWORK_CHANGED` payload is mostly preserved because invalid JSON falls back to full reload.

### ⚠️ Issues found
- **Medium — `changedKeys` is not validated before `.filter()`**
  
  `[net2/FireRouter.js:411]`
  
  If the message parses successfully but does not contain an array `changedKeys` value, for example `{}` or `{"changedKeys":null}`, this will throw at runtime. Since this is a pub/sub message boundary, it should validate with `Array.isArray(changedKeys)` and fall back to full reload for unknown payload shapes.

- **Medium — `Object.keys(routerConfig)` can throw if cached config is not initialized**
  
  `[net2/FireRouter.js:1199]`
  
  The previous `_.omit(routerConfig, ['apc'])` path was tolerant of nullish input, but `Object.keys(routerConfig)` is not. If `setConfig()` runs before `routerConfig` is loaded or after a failed reload path, this can crash before sending the FireRouter API request. Consider defaulting to `{}` or ensuring `routerConfig` is initialized before computing changed keys.

### 💡 Suggestions
- Consider defining `essentialKeys` as a module-level constant so the criteria are easy to audit and reuse in tests.
- Add a small unit/integration test around these message cases:
  - empty legacy payload triggers full reload
  - `{"changedKeys":["apc","ts","reqId"]}` avoids full reload
  - `{"changedKeys":["interface"]}` triggers full reload
  - malformed or non-array `changedKeys` safely falls back to full reload

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9349  
Head SHA: `21e5b350af381ce4e9c537d8777fa08ae04abea6`  
Checked at: 2026-07-17 20:05:01 CST

MelvinTo (approved):

PR #9351, opened 2026-07-20
fix bug, app timelimit not work for VPN device

caixingcheng:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The PR directly addresses linked issue `firewalla/firecommit#8815`: explicit `timeout: 0` is no longer dropped by `Ipset.add()`.
- The added `timeout 0` in `Identity.updateIPs()` correctly handles the raw `Ipset.restore()` path, which would otherwise bypass the fixed helper.
- IPv4 and IPv6 identity tag ipsets are both covered.
- The rationale comment in `Identity.js` is useful and explains the permanent-entry behavior clearly.

Issue requirement coverage: **Met**. The issue requires preserving explicit `timeout: 0` so permanent ipset entries do not fall back to the set default timeout. This PR covers both `Ipset.add(..., { timeout: 0 })` and the identity restore path.

### ⚠️ Issues found

No blocking correctness, security, concurrency, or networking issues found in the changed diff.

### 💡 Suggestions

- Consider adding a focused unit test for `Ipset.add()` verifying that `{ timeout: 0 }` emits `timeout 0`.
- The new timeout guard accepts numeric-looking values but appends the original value. Normalizing once, for example `const timeoutNumber = Number(timeout)`, would avoid surprising behavior if a non-number value is ever passed.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9351  
Head SHA: `492fad8fa2835d95ad58fb8934f7cd246512b703`  
Checked at: 2026-07-20 16:20:37 CST

jasonlyc (approved):

PR #9352, opened 2026-07-20
Add inactive ap/switch to init.hosts

xinnige:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The change is narrowly scoped to `getHostsAsync()` and only affects FireRouter-managed platforms.
- MAC validation is in place before adding configured APC assets to `visibleMACs`.
- Failure to read FireRouter config is contained with a `try/catch`, so host initialization will not fail outright if config fetch fails.
- The intent matches the current init path where `includeInactiveHosts: false` is used but pinned/DHCP-reserved hosts are still force-included.

### ⚠️ Issues found

- **Low:** No linked issue found. The PR body is empty and GitHub metadata/timeline did not expose a `Closes/Fixes` issue reference, so requirement coverage could only be inferred from the PR title.
- **Medium:** The implementation only adds APC asset MACs to the lookup set, but still depends on an existing `host:mac:<MAC>` Redis entry. If an inactive AP/switch has already aged out of host storage, `hgetall` returns empty and the later defensive check skips it, so the device still will not appear in `init.hosts`. If configured APC assets are meant to be authoritative, this needs either a fallback host object from asset config/status or a guarantee that managed AP/switch host records never expire.

### 💡 Suggestions

- Add a focused test around `getHostsAsync({ includeInactiveHosts: false })` where:
  - an inactive managed AP/switch exists in `apc.assets`
  - it is not returned by `getMACsByTime()`
  - it is still included when its `host:mac` record exists
  - behavior is explicit when the `host:mac` record is missing
- Consider extracting the APC asset MAC collection into a small helper to make the behavior testable without exercising the full host loading path.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9352  
Head SHA: `5a6bd8b6fd72e0f8437147205059f19dc6a96778`  
Checked at: 2026-07-20 17:45:42 CST

jasonlyc (approved):

PR #9356, opened 2026-07-21
Purple 1.983 Prod Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The change correctly avoids the disruptive `iptables-save` / `iptables-restore` / `rmmod` cycle when TLS modules are already loaded and match the bundled kernel/userland artifacts.
- Platform support checks for `xt_tls` and `xt_udp_tls` are kept inside `tlsModuleNeedsUpdate`, so unsupported modules are not accidentally queued.
- The comparison uses kernel module `srcversion` plus xtables `.so` checksum, which is a reasonable signal for deciding whether reload/install is needed.

### ⚠️ Issues found

[control/install_iptables_setup.sh:369]  
The new conditional update block still runs `iptables-restore` during `--dry-run` when any TLS module needs update. The script advertises dry-run as “no iptables-restore or ipset restore will be executed,” and the final restore path respects `DRY_RUN`, but this intermediate tls-clean restore path does not. Please gate these two temporary restores, and likely the `rmmod` / `installTLSModule` calls too, behind `DRY_RUN == false`, or make dry-run only report the modules that would be updated.

### 💡 Suggestions

### Issue requirement coverage

No linked issue found.

### Overall findings

- **Medium:** `--dry-run` can still mutate firewall/kernel-module state when TLS modules need update.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9356  
Head SHA: `3b1913ca302735d071007f695e8bde85fb7f65d9`  
Checked at: 2026-07-21 14:26:14 CST

PR #9362, opened 2026-07-24
ci: add release-on-merge workflow (auto tag per model/channel)

MelvinTo:
Adds `.github/workflows/release-on-merge.yml` — when a PR is **merged** into a tracked branch, it creates a GitHub release + tag.

## Tag schema
`<model>-<channel>-v<version>-<patch>` — e.g. `purple-alpha-v1.983-001`
- **`<model>-<channel>`** — hardcoded map from the merged branch (reverse of the `model:template` mapping; `release_6_0`=production, `beta_6_0`=beta, `beta_7_0`=alpha). E.g. `beta_13_0 → purple-alpha`, `release_9_0 → purple-production`.
- **`v<version>`** — `.version` from `net2/config.json` on the merged branch.
- **`<patch>`** — auto-increments (`001`, `002`, …) per `(model-channel, version)` by scanning existing tags; resets when the version bumps.

Restricted to the mapped branches via the `pull_request.branches` filter, and gated on `merged == true`. Includes a small retry loop to avoid patch collisions on near-simultaneous merges.

## ⚠️ Rollout note (important)
`pull_request` workflows execute the YAML from the **base branch** of the PR. So this file must exist on each target branch (`beta_*`, `release_*`) for merges there to tag — having it only on `master` will just cover master-targeted PRs. It will propagate as `master` is merged down, or we can fan it out to the branch set explicitly.

## Follow-ups to confirm
- Channel word is `production` (per "release_6_0 is production") — say if you'd prefer `release`/`prod`.
- Uses the default `GITHUB_TOKEN`, so the created tag will **not** trigger any `on: push: tags:` build. If a tag-triggered build is needed, switch to a PAT/App token.

### DIFF
--- a/.github/workflows/release-on-merge.yml
+++ b/.github/workflows/release-on-merge.yml
@@ -1,0 +1,99 @@
+name: Release on merge
+on:
+  pull_request:
+    types: [closed]
+    branches:
+      - release_7_0
+      - release_8_0
+      - release_9_0
+      - release_10_0
+      - release_11_0
+      - release_12_0
+      - release_13_0
+      - beta_8_0
+      - beta_9_0
+      - beta_10_0
+      - beta_11_0
+      - beta_12_0
+      - beta_13_0
+      - beta_14_0
+      - beta_15_0
+      - beta_16_0
+      - beta_17_0
+      - beta_18_0
+      - beta_19_0
+      - beta_20_0
+      - beta_21_0
+      - master
+
+permissions:
+  contents: write
+
+jobs:
+  release:
+    if: github.event.pull_request.merged == true
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          ref: ${{ github.event.pull_request.base.ref }}
+          fetch-depth: 0
+          fetch-tags: true
+
+      - name: Create release tag
+        env:
+          BRANCH: ${{ github.event.pull_request.base.ref }}
+          TARGET: ${{ github.event.pull_request.merge_commit_sha }}
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          set -euo pipefail
+
+          # actual branch -> "<model>-<channel>" (reverse of the model:template mapping;
+          # release_6_0=production, beta_6_0=beta, beta_7_0=alpha)
+          case "$BRANCH" in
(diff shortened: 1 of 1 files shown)


jasonlyc (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The workflow is correctly gated on `pull_request.closed` plus `github.event.pull_request.merged == true`, so closed-but-unmerged PRs will not create releases.
- Branch-to-label mapping is explicit and easy to audit.
- Tag generation is scoped by `<model>-<channel>-v<version>`, so patch numbers reset per version as intended.
- The retry loop addresses the main expected collision case when two merges compute the same next patch tag.
- `set -euo pipefail` and quoted shell variables are good defaults here.
- No linked issue found.

### ⚠️ Issues found

- **Medium: version can be read from the wrong commit under merge concurrency.**  
  `.github/workflows/release-on-merge.yml:29-34` checks out `${{ github.event.pull_request.base.ref }}`, while the release target is `${{ github.event.pull_request.merge_commit_sha }}`. If another PR lands on the same branch before this job reaches checkout, `net2/config.json` may be read from the newer branch head, but the release tag will point at the earlier merge commit. That can produce a tag with the wrong version prefix.  
  Recommended fix: checkout the merge commit SHA directly, or explicitly `git checkout "$TARGET"` before reading `net2/config.json`.

- **Low: retry loop treats every `gh release create` failure as a tag race.**  
  `.github/workflows/release-on-merge.yml:76-84` retries regardless of failure reason. Auth failures, invalid tag names, API errors, or release-note generation failures will be reported as “likely tag race,” which can hide the real problem. Consider checking the error output and only retrying known already-exists/collision cases, or at least logging the `gh` error clearly.

### 💡 Suggestions

- Validate `VER` against the expected version format before building the tag name. The current non-empty check prevents missing versions, but a malformed `.version` could still create unexpected release/tag names.
- Consider adding `concurrency` per target branch if release ordering matters operationally. The existing retry handles patch collisions, but it does not guarantee release numbering follows merge order.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9362  
Head SHA: `df2e80c62baf88aff89bbf6a3af51527729d6996`  
Checked at: 2026-07-24 11:49:51 CST

PR #9363, opened 2026-07-24
support goldmax

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds explicit GoldPlus2 platform loading and platform-specific service names, assets, profiles, and hooks.
- Handles kernel 6.6 IFB behavior by explicitly creating `ifb0` / `ifb1`.
- Makes CPU temperature handling more defensive when multi-CPU platforms return a missing/empty list.
- Updates NTP redirect logic to account for both `ntp` and `chrony`.

Issue requirement coverage: No linked issue found. The PR body is empty and no linked issue was visible from PR metadata.

### ⚠️ Issues found

**High** `[scripts/bootstrap.js:216]`  
`execSync('qrencode ... ' + JSON.stringify(url))` is still shell execution with attacker-controlled input if `FW_ACTIVATE_BASE` is set. Double quotes do not prevent shell expansion such as `$()` inside the URL. Use `execFileSync('qrencode', ['-t', 'UTF8', '--', url], ...)` or `spawnSync` with argv to avoid shell injection.

**Medium** `[sensor/NTPRedirectPlugin.js:94]`  
Adding a bare `allow` to `chrony.conf` permits NTP queries from all source addresses. The comment relies on WAN firewall behavior, but if firewall rules are delayed, disabled, or reordered during boot, chrony can answer externally. Prefer limiting `allow` to monitored LAN CIDRs, loopback, or generated local interface subnets.

**Medium** `[platform/goldplus2/GoldPlus2Platform.js:369]`  
`ledBooting()` calls `this.updateLEDDisplay(...)` without `await`. Because this is an async method, it returns before the LED update completes and the surrounding `try/catch` will not catch rejected errors. Use `await this.updateLEDDisplay({ boot_state: "booting" })`.

**Low** `[scripts/apply_profile.sh:317]` and `[scripts/apply_profile.sh:362]`  
`mapfile -t cpu_arr < <(parse_cpu_list "$cpu_spec") || cpu_arr=()` does not reliably catch `parse_cpu_list` failures because the failing function runs in process substitution. Invalid CPU specs can still produce partial output and continue pinning threads. Parse into a temp variable or validate the full CPU list before applying any `taskset`.

### 💡 Suggestions

- Consider adding a small test or dry-run validation for `threaded_napi` profile parsing, since malformed CPU placement can silently degrade packet processing.
- Consider centralizing the duplicated non-glob/glob NAPI pinning logic in `apply_profile.sh`; the two branches are nearly identical and will be easy to drift.
- For `chrony`, make config edits idempotent with a managed block or generated include file instead of appending lines directly to `/etc/chrony/chrony.conf`.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9363  
Head SHA: `c1d3024db01e3b0c8a474c180f66e161773eab16`  
Checked at: 2026-07-24 15:47:04 CST

MelvinTo (approved):

PR #9364, opened 2026-07-24
fix bug, failed to read udp_tls module version

caixingcheng:
Fixed issue could not read version when the udp_tls module was located in the system's default path.

### DIFF
--- a/net2/KernelCrashMonitor.js
+++ b/net2/KernelCrashMonitor.js
@@ -79,13 +79,28 @@ async function waitForLockReleaseAndRefreshCache() {
   cachedShouldDisableUdpTls = crashInfo.shouldDisableUdpTls === true;
 }
 
-// parse "version:" and "srcversion:" lines from modinfo output
-async function getModuleVersion(koPathOrName) {
-  const result = await execFile('modinfo', [koPathOrName]).catch((err) => {
-    log.error("Failed to run modinfo for", koPathOrName, err.message);
+// parse "version:" and "srcversion:" lines from modinfo output.
+// try `modinfo koPath` first; when that fails (koPath may not exist yet), fall
+// back to `modinfo modName` so the version can still be resolved from an
+// already-loaded/installed module by name.
+async function getModuleVersion(modName, koPath) {
+  let result = null;
+  if (koPath) {
+    result = await execFile('modinfo', [koPath]).catch((err) => {
+      log.debug("Failed to run modinfo for", koPath, err.message);
+      return null;
+    });
+  }
+  if (!result && modName) {
+    result = await execFile('modinfo', [modName]).catch((err) => {
+      log.debug("Failed to run modinfo for", modName, err.message);
+      return null;
+    });
+  }
+  if (!result) {
+    log.warn("Failed to get module version for", modName, koPath);
     return null;
-  });
-  if (!result) return null;
+  }
 
   let version = '';
   let srcversion = '';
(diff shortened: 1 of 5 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The JS call path now carries both module name and optional `.ko` path, which addresses the PR goal of resolving `xt_udp_tls` when it lives in the system module path.
- `getModuleVersion()` correctly tries the explicit path first, then falls back to `modinfo xt_udp_tls`.
- Existing kernel crash monitor tests were updated, and a focused fallback test was added.
- PR metadata/body does not link a GitHub issue. No linked issue found.

### ⚠️ Issues found

- Low: `[platform/platform.sh:285]` The new `jq -c` pipeline moved `2>/dev/null` to the second `jq` only. If `crash_info` contains invalid JSON, parse errors from the first `jq` can now leak to stderr/logs, whereas the previous command suppressed them. Prefer redirecting the whole pipeline or the first `jq` as well.

### 💡 Suggestions

- Add a test for `checkPstoreAndUpdateRedis('xt_udp_tls', null)` or a failing `koPath` path to cover the startup path fixed by this PR, not only `onUdpTlsModuleLoaded()`.
- The comment saying fallback can resolve from an “already-loaded” module is slightly misleading because `modinfo <name>` resolves module metadata from installed module files, not from kernel memory. Consider wording it as “installed module by name.”

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9364  
Head SHA: `de7a040745ade6813424e2d6906b04c619f3a7ce`  
Checked at: 2026-07-24 16:40:20 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The JS call path now passes both `modName` and optional `koPath`, which fixes the system-default module path case described by the PR.
- `getModuleVersion()` preserves the preferred path lookup first, then falls back to `modinfo xt_udp_tls`.
- `net2/main.js` no longer passes the module name as a fake path, making the API contract clearer.
- Existing unit tests were updated, and a focused fallback test was added.
- Linked issue coverage: `firewalla/firecommit#8680` requires pstore-based udp_tls crash detection and disabling. This PR appears to support that flow by fixing version lookup when the module is installed in the default path.

### ⚠️ Issues found

- Low: `platform/platform.sh:285` changes the `jq` pipeline so `2>/dev/null` now applies only to the first `jq` command before piping into `jq -c`. If the second `jq` fails or receives unexpected output, stderr may leak where the previous single-command form suppressed errors. Prefer redirecting the whole pipeline or folding compact output into the first `jq` invocation.

### 💡 Suggestions

- Add a unit test for `checkPstoreAndUpdateRedis('xt_udp_tls', null)` or a failing `koPath` during startup. The new fallback test covers `onUdpTlsModuleLoaded()`, but the PR’s main startup path also depends on this behavior.
- The comment saying fallback can resolve from an “already-loaded/installed module” is slightly misleading. `modinfo <name>` reads installed module metadata, not metadata from the currently loaded kernel module. “installed module by name” would be more precise.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9364  
Head SHA: `05b900ff0491d1b0d5cdc16bfe26889dde1e4d70`  
Checked at: 2026-07-24 17:07:06 CST

jasonlyc (approved):

PR #9366, opened 2026-07-24
support live throughput on phyIntf and wlanIntfs

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good
- Adds `phyIntf` throughput using the existing `/sys/class/net/*/statistics` sampling path.
- Validates externally supplied physical interface names before reading sysfs.
- Adds Wi-Fi AP discovery for both integrated AP assets and local `iw dev`.
- Filters managed/STA Wi-Fi interfaces so Wi-Fi WAN and scan interfaces are not reported as LAN AP throughput.
- Includes focused unit coverage for parser behavior, integrated AP discovery, and platform path selection.

### ⚠️ Issues found
- **Medium** `[sensor/LiveStatsPlugin.js:277]`  
  The linked issue asks for new types `phyIntf` and `wlanIntf`, but the request dispatch only accepts `wlanIntfs`. The returned rows use `type: 'wlanIntf'`, so callers following the issue/API requirement and sending `type: 'wlanIntf'` will get no Wi-Fi throughput. Please either support `case 'wlanIntf'` or confirm/update the API contract and frontend caller to use the plural type.

- **Low** `[test/test_livestats_wlan.js:78]`  
  The test mutates shared module singletons (`fwapc.getAssetsStatus`, `platform.hasIntegratedAPAssets`) and never restores them. Since these are cached `require()` instances, later tests in the same Mocha process can inherit these mocks depending on execution order. Capture originals in `beforeEach`/`before` and restore them in `afterEach`/`after`.

### 💡 Suggestions
- Add a small test for the top-level throughput dispatcher so the accepted query type is locked down, especially for `phyIntf` and Wi-Fi.
- Consider rejecting `.` / `..` in `validIntfName()` even though slash injection is blocked. It keeps the guard semantically aligned with “real interface name exists” rather than “some sysfs path exists.”

### Verdict
COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9366  
Head SHA: `4908a7799bd08710f34c0d6b43aaebc0f05bf0ef`  
Checked at: 2026-07-24 18:42:05 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds the requested live throughput support for `phyIntf` and Wi-Fi interfaces.
- Validates interface names before reading `/sys/class/net`, which is the right direction for the new `phyIntf` path.
- Wi-Fi AP parsing handles multi-BSS hostapd output correctly, including selecting the SSID for the queried interface.
- Tests cover AP, Wi-Fi WAN, idle station, escaped SSIDs, band mapping, and interface enumeration.

### ⚠️ Issues found

[sensor/LiveStatsPlugin.js:817]  
`validIntfName()` currently accepts `.` and `..` because the regex allows dot-only names. That means requests like `type=phyIntf,target=..` can pass the existence check against `/sys/class/net/..`, even though they are not real NIC names. Please explicitly reject `.` and `..` before `fsp.access()`, or resolve and verify the final path remains an actual entry under `/sys/class/net`.

[test/test_livestats_wlan.js:144]  
The test file monkey-patches `fireRouter.getWanIntfNames` but never restores it. Since `fireRouter` is a shared module singleton, this can leak into later tests in the same process and cause order-dependent failures. Please save the original method and restore it in `afterEach`/`after`.

### 💡 Suggestions

- Consider using `execFile`/`spawn` argument arrays for `find` and `wpa_cli` instead of shell-interpolated strings. Interface names are validated, but avoiding shell parsing here would make the command construction safer and less sensitive to runtime path formatting.
- Add a focused test for invalid `phyIntf` targets such as `.`, `..`, and names with disallowed characters.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9366  
Head SHA: `9c6ad16130c6807bef3371b94c7d0bdc69470ee0`  
Checked at: 2026-07-29 16:37:30 CST

jasonlyc (approved):

PR #9369, opened 2026-07-27
sync gold plus2 branches

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The `KernelCrashMonitor` API direction is cleaner: separating `modName` from `koPath` avoids overloading one argument with two meanings.
- The fallback from `modinfo <koPath>` to `modinfo <modName>` addresses the likely runtime case where the `.ko` path is not present yet but the module can still be resolved by name.
- Tests were updated broadly for the new signature, and a focused fallback test was added.
- `goldplus2` bootstrap platform detection looks consistent with existing platform loader naming.

### ⚠️ Issues found

- **Medium:** `api/bin/www:77` still calls `checkPstoreAndUpdateRedis(udpTlsKoPath)` using the old one-argument contract. With the new signature, that path is interpreted as `modName`, leaving `koPath` unset, so the new fallback behavior is skipped for FireApi startup. If `getTlsKoPath('xt_udp_tls')` returns a missing `.ko` path, FireApi will still only run `modinfo <path>` and will not retry `modinfo xt_udp_tls`. This partially defeats the PR’s main fix outside `net2/main.js`.

- **Low:** The new fallback test covers `onUdpTlsModuleLoaded`, but not `checkPstoreAndUpdateRedis`, which is the startup path where this bug matters most. The missed `api/bin/www` caller would likely have been caught by a startup-path test.

### 💡 Suggestions

- Update `api/bin/www` to match `net2/main.js`:
  `checkPstoreAndUpdateRedis('xt_udp_tls', udpTlsKoPath)` and use `null` on `getTlsKoPath` failure.
- Add a test where `checkPstoreAndUpdateRedis('xt_udp_tls', '/lib/modules/xt_udp_tls.ko')` falls back to `modinfo xt_udp_tls` after the path lookup fails.
- No linked issue found in the PR title/body/comments, so issue requirement coverage could not be verified.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9369  
Head SHA: `5d1c57e518ffa335555edb90b7d06ef38276a488`  
Checked at: 2026-07-27 11:57:04 CST

PR #9371, opened 2026-07-27
Add backoff radius image download

xinnige:
https://github.com/firewalla/firecommit/issues/8888

### DIFF
--- a/control/CategoryUpdater.js
+++ b/control/CategoryUpdater.js
@@ -1096,7 +1096,7 @@ class CategoryUpdater extends CategoryUpdaterBase {
 
     const sigCfg = this.getSignatureConfig(sigId);
     if (!sigCfg || !sigCfg.categories || !_.isArray(sigCfg.categories) || !sigCfg.categories.includes(category)) {
-      log.info(`Signature ID ${sigId} is not found or not matched with signature config, skip adding sig detected server ${sigEntry.id} to category ${category}`);
+      log.info(`Signature ID ${sigId} is not found or not matched with signature config, skip adding sig detected server to category ${category}`);
       return;
     }
     let serverEntry  = this.composeSigDetectedServerEntry(remoteAddr, protocol, remotePorts, sigId);
--- a/extension/freeradius/freeradius.js
+++ b/extension/freeradius/freeradius.js
@@ -29,6 +29,7 @@ const platform = PlatformLoader.getPlatform()
 const f = require('../../net2/Firewalla.js');
 const fr = require('../../net2/FireRouter.js');
 const log = require('../../net2/logger.js')(__filename);
+const rclient = require('../../util/redis_manager.js').getRedisClient();
 const util = require('../../util/util.js');
 const { Rule } = require('../../net2/Iptables.js');
 const iptc = require('../../control/IptablesControl.js');
@@ -40,6 +41,11 @@ const configDir = `${f.getUserConfigFolder()}/freeradius`
 const logDir = `${f.getUserHome()}/.forever/freeradius`
 const certsDir = `${f.getHiddenFolder()}/certs/freeradius`
 const fwrcFile = `${f.getUserHome()}/.fwrc`
+
+// shared with scripts/update_freeradius.sh so cron and firemain honor the same backoff window
+const PULL_BACKOFF_KEY = "freeradius:image:pull:backoff";
+// local disk / overlayfs faults: back off
+const DISK_FAULT_RE = /(failed to extract layer|failed to convert whiteout|failed to register layer|operation not permitted|no space left on device|input\/output error|read-only file system)/i;
 
 const sleep = (ms) => new Promise(resolve => setTimeout(resolve, ms));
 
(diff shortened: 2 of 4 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds a shared Redis backoff key so `firemain` and `scripts/update_freeradius.sh` can both avoid repeated FreeRADIUS image pulls.
- Distinguishes likely local disk / overlayfs failures from transient pull failures and uses longer backoff for disk faults.
- Avoids restarting the FreeRADIUS container when the requested image is missing, which directly addresses the churn described in the issue.
- Adds focused unit coverage for disk-fault backoff, transient backoff, skip behavior, clearing on success, delay capping, and Redis-only cross-process state.

### ⚠️ Issues found

**Medium** `[extension/freeradius/freeradius.js:722]`  
`_recordPullFailure()` only increments attempts from `this.pullBackoff`, not from the Redis state. If a fresh `firemain` process records a new failure after cron/script already persisted `{attempts: N}`, or after an expired Redis backoff still exists, JavaScript resets the sequence to attempt `1`. That breaks the intended shared exponential backoff between cron and firemain.

Suggested fix: read and parse the existing Redis state in `_recordPullFailure()` when `this.pullBackoff` is absent, then increment from that value.

**Low** `[scripts/update_freeradius.sh:136]`  
When the script skips a pull due to backoff and the current image already exists, the later comparison prints `image ... is up to date`. That is misleading: the script did not verify the remote image and only confirmed the local image ID did not change. This matches the issue comment’s sample output, but it can confuse operations/debugging.

Suggested fix: track `pull_skipped=true` and print something like `using existing local image due to image-pull backoff` instead of `is up to date`.

### 💡 Suggestions

- Consider adding one test for JS backoff continuity from an existing Redis record, for example Redis has `attempts: 3` and `_recordPullFailure()` should persist `attempts: 4`.
- Consider including `lastError` in the shell-side Redis state too, matching the JS side, since it is useful for diagnosing repeated device-local failures.
- Issue requirement coverage:
  - **Met:** Detects the reported overlayfs whiteout failure (`failed to convert whiteout ... operation not permitted`).
  - **Met:** Prevents repeated pull attempts during a backoff window from both JS and cron script paths.
  - **Met:** Avoids container restart when the target image cannot be pulled.
  - **Partially met:** Shared exponential backoff is not fully consistent because JS does not increment from existing Redis attempts.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9371  
Head SHA: `b146c77cf4647faa76b524bcabc4b7d3dca52744`  
Checked at: 2026-07-27 16:16:14 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Adds shared Redis-backed pull backoff so `firemain` and `scripts/update_freeradius.sh` can avoid repeated FreeRADIUS image downloads.
- Detects the reported overlayfs whiteout failure pattern from linked issue `firewalla/firecommit#8888`.
- Uses longer backoff for likely disk/local overlayfs faults and shorter backoff for transient pull errors.
- Skips container restart when the requested image is unavailable, avoiding churn after a failed/backed-off pull.
- Adds focused tests for JS backoff behavior, skip behavior, success clearing, delay capping, and Redis-only backoff reads.
- Fixes the unrelated `CategoryUpdater.js` undefined `sigEntry` log reference.

### ⚠️ Issues found

**Medium** `[extension/freeradius/freeradius.js:722]`

`_recordPullFailure()` only increments attempts from `this.pullBackoff`. If a fresh `firemain` process records a failure after cron/script has already persisted `{attempts: N}` in Redis, JS resets the sequence back to attempt `1`. That weakens the shared exponential backoff between cron and firemain.

Suggested fix: when `this.pullBackoff` is absent, read and parse the existing Redis state before calculating `attempts`.

**Low** `[scripts/update_freeradius.sh:136]`

When the script skips `docker pull` because of backoff, the later image comparison can still print that the image “is up to date.” That is operationally misleading because the script did not check the remote image, only reused the local image.

Suggested fix: track a `pull_skipped=true` flag and print a distinct message such as `using existing local image due to image-pull backoff`.

### 💡 Suggestions

- Add a unit test where Redis already contains `attempts: 3`, `freeradius.pullBackoff` is null, and `_recordPullFailure()` should persist `attempts: 4`.
- Consider storing `lastError` from the shell script too, matching the JS side for easier debugging.

Issue requirement coverage:
- **Met:** Prevents repeated FreeRADIUS image pulls after the reported disk/overlayfs failure.
- **Met:** Detects the reported `failed to convert whiteout ... operation not permitted` error.
- **Met:** Shares a backoff key between cron script and firemain paths.
- **Met:** Avoids restarting the container when the target image cannot be pulled.
- **Partially met:** Shared exponential backoff is inconsistent because JS does not continue from Redis `attempts`.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9371  
Head SHA: `46c830294a74ac04f7dc31785af2c4a107d37663`  
Checked at: 2026-07-27 16:39:46 CST

jasonlyc (approved):

PR #9372, opened 2026-07-27
sync gold 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- No linked issue found.
- The PR adds useful targeted tests for `/v1/me`, dnsmasq reload behavior, KernelCrashMonitor, and FreeRadius pull backoff.
- Several shell-exec paths moved toward `execFile`, which improves safety and reduces shell overhead.
- `/me` SPA path handling keeps segment decoding and traversal checks from the existing `time_limits` approach.

### ⚠️ Issues found

**Medium: release workflow misses the new Gold Plus 2 release branches**

[.github/workflows/release-on-merge.yml:6]

This PR adds Gold Plus 2 branch mappings in `scripts/fireupgrade.sh` for `release_14_0`, `beta_22_0`, and `beta_23_0`, and those branches already exist, but the new release workflow trigger and label map stop at Orange (`release_13_0`, `beta_21_0`). Merges into the Gold Plus 2 release/beta branches will not create releases/tags.

Please add those branches to both `on.pull_request.branches` and the `case "$BRANCH"` label mapping, for example `goldplus2-production`, `goldplus2-beta`, and `goldplus2-alpha`.

**Medium: FreeRadius JS backoff loses persisted attempt count after process restart**

[extension/freeradius/freeradius.js:724]

`_recordPullFailure()` increments from `this.pullBackoff.attempts` only. If firemain restarts, the in-memory state is empty; `_inPullBackoff()` reads Redis only to decide whether the window is active, but it never loads the persisted `attempts`. After the Redis backoff expires, the next JS-side failure records attempt `1` again, so exponential backoff does not survive process restarts even though the state is persisted for cron/cross-process coordination.

Please parse the Redis state and seed `attempts` from it before calculating the next delay, matching the shell script behavior.

**Low: `/v1/me` can spawn many ping processes per request**

[api/routes/me.js:218]

The topology is cached, but latency is recalculated on every request with `Promise.all()` across all hops, and each hop runs `ping -c 5`. Any local client can repeatedly hit `/v1/me` and cause bursts of ping subprocesses. On larger topologies this can become noisy or expensive.

Consider limiting concurrency, capping hop count, reducing ping count, or caching latency separately with a short TTL.

### 💡 Suggestions

- Add a release-workflow unit/smoke check or script-generated branch map so `fireupgrade.sh` and `release-on-merge.yml` cannot drift independently.
- Avoid returning raw internal exception messages from `/v1/me` 500 responses; log the details server-side and return a generic error body to clients.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9372  
Head SHA: `6346a5480451c34c4148ecc3a2b33dada38a6b2e`  
Checked at: 2026-07-27 18:31:49 CST

PR #9387, opened 2026-07-30
fix bugs for tls adblock when apply to all but exclude some

yijin-png:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The ACL audit log change correctly identifies adblock TLS hits, records them as adblock hits, skips unrelated policy backtrace, and removes the reserved pid from the emitted record.
- The new chain hierarchy is the right direction for supporting inherited scopes with exclusions: global → network → group → device/identity → mark.
- The refresh queue/debounce is a good safeguard against interleaved iptables rewrites.

### ⚠️ Issues found

- **High** `[sensor/AdblockPlugin.js:728]` Stale TLS entry rules can survive when a scope is reset or deleted. `_syncTlsEntryRules()` only deletes entry rules for scopes currently returned by `_collectTlsScopes()`. But reset/delete paths can remove scope keys before the delayed refresh runs, for example deleted tags are reset then removed at lines 571-578, and missing networks/identities are deleted at lines 580-590. In those cases, the old `FW_FIREWALL_*_BLOCK -> FW_ADBLOCK_*` jump is no longer generated in either `enabled` or `rest`, so it is never deleted. This can keep adblock TLS active for a removed/reset device/tag/network/identity and can also leave references to `FW_ADBLOCK_*` chains when `_disableTlsBlock()` destroys those chains.

- **Medium** `[sensor/AdblockPlugin.js:671]` `_disableTlsBlock()` destroys the custom chains after deleting only currently known scope entry rules. If stale entry rules exist, iptables restore may fail because built-in chains still jump to deleted custom chains. This is the same lifecycle problem as above, but the failure mode is broader: disabling strict mode/admin switch may not cleanly remove the ruleset.

  - `Met`: TLS adblock audit records report `reason=adblock` and remove the reserved pid before normal rule attribution.
  - `Partially met`: “Apply to all but exclude some” is handled for currently known scopes, but reset/deleted scopes can leave stale entry rules, so exclusions/removals are not guaranteed to take effect over time.

### 💡 Suggestions

- Track previously installed TLS entry scopes separately, or compute deletions from the existing iptables state rather than only from current settings.
- For reset/delete paths, pass the concrete target into the refresh/removal path before deleting it from the settings map, or keep tombstones until the queued refresh completes.
  - global on + device/tag/network excluded
  - per-device on → reset
  - tag deleted after being enabled/excluded
  - strict mode off after scoped TLS rules were previously installed

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9387  
Head SHA: `3df02f32682338f45cb1fe84d7ae057ea5cb278f`  
Checked at: 2026-07-30 18:46:48 CST

PR #9398, opened 2026-08-03
monitor RSPAN VLAN traffic in Zeek with dedup

jasonlyc:
- create ifb_pcap_rspan ifb device; redirect RSPAN VLAN ingress to it via tc mirred in FireRouter
- add ifb_pcap_rspan (or physical parent VLANs) to Zeek listen set in PcapPlugin; assign worker-rspan id in BroControl
- extend BroDetect same-network dedup to handle RSPAN as a third source alongside zeek and AP/switch bridge, using stpPort, wireless device status (isWirelessDevice), and LRU 5-tuple dedup
- track last-seen-on-wifi timestamp in Host/APCMsgSensor to support wireless device detection in dedup logic

### DIFF
--- a/net2/BroControl.js
+++ b/net2/BroControl.js
@@ -24,6 +24,7 @@ const Promise = require('bluebird');
 const fs = require('fs');
 Promise.promisifyAll(fs);
 const _ = require('lodash');
+const Constants = require('./Constants.js');
 const platform = require('../platform/PlatformLoader.js').getPlatform();
 const BRO_PROC_NAME = platform.getBroProcName();
 
@@ -75,9 +76,10 @@ class BroControl {
       // subsequent filters in restrict_filters will be applied properly
       if (intfType === "1")
         workerScript.push(`redef capture_filters += [["shift-vlan-offset"] = "(not ether proto 0x8100) or (ether proto 0x8100 and vlan)"];`);
+      const workerId = intf === Constants.INTF_PCAP_RSPAN ? "worker-rspan" : `worker-${index++}`;
       workerCfg.push(
         `\n`,
-        `[worker-${index++}]\n`,
+        `[${workerId}]\n`,
         `type=worker\n`,
         `host=localhost\n`,
         `interface=${intf}\n`,
--- a/net2/BroDetect.js
+++ b/net2/BroDetect.js
@@ -88,6 +88,7 @@ const { getUniqueTs, extractIP } = require('./FlowUtil.js')
 
 const LRU = require('lru-cache');
 const Constants = require('./Constants.js');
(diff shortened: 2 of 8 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Clean separation of the RSPAN capture path using a dedicated IFB device and Zeek worker ID.
- RSPAN configuration changes correctly trigger pcap restart and filter refresh.
- Removed interfaces have their stale `clsact` qdisc cleaned up.
- Wireless-aware source preference is clearly documented.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/BroDetect.js:1077`: Deduplication fails when sources observe opposite directions.**  
  The key is directional (`orig → resp`). The PR notes that RSPAN may capture only one direction; in that case Zeek can assign the RSPAN record’s originator opposite to the AP/bridge record. The two copies then generate different keys, bypass the LRU, and are both retained. Canonicalize the endpoint/port pairs for deduplication, while preserving protocol, or explicitly test both forward and reverse keys.

- **Medium — `net2/BroDetect.js:1008`: VLAN IDs are treated as globally identifying an RSPAN source.**  
  `getRspanVids().has(obj.vlan)` classifies any observed packet carrying the same VID as RSPAN, regardless of its capture interface. VLAN IDs can be reused on different parent interfaces, so ordinary Zeek traffic may be misclassified and dropped by the RSPAN rules. Prefer the dedicated `worker-rspan` identity; if a VID fallback is required, qualify it with the capture interface or parent interface.

- **Medium — `net2/Host.js:1365`: Wireless classification expires after two minutes without proving the device moved to Ethernet.**  
  A quiet but still-associated wireless client becomes “wired” after the TTL. If an RSPAN record arrives first, it can win the LRU and cause the later authoritative AP flow to be dropped. Track current AP association/disassociation state, or use a lifetime aligned with guaranteed APC refresh behavior.

### 💡 Suggestions

- Add focused tests covering reversed RSPAN/AP tuples, duplicate VLAN IDs on different parent interfaces, quiet wireless clients, RSPAN configuration removal, and repeated `resetPcapRspan()` calls.
- Use `Constants.WORKER_ID_RSPAN` in `BroControl.js` instead of duplicating the `"worker-rspan"` literal.
- Consider `tc filter replace` with an explicit preference/handle to make filter refresh idempotent.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9398  
Head SHA: `0c66d47ca562831948ec1ef54328899611f21a11`  
Checked at: 2026-08-03 17:21:37 CST

MelvinTo (approved):

PR #9415, opened 2026-08-05
fix bug, failed to detect module version on orange platform

caixingcheng:
problem is because of MODULE_VERSION was removed from the new kernel, fix is use the buildid instead.

### DIFF
--- a/control/install_iptables_setup.sh
+++ b/control/install_iptables_setup.sh
@@ -304,10 +304,7 @@ cat "$qos_file"
   echo 'COMMIT'
 } >> "$ip6tables_file"
 
-# Return 0 (need install/update) only when the module is supported on this platform AND
-# it is not loaded yet, or the loaded ko srcversion / installed .so checksum differs from
-# the bundled version. Return 1 otherwise so we can skip the disruptive rmmod/restore cycle.
-function tlsModuleNeedsUpdate() {
+function tlsModuleSupported() {
   local module_name=$1
   if [[ ${module_name} = "xt_tls" && ${XT_TLS_SUPPORTED} != "yes" ]]; then
     return 1
@@ -315,6 +312,17 @@ function tlsModuleNeedsUpdate() {
   if [[ ${module_name} = "xt_udp_tls" && ${XT_UDP_TLS_SUPPORTED} != "yes" ]]; then
     return 1
   fi
+  return 0
+}
+
+# Return 0 (needs a kernel module (re)load) only when the module is supported on this
+# platform AND it is not loaded yet, or the loaded ko differs from the bundled one.
+# Return 1 otherwise so we can skip the disruptive rmmod/restore cycle.
+# The userspace .so is checked separately by tlsSoNeedsUpdate: installing it does not
+# require reloading the module, so a .so-only change must not trigger the cycle.
+function tlsKoNeedsUpdate() {
+  local module_name=$1
+  tlsModuleSupported "${module_name}" || return 1
 
   # not loaded yet -> needs install
   if ! lsmod | grep -wq "${module_name}"; then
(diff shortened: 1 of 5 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Cleanly separates kernel-module reloads from userspace `.so` updates, avoiding unnecessary disruption.
- Adds build-ID support for kernels without `MODULE_VERSION` or `srcversion`.
- Handles compressed and unusually named module files.
- Symlink and target mtimes are both considered.
- KernelCrashMonitor tests cover compatibility with legacy Redis records, missing identities, rebuilds, and symlink changes.

### ⚠️ Issues found

- **High — Linked issue requirement is not fully satisfied.**  
  Issue `firewalla/firecommit#8679` requires firewall initialization to remain intact when a TLS module fails to load. The reload path still removes TLS rules, unloads the existing module, attempts installation, and then performs the full restore. If the new module fails to load, the final restore can still fail on `-m tls`/`-m udp_tls`, leaving the ruleset incomplete. The issue’s own validation notes acknowledge this residual failure mode. Either isolate failed reloads with rollback/recovery, or remove the closing linkage and explicitly scope this PR to reducing unnecessary reloads.

- **Medium — Different identity types are compared as if directly comparable.**  
  [`net2/KernelCrashMonitor.js`, `isSameIdentity()`] checks `koId` whenever both values are non-empty, without verifying their prefixes match. For example, `srcversion:ABC` versus `buildid:XYZ` is treated as a confirmed module change even when both records contain the same `srcversion`. Compare `koId` only when both use the same identity type; otherwise fall back to the next common field.

- **Medium — A corrupt module can still receive a SHA-256 identity.**  
  [`scripts/tls_module_id.sh`, `file_ids()`] validates the decompressed image using `vermagic=`, but if validation fails it still hashes the original file. Consequently, `describe` can identify a truncated or corrupt module as `sha256:*`, and KernelCrashMonitor may clear its safety disable because that hash differs from the stored identity. Restrict the SHA-256 fallback to module images that passed validation, or ensure an unvalidated hash cannot re-enable UDP TLS.

### 💡 Suggestions

- Add an integration test where a genuinely required reload fails and verify that the previous module/ruleset remains usable.
- Add unit tests for mixed `koId` types and corrupt/truncated module files.
- Document whether SHA-256 represents a validated module image or merely arbitrary file contents.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9415  
Head SHA: `91757e94cd1bcd4c5911fb28b64e06d08945f083`  
Checked at: 2026-08-05 11:41:56 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Separates kernel-module reloads from userspace `.so` updates, reducing disruptive iptables restores.
- Uses build IDs when `version` and `srcversion` are unavailable on Orange.
- Safely handles compressed, unusually named, symlinked, and corrupt module files.
- Maintains compatibility with legacy Redis identity records.
- Adds strong unit coverage for identity comparison, symlink timestamps, corrupt modules, and crash-disable behavior.
- Reviewed all 5 changed files.

### ⚠️ Issues found

- **High — Linked issue only partially satisfied.** `firewalla/firecommit#8679` requires firewall initialization to remain intact when TLS module loading fails. This PR reduces unnecessary reloads, but when a reload is genuinely required it still removes TLS rules, unloads the working module, ignores a failed `installTLSModule`, and restores rules requiring the unavailable match. That can still leave the firewall ruleset incomplete. Add rollback/recovery for failed reloads, or clarify that this PR mitigates frequency rather than fully resolving the linked issue.

- **Medium — Symlink mtime can suppress a genuine crash.** `getModuleFileMtimeSec()` treats the newer of the target and symlink mtimes as the module installation time. Recreating or touching a symlink—even if it still points to the same module—after a crash makes that crash appear to predate an upgrade. Only use the symlink timestamp when the target identity demonstrably changed, or persist the resolved target/identity for comparison.

### 💡 Suggestions

- Add an integration test where a required `insmod`/`modprobe` fails and verify the previous module and firewall rules remain usable.
- Add a test for recreating a symlink to the same target after a crash.
- Propagate and check failures from `installTLSSharedObject`, `installTLSModule`, and the intermediate/final `iptables-restore` operations.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9415  
Head SHA: `1da23cacad272e8018ce0dc43ef1c034dbf8d1b0`  
Checked at: 2026-08-05 14:46:16 CST

jasonlyc (approved):

PR #9419, opened 2026-08-05
sync gold 1.983 alpha staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Access requests have explicit pending/approved/denied/expired states and archival TTL handling.
- Time-limit matching accounts for user and affiliated tags, multiple apps, schedules, and disabled policies.
- CIDR exception matching now supports both IPv4 and IPv6.
- Multi-target disturb policies separate per-target QoS configuration and QUIC handling.

### ⚠️ Issues found

[alarm/AccessRequestManager.js:487]  
**High — Approval can partially mutate policies before validation fails.** App validation occurs only after matching policies have already had `extraQuota` updated and reenforcement queued. For example, `app === "all"` passes matching, modifies multiple policies, then fails the supported-app check at line 506. The request remains pending despite quota changes being applied. Validate the app list and approved quota before performing any policy updates, and make the operation transactional or rollback-safe.

[alarm/AccessRequestManager.js:302]  
**High — Request creation is not atomic.** Concurrent calls for the same `(userId, app)` can both observe a missing lookup, create separate request IDs, and add both to the pending set. Only the last lookup value remains reachable for updates. Use a Redis transaction/Lua script or an atomic `SET NX` claim covering the lookup, request record, and pending-set membership.

[alarm/AccessRequestManager.js:434]  
**High — Concurrent approvals can approve the same request twice.** The pending-state check and transition are separate Redis operations. Two callers can both observe `pending`, update policies, and independently create bypass policies before either archives the request. Atomically claim the request for approval before applying side effects, with a recoverable intermediate state if policy enforcement fails.

[alarm/AccessRequestManager.js:223]  
**Medium — The exported function contradicts its documented optional-app contract.** `findMatchingTimeLimitRules(userId)` throws because it unconditionally calls `app.split(',')`, although its documentation says omitting `app` should return all matching policies. Default `app` to `"all"` or reject omission explicitly and update the contract.

### 💡 Suggestions

- Add concurrency tests for duplicate request submission and simultaneous approval.
- Add failure-injection tests proving that unsupported apps and policy-update failures cannot leave partially applied quota.
- Add coverage for the documented omitted-app behavior and the `"all"` approval path.
- No linked issue found in the PR title, body, or comments, so issue requirement coverage could not be verified.
- This PR changes 427 files across many subsystems; splitting unrelated staging changes would materially improve reviewability and regression isolation.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9419  
Head SHA: `486d4ee7d6bfb0d9079e03edb1d00f2f19e9abfc`  
Checked at: 2026-08-05 17:14:26 CST

PR #9429, opened 2026-08-06
Orange 1.983 Alpha Patch

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Avoids disruptive `iptables-save`/restore and module reloads when TLS artifacts are unchanged.
- Handles unloaded and unhealthy modules explicitly.
- Reduces persistent Socket.IO retry load and improves reconnect diagnostics.
- Fixes the leaked `checkId` interval.
- Limits session expiration to non-MSP sessions and gives pre-existing sessions a fresh TTL.
- No linked issue found.

### ⚠️ Issues found

- **Medium — build-ID comparison can silently miss changed modules**  
  `[control/install_iptables_setup.sh:305]`  
  The parser assumes every GNU build ID is exactly 20 bytes and uses a hard-coded note header containing `descsz=0x14`. GNU build IDs may use other lengths. When parsing fails, `ko_buildid` is empty and the function eventually reports that no update is needed. Consequently, a changed module with a non-20-byte build ID can remain loaded indefinitely. Parse the ELF note according to its declared descriptor size—preferably using `readelf -n`—and treat an unavailable or uncomparable identity conservatively.

- **Medium — comparison failures are treated as successful matches**  
  `[control/install_iptables_setup.sh:337]`  
  If `get_tls_ko_path`, `modinfo`, `/sys/module/.../srcversion`, or build-ID extraction fails, the function can return “no update needed” without establishing that the loaded and bundled modules are identical. This is especially risky during packaging or platform transitions. Distinguish “equal” from “unable to compare”; log the latter and reinstall, or use another reliable artifact identity.

### 💡 Suggestions

- Add shell tests covering matching and mismatching `srcversion`, absent `srcversion`, varying GNU build-ID lengths, missing `.ko`, missing `.so`, and command failures.
- Add Guardian tests using fake timers for legacy-session migration, activity refresh, the 24-hour boundary, MSP exemption, and interval cleanup.
- Consider awaiting `touchActiveTs()` in `onMessage()` if preserving activity is lifecycle-critical; the current fire-and-forget write could race with shutdown or process termination.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9429  
Head SHA: `0a117457d93ca062a6735c96f6651951612dd940`  
Checked at: 2026-08-06 15:42:49 CST

PR #9431, opened 2026-08-06
category: fix stale target-list hashset polling

wzjjack:
Fixes firewalla/firecommit#9107

Boxes were requesting MSP target-list hashsets (`app.TL-<uuid>`) that fishbone no longer authorized them for, forever — logged server-side as `Unauthorized hashset request`, growing daily across 78+ boxes.

## Root cause

Once a `TL-<uuid>` category is activated (via a rule or an exception muting alarms on it), nothing ever deactivates it:
- Deleting the rule/exception that referenced it never called `deactivateCategory()`, so `CategoryUpdateSensor` kept polling its hashset every 8h regardless.
- `Guardian.reset()` (on MSP unbind) cleaned up rules but never cleaned up exceptions, so a lingering exception alone could keep the target list active permanently, surviving any number of reboots.

## Fix

- `PolicyManager2`/`ExceptionManager`: when the last rule or exception referencing a user target list is removed, deactivate the category immediately instead of leaving it active until the next reboot.
- `Guardian.reset()`: also delete exceptions referencing an MSP target list on unbind, not just rules.
- `CategoryUpdateSensor`: fall back to the default poll interval if the config is missing/invalid, instead of `setInterval` silently clamping a `NaN` delay to 1ms (found while testing the above with a shortened poll interval).

## Verification

see [comment](https://github.com/firewalla/firecommit/issues/9107#issuecomment-5202323078) on the issue for before/after redis + log evidence.

### DIFF
--- a/alarm/ExceptionManager.js
+++ b/alarm/ExceptionManager.js
@@ -35,6 +35,8 @@ const exceptionPrefix = "exception:";
 const _ = require('lodash');
 const Alarm = require('../alarm/Alarm.js');
 const CategoryMatcher = require('./CategoryMatcher');
+const CategoryUpdater = require('../control/CategoryUpdater.js');
+const categoryUpdater = new CategoryUpdater();
 
 const sem = require('../sensor/SensorEventManager').getInstance();
 const firewalla = require('../net2/Firewalla');
@@ -109,6 +111,23 @@ module.exports = class {
       }
       this.categoryMap = newCategoryMap;
     }
+  }
+
+  // whether any persisted exception still references this category
+  async hasException(category) {
+    if (!category) return false;
+    const exceptions = await this.loadExceptionsAsync();
+    return exceptions.some(e => e.getCategory() === category);
+  }
+
+  // user target list categories are only kept alive by rules/exceptions referencing them;
+  // once the last exception referencing one is removed, stop polling its hashset if no rule
+  // is using it either, instead of leaving it active until the next reboot
+  async _maybeDeactivateCategory(category) {
+    if (!category || !categoryUpdater.isUserTargetList(category)) return;
+    if (categoryUpdater.hasActivePolicies(category)) return;
+    if (await this.hasException(category)) return;
+    await categoryUpdater.deactivateCategory(category);
   }
 
   getExceptionKey(exceptionID) {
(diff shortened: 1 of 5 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly deactivates target-list categories only after both policy and exception references are gone.
- Bulk exception deletion preserves category information before removing Redis records.
- Uses the shared `CategoryUpdater` instance, so policy-reference state is consistently observed within each process.
- Linked issue `firewalla/firecommit#9107` was reviewed:
  - **Met:** Stop polling after the final policy is removed.
  - **Met:** Stop polling after the final exception is removed.
  - **Partially met:** Clean up MSP exceptions during unbind; identification is incomplete when MSP target-list data is stale or no longer contains the revoked list.
  - **Met:** Preserve categories still referenced by another policy or exception.

### ⚠️ Issues found

- **High — MSP exceptions can remain permanently after authorization disappears**

  `[sensor/Guardian.js:519]`

  `isMspRelatedException()` identifies an MSP exception exclusively by looking up its category in `mspData.targetlists`. This misses precisely the important stale/revoked cases described by the issue: if the target list was deleted, authorization was revoked, or the box was offline during the change, that list may no longer be present in `mspData`. The exception then survives `reset()` and continues reactivating and polling the unauthorized hashset after reboot.

  The reproduced exception contains `if.targetList: "1"`, which provides a durable local marker. Use that marker, or persist equivalent MSP ownership metadata when creating the exception, instead of relying solely on the potentially stale target-list snapshot.

- **Medium — Invalid interval values can still produce a 1 ms polling loop**

  `[sensor/CategoryUpdateSensor.js:643]`

  The `value || default` fallback handles `undefined`, `null`, `NaN`, and zero, but negative numbers, `Infinity`, and non-empty nonnumeric strings remain truthy. Passing such values to `setInterval()` can still be clamped to a very short delay, contrary to the stated goal of handling invalid configuration. Normalize each value with `Number()` and require `Number.isFinite(value) && value > 0` before using it.

### 💡 Suggestions

  - MSP unbind when the exception has `if.targetList=1` but its category is absent from `mspData.targetlists`.
  - Multiple policies/exceptions referencing the same category.
  - Partial and malformed interval configuration.
- Consider a shared `maybeDeactivateCategory()` helper to avoid duplicating the policy/exception reference checks across managers.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9431  
Head SHA: `4518a7ac5fe9678c3d2b51b0592f73623edc60d7`  
Checked at: 2026-08-06 17:02:49 CST

jasonlyc (approved):

PR #9433, opened 2026-08-07
sync gold 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Release verification uses a dedicated keyring, validates signed annotated tags against trusted fingerprints, and adds rollback protection via a monotonic minimum-version floor.
- Kernel-module identity handling is substantially improved for modules without `version` or `srcversion`, with good test coverage.
- The AP speed-test redirect validates IPv4 addresses and trusts forwarded headers only from loopback proxies.
- Target-list cleanup addresses both policy and exception lifecycles.
- RSPAN support includes dedicated capture plumbing, worker identification, and flow deduplication.
- Reviewed all 62 changed files, with particular focus on upgrade security, networking, category lifecycle, and shell execution.

Issue requirement coverage for `firewalla/firecommit#9107`:

- **Met:** Deleting the last policy reference deactivates the target-list category.
- **Met:** Deleting the last exception deactivates the category when no policy remains.
- **Met:** MSP reset removes target-list exceptions using cached MSP target-list data.
- **Met:** Polling stops without requiring a reboot.
- **Partially met:** No direct handling exists for an unauthorized hashset response; stale references not represented in cached MSP data can still continue polling.

### ⚠️ Issues found

[net2/BroDetect.js:1008]  
**Medium — VLAN ID alone is insufficient to identify RSPAN traffic.** `rspanVids` discards the parent/interface context, so any ordinary captured traffic carrying the same VLAN ID as an RSPAN VLAN is classified as RSPAN. VLAN IDs are only locally significant and can be reused across interfaces. This can select the RSPAN deduplication path and drop otherwise valid same-network flows. Prefer the dedicated `worker-rspan` identity, or retain and match the relevant capture interface plus VID.

[scripts/upgrade_verify.sh:284]  
**High — verification fails open whenever the OTA configuration request fails.** Even after release verification has been enabled, a timeout, malformed response, DNS failure, or temporary OTA outage makes `uv_is_enforced` return false and permits an unsigned update. This defeats the protection precisely during degraded or adversarial network conditions. Persist the last successfully fetched enforcement state and fail closed once enforcement has ever been enabled; only an explicit authenticated `false` should disable it.

[sensor/Guardian.js:515]  
**Medium — target-list cleanup depends entirely on potentially stale cached MSP data.** `isMspRelatedException()` only recognizes categories present in `mspData.targetlists`. If a target list was deleted or membership changed while the box was offline, the stale cache may omit the unauthorized category, leaving its exception active and preserving the recurring polling described in the linked issue. Consider using the exception’s target-list marker and MSP ownership metadata, or explicitly deactivate unauthorized `TL-*` categories after server rejection.

### 💡 Suggestions

- Add a test where an ordinary VLAN and an RSPAN VLAN reuse the same VID on different interfaces.
- Add upgrade-gate tests for timeout, invalid JSON, and recovery after enforcement was previously enabled.
- Add an MSP-reset regression test where the stale exception’s target list is absent from `mspData.targetlists`.
- Split future staging synchronization PRs by subsystem where practical; this 62-file change combines security, networking, UI routing, device support, and lifecycle fixes, making regression isolation difficult.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9433  
Head SHA: `25701a594645372bcf8b32c476e895fd282ac9e7`  
Checked at: 2026-08-07 11:11:43 CST

PR #9443, opened 2026-08-10
sync orange 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Security-sensitive shell executions generally move from shell interpolation to `execFile`/argument arrays.
- Release-tag verification adds trusted-key validation and monotonic minimum-version enforcement.
- New networking code validates IP addresses and handles IPv4-mapped IPv6 addresses.
- Relevant unit tests were added for kernel crash handling, DNS probing, topology traversal, dynamic reloads, and WLAN statistics.
- No linked issue found.

### ⚠️ Issues found

- **High — Supply-chain verification can silently fall back to an unverified branch tip.**  
  `[scripts/switch_branch.sh:74]`  
  When the node-module pin is missing, the code fetches and checks out `origin/$tgt_branch` without signature or commit-pin verification. This undermines the new verified-upgrade boundary: the main repository is authenticated, but executable dependencies can still come from a mutable branch. A missing pin should fail closed for verified upgrades.

- **Medium — A failed pinned dependency update leaves mismatched application and dependency revisions.**  
  `[scripts/switch_branch.sh:71]`  
  The Firewalla target branch is checked out before `uv_sync_node_modules()`. If that sync fails, the error is logged but `switch_branch()` continues successfully with the previous `~/.node_modules`. This can boot a release against incompatible dependencies. Abort the switch, or stage both repositories and activate them only after both updates succeed.

- **Medium — Partial clone failures are not cleaned up as documented.**  
  `[scripts/upgrade_verify.sh:156]`  
  If the first clone creates a partial destination and fails, the immediate retry commonly fails because the directory already exists. The function then returns before setting `fresh=1`, so its cleanup path never runs. Remove the partial destination before retrying and route all failures through one cleanup path.

### 💡 Suggestions

- Make pin presence mandatory whenever release verification is enabled.
- Treat the application checkout and pinned node-module checkout as one transactional operation.
- Add failure-path tests covering missing pins, unreachable pinned commits, partial clone failures, and node-module sync failure after the main checkout.
- Because this sync contains 157 changed files and several unrelated features, splitting future staging syncs into smaller thematic PRs would materially improve review and rollback safety.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9443  
Head SHA: `71b04dbd2578f080fbb4f764e00d553ea920b27d`  
Checked at: 2026-08-10 18:55:49 CST

PR #9452, opened 2026-08-17
fix bug, bf file checksum not align with redis

caixingcheng:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly removes data for the removed `part`, rather than the parent category.
- Validates downloaded Bloom-filter data against its MD5 checksum.
- Repairs stale Redis metadata when the file is already current.
- Uses rename-based replacement to prevent readers from observing partially written files.
- No linked issue found.

### ⚠️ Issues found

- **High — concurrent updates can still produce mismatched file and metadata.**  
  `[sensor/CategoryUpdateSensor.js:439]`

  The temporary filename only includes `process.pid`, so concurrent updates for the same category use the same file. `regularJob`, activation events, and debug-triggered jobs can overlap.

  One possible interleaving:

  1. Update A writes the temporary file.
  2. Update B overwrites that same temporary file.
  3. Update A renames B’s data into place.
  4. Update A stores metadata for A.
  5. Update B’s rename fails because the temporary file is gone.

  This recreates the exact file/checksum inconsistency the PR intends to prevent. Serialize updates per category, or use unique temporary filenames and ensure the file replacement plus metadata update cannot be reordered across competing updates.

### 💡 Suggestions

- Add tests covering:
  - stale Redis metadata repair;
  - checksum rejection;
  - removed multipart filter cleanup;
  - two concurrent updates for the same category.
- Consider validating `obj.info.s` against `buf.length`, since the size is forwarded to consumers but currently remains trusted cloud metadata.
- Preserve the cause when Redis metadata cannot be parsed instead of silently treating every read or parse failure as missing metadata; logging it would improve diagnosis.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9452  
Head SHA: `a527aab589acc83d2f1f9c27cace2c0cabe2fbb1`  
Checked at: 2026-08-17 18:27:22 CST

jasonlyc (approved):

PR #9458, opened 2026-08-18
 sync tag verification to gmx alpha

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Verification uses a dedicated keyring outside the updateable Git tree.
- Official repository URL matching rejects owner, hostname, and repository suffix tricks.
- Signer fingerprints are explicitly matched instead of trusting any key present in the keyring.
- Node modules are pinned to signed-tree revisions, with malformed and unreachable pins rejected.
- Tests cover signature validation, version floors, remote matching, key import, dry-run behavior, and node-module pinning.
- No linked issue found, so issue requirement coverage could not be assessed.

### ⚠️ Issues found

- **High — Enforcement fails open during an OTA configuration outage**  
  [`scripts/upgrade_verify.sh:279`](https://github.com/firewalla/firewalla/blob/8164be6d40352f3b74ec1ee0ec5fd6953ff93c58/scripts/upgrade_verify.sh#L279)

  `uv_is_enforced` treats timeouts, malformed responses, DNS failures, TLS failures, and missing `jq` identically to an explicit disabled value. Consequently, even after release verification has been enabled, an attacker who can disrupt access to `ota.firewalla.com` can make an unsigned update pass through the dry-run path.

  Persist the enabled state locally and make it monotonic, similar to the minimum-version floor. Once enforcement has been observed as enabled, configuration retrieval failures must fail closed. Only a successfully authenticated explicit disable mechanism should turn it off.

- **Medium — Version comparison silently ignores components after the fourth**  
  [`scripts/upgrade_verify.sh:58`](https://github.com/firewalla/firewalla/blob/8164be6d40352f3b74ec1ee0ec5fd6953ff93c58/scripts/upgrade_verify.sh#L58)

  `uv_version_ge` compares only indices `0..3`, although accepted version input allows any number of dot-separated components. For example, `1.2.3.4.0` and `1.2.3.4.1` compare as equal. This can undermine rollback protection if future release versions contain more than four components.

  Iterate to the maximum component count of both inputs, or validate and enforce an explicitly documented component limit.

### 💡 Suggestions

- Add tests showing that enforcement remains active when the OTA configuration request subsequently times out or returns invalid JSON.
- Add version-comparison cases with more than four components and malformed cached floor values.
- Validate the cached floor in `uv_get_version_floor`; a corrupted value currently reaches Bash arithmetic evaluation and can disrupt upgrades.
- Use a cleanup trap in `test_upgrade_verify.sh` so temporary keyrings and repositories are removed when the test exits early.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9458  
Head SHA: `69b5a29779af67386edd9d6cc7a9ab38b42645de`  
Checked at: 2026-08-19 10:08:05 CST

PR #9470, opened 2026-08-20
disable chrony ProtectHome on goldplus2 to fix NTP sync

melo-crypto:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Scope is limited to Gold Plus 2.
- The hook is idempotent when `ProtectHome=no` is already present.
- The original read-only mount state is preserved.
- `systemctl daemon-reload` is correctly invoked before restarting Chrony.
- No linked issue found.

### ⚠️ Issues found

- **High — `platform/goldplus2/hooks/before_main_run/09_fix_chrony_protecthome.sh:22`**  
  Reading and overwriting the same file through a pipeline is unsafe. `tee` may truncate `chrony.service` before `sed` finishes opening or reading it, potentially leaving the systemd unit partially written or empty. Generate the modified content in a temporary file first, validate it, then copy it over the existing inode using a controlled operation.

- **Medium — `platform/goldplus2/hooks/before_main_run/09_fix_chrony_protecthome.sh:22`**  
  The substitution only works when a line beginning exactly with `ProtectHome=` already exists. If the directive is absent or indented, the script makes no change but still reloads systemd and restarts Chrony on every execution. Handle both replacement and insertion, then verify that the resulting file contains an effective `ProtectHome=no`.

- **Medium — `platform/goldplus2/hooks/before_main_run/09_fix_chrony_protecthome.sh:35`**  
  Failures from the rewrite are ignored, and the partition is remounted read-only before confirming success. An unsuccessful or partial write is therefore followed by `daemon-reload` and a Chrony restart. Check every remount/write operation, validate the unit with `systemd-analyze verify`, and restart only after successful modification.

### 💡 Suggestions

- Add a cleanup trap so `/media/root-ro` is restored to read-only if any intermediate command fails.
- Quote `${FIREWALLA_HOME}` when sourcing `platform.sh`.
- Replace the command chain used to detect mount state with `findmnt`, if available, for clearer and more reliable parsing.
- Add a test covering an existing directive, a missing directive, an indented directive, and a failed write.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9470  
Head SHA: `4c615808c6bcf04f4139bae917aea8e8b30021fc`  
Checked at: 2026-08-20 16:43:59 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Scope is correctly limited to Gold Plus 2.
- The hook exits cleanly when the Chrony unit is absent.
- The temporary output is validated before replacing the service file.
- The original read-only state is generally preserved.
- `daemon-reload` occurs before restarting Chrony.
- No linked issue found.

### ⚠️ Issues found

- **High — `platform/goldplus2/hooks/before_main_run/09_fix_chrony_protecthome.sh:23-27`**  
  The matching logic ignores leading whitespace. If the unit contains an indented `ProtectHome=yes`, the script inserts `ProtectHome=no` immediately after `[Service]`, while the later `ProtectHome=yes` remains and overrides it. The validation still succeeds because it only checks that any exact `ProtectHome=no` line exists. Match optional whitespace and replace every active `ProtectHome` assignment, then verify the effective unit configuration before restarting Chrony.

- **Medium — `platform/goldplus2/hooks/before_main_run/09_fix_chrony_protecthome.sh:50-52`**  
  Failure to remount `/media/root-ro` read-only is ignored. The hook can therefore report success while leaving the root filesystem writable. Capture the remount result and fail explicitly if restoration does not succeed.

- **Medium — `platform/goldplus2/hooks/before_main_run/09_fix_chrony_protecthome.sh:33-35`**  
  The final write and temporary-file removal are not checked independently. Because the function returns the status of `rm`, a failed overwrite followed by successful cleanup can return success. Although the later content check catches many failures, it cannot distinguish an unsuccessful update from a pre-existing matching line. Explicitly check the `tee` result and return failure before cleanup.

### 💡 Suggestions

- Use a cleanup trap to restore the read-only mount state on every exit path and signal.
- Use `findmnt` or `/proc/self/mountinfo` instead of parsing human-oriented `mount` output.
- Quote the sourced path: `source "${FIREWALLA_HOME}/platform/platform.sh"`.
- Validate the effective setting with `systemctl show chrony.service -p ProtectHome` or an appropriate systemd verification command.
- Add tests for indented directives, duplicate directives, failed writes, and failed read-only remounts.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9470  
Head SHA: `987769d1252ead702e0dd3527fb0f17ddab5c580`  
Checked at: 2026-08-20 16:58:45 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Change is scoped to Gold Plus 2.
- Preserves the original read-only mount state during the normal success path.
- Uses a temporary file and validates the transformed unit before overwriting it.
- Verifies systemd’s effective `ProtectHome` value after `daemon-reload`.
- No linked issue found.

### ⚠️ Issues found

- **Medium — restart failures are silently ignored**  
  `[platform/goldplus2/hooks/before_main_run/09_fix_chrony_protecthome.sh:74]`  
  `systemctl restart chrony.service` can fail, but the script still exits successfully. That would report the repair as complete while chrony continues running with its previous configuration. Check the return status and exit nonzero on failure.

- **Medium — a failed read-only remount may persist across retries**  
  `[platform/goldplus2/hooks/before_main_run/09_fix_chrony_protecthome.sh:47]`  
  If writing succeeds but remounting `/media/root-ro` read-only fails, the script exits with the service file already containing `ProtectHome=no`. On the next invocation, the outer condition is false, so the script never retries the remount and may report success while the filesystem remains writable. Mount-state recovery must not be conditional solely on whether the unit still needs editing.

### 💡 Suggestions

- Check failures from both `systemctl daemon-reload` and `systemctl restart`.
- Use clearer arithmetic/status tests such as `(( writable != 0 ))`.
- Consider an `EXIT` trap to restore the original mount state on every path after this script remounts the filesystem writable.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9470  
Head SHA: `664b1549d506bce1d46400bd75aac67fd4ca8883`  
Checked at: 2026-08-20 17:14:57 CST

jasonlyc (approved):

PR #9482, opened 2026-08-25
sync gp sfp staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Broad test coverage was added for NIC error-counter parsing across multiple drivers.
- FreeRADIUS image mismatch and pull-backoff handling aims to preserve the currently running service during failures.
- Release tags now match `getReleaseType()` naming and avoid oversized generated release notes.
- No linked issue was found in the PR title, body, or comments.

### ⚠️ Issues found

[platform/goldplus2/hooks/before_main_run/09_fix_chrony_protecthome.sh:75]

**Medium — A failed chrony restart is silently treated as success.**

The script does not check the exit status of `systemctl restart chrony.service`. If restart fails, the hook still exits successfully. Worse, because `ProtectHome=no` has already been written, subsequent executions skip the entire repair block, so the restart is never retried and chrony may continue running with the old sandbox configuration.

Please propagate the restart failure, for example by exiting non-zero when `systemctl restart` fails. The initial fast-path should also verify systemd’s effective property or the service state so a partially completed previous run can recover.

### 💡 Suggestions

- Add a shell test covering restart failure followed by a second invocation, ensuring the hook retries instead of short-circuiting.
- Consider checking the return value of `systemctl daemon-reload` explicitly for clearer failure reporting.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9482  
Head SHA: `814f2c5da1794b82fc63246d71940414aad0e63e`  
Checked at: 2026-08-25 16:49:10 CST

PR #9488, opened 2026-08-26
sycn purple 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Release tags now match the dotted version format expected by upgrade verification.
- FreeRADIUS changes preserve an existing container when the replacement image is unavailable and add pull-loop backoff.
- Permanent MAC validation rejects malformed, zero, and broadcast addresses.
- Ethernet statistics parsing handles multiple drivers and includes strong fixture-based coverage.
- Adblock statistics use a single Redis transaction and include reset-generation protection.
- No linked issue was found in the PR title, body, or comments.

### ⚠️ Issues found

- **Medium — overly broad rule matching adds latency and may misattribute records**  
  [sensor/ACLAuditLogPlugin.js:679]

  `getMatchedPids()` now runs for every blocked DNS record without a PID, although the stated behavior and tests concern only records whose reason is `adblock`. This adds a potentially expensive asynchronous rule lookup to the general blocked-DNS hot path and can attach a policy ID to records produced by unrelated blocking mechanisms. Restrict this lookup to `record.reason === 'adblock'`.

- **Medium — Ethernet statistics command permits shell interpretation of interface names**  
  [extension/sysinfo/SysInfo.js:597]

  `nic` is interpolated into a shell command. Linux interface names can contain shell-significant characters, so a locally created or otherwise unexpected interface name could alter the command. The function is now applied across all platform-provided Ethernet interfaces, increasing the exposure compared with the previous fixed commands. Use `execFile('ethtool', ['-S', nic])` instead.

- **Low — sequential NIC collection unnecessarily extends check-in time**  
  [extension/sysinfo/SysInfo.js:628]

  Each NIC waits for its speed read and `ethtool` invocation before the next NIC is processed. On platforms with several ports—or when `ethtool` is slow—this directly increases check-in latency. Collect per-interface data concurrently with `Promise.all`, while preserving per-NIC error handling.

### 💡 Suggestions

- Add a regression test proving that non-adblock blocked records do not invoke `getMatchedPids()`.
- Add focused tests for FreeRADIUS service enable/disable transitions and image-mismatch recovery.
- Add tests for the new `TimeSeries.queueHits()` result slicing, especially when granularities span different numbers of Redis keys.
- Consider using a systemd drop-in for the Chrony override instead of rewriting the vendor unit, which package upgrades may replace.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9488  
Head SHA: `814f2c5da1794b82fc63246d71940414aad0e63e`  
Checked at: 2026-08-26 11:14:57 CST

PR #9501, opened 2026-08-28
use gpgv instead of gpg, update key from S3

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly replaces unavailable `gpg` usage with verify-only `gpgv`.
- Official repositories fail closed when the trusted keyring is unavailable.
- Asset-delivered keyrings take precedence over the bootstrap key, supporting rotation and revocation.
- Network operations now have bounded timeouts.
- Existing version-floor and non-official/test-repository behavior remain intact.
- Linked issue firewalla/firecommit#8050 is substantially covered.

### ⚠️ Issues found

- **Medium — `scripts/update_assets.sh:18,54`: Single-asset arguments are still evaluated as shell code.**  
  The comment says nothing supplied by the caller is evaluated, but the arguments are concatenated into `SINGLE_ENTRY` and later passed through `eval`. Paths containing whitespace will be parsed incorrectly, and shell metacharacters can execute commands. Process single-asset arguments directly as an array instead of routing them through the legacy text parser.

- **Low — `test/test_upgrade_verify.sh:145`: Key-rotation verification result is never asserted.**  
  `RC_ASSET` captures the result after installing the asset-delivered keyring but is not passed to `check`. Consequently, the test suite can pass even if verification does not actually use the rotated key. Add an explicit assertion for the expected return code.

- **Low — `scripts/upgrade_verify.sh:127`: Trusted keyring replacement is non-atomic.**  
  `cp -f` can leave the persistent keyring truncated if the process is interrupted. Copy to a temporary file in the destination directory, validate it with `gpgv` or `gpg --show-keys` where available, then atomically rename it.

### 💡 Suggestions

- Add tests for single-asset paths containing spaces and shell metacharacters.
- Add an end-to-end positive rotation test signed by the new asset key, plus a negative test proving the old key is no longer accepted.
- Validate that the downloaded hash is exactly 64 hexadecimal characters, rather than checking length alone.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9501  
Head SHA: `b62f7ba919e2347347827ac009f36bcfc4dffbda`  
Checked at: 2026-08-28 16:37:59 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Replaces unavailable `gpg` operations with verify-only `gpgv`.
- Preserves fail-closed behavior for official repositories.
- Uses atomic replacement for the installed release keyring.
- Authenticates the S3-delivered keyring through the existing signed-assets pipeline.
- Retains the installed rotated key when S3 is temporarily unavailable.
- Avoids `eval` for single-asset arguments and adds bounded network operations.
- Linked issue #8050: signed-tag verification, version-floor enforcement, key rotation, and recovery behavior remain covered.

### ⚠️ Issues found

[scripts/upgrade_verify.sh:277]

**Medium — Existing test boxes silently lose verification after upgrade.**

The previous implementation used `/home/pi/.upgrade-gnupg-test`, while this PR only checks `/home/pi/.upgrade-keys/test.gpg`. Existing test boxes following issue #8050’s documented setup will therefore appear to have no test key and silently enter “skip verification” mode.

Please provide a migration path for the existing test keyring, fail with an actionable message when the legacy directory is detected, or update/provision all test environments atomically with this change.

### 💡 Suggestions

- Add an end-to-end test for `update_assets.sh` single-asset mode covering signature verification, paths containing spaces, download failure, and concurrent invocation.
- Update issue #8050’s test guide to document binary `gpgv` keyrings and the new paths.
- Consider validating that a downloaded `release_pub.gpg` is a parseable OpenPGP keyring before replacing the last known-good installed keyring.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9501  
Head SHA: `f7b84bdc50507b20d8eab37940eeca2366b08bdf`  
Checked at: 2026-08-28 17:35:31 CST

jasonlyc (approved):

PR #9510, opened 2026-08-31
sync orange 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Strong security hardening around shell execution, profile names, CPU IDs, package actions, and WireGuard private keys.
- Alarm-window logic handles calendar dates, time zones, and DST carefully, with extensive tests.
- Concurrent SysInfo refreshes are deduplicated and update-loop failures no longer permanently stop collection.
- WireGuard configuration parsing is now consistently case-insensitive.
- No linked issue found for PR #9510.

### ⚠️ Issues found

- **Medium — `controllers/netbot.js:3998`: apt operations may fail after exceeding `execFile`’s output buffer.**  
  The previous pipeline streamed output through `tee`. The new `execFile()` call captures all stdout/stderr in memory using a bounded buffer. Long `upgrade` or `dist-upgrade` output can exceed that buffer, terminate the child, and incorrectly report an upgrade failure. Use `spawn()` with streamed output, ideally preserving `/var/log/fwapt.log`, or explicitly configure a sufficiently large buffer.

- **Medium — `diagnostic/portcheck.js:91`: checking availability temporarily opens the requested port on all interfaces.**  
  `nc -l -p` creates a real listener for up to 200 ms. An inbound connection during that interval can make `nc` exit successfully and produce a false result; it also exposes a transient service endpoint. Prefer a read-only socket-table query such as `ss`, or bind explicitly to loopback if bind-based probing is unavoidable.

- **Low — `test/setup.sh:5`: setup destructively removes `node_modules` without validating the working directory.**  
  Although the script first attempts `cd ~/firewalla`, failure does not stop execution because `set -e` is absent. It could consequently remove `node_modules` from the caller’s directory. Add strict mode and derive the repository root from the script location before modifying anything.

### 💡 Suggestions

- Preserve apt command output in `/var/log/fwapt.log`; losing that log makes field diagnosis of failed upgrades significantly harder.
- Add tests covering large apt output and a concurrent connection during port probing.
- Given the unusually broad staging PR, consider requiring all newly added shell and alarm-window tests to pass before merge.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9510  
Head SHA: `d306e73d8eef9cf5c1a66e1c289a9b495c43cd16`  
Checked at: 2026-08-31 17:19:53 CST

PR #9517, opened 2026-08-31
Suspend DMZ allow-source rules in non-router mode

hhirokazu:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Synchronizes DMZ allow rules during startup and mode changes.
- Reuses the existing policy update and reenforcement mechanisms.
- Serializes mode-change processing with `LOCK_SHARED`.
- No linked issue found.

### ⚠️ Issues found

- **High — `extension/portforward/portforward.js:479`:** Every DMZ policy is forcibly enabled when router mode starts. This also re-enables policies that were already disabled by the user or another subsystem before the mode transition. Likewise, clearing `idleTs` destroys existing pause state. Preserve the original state or record whether the rule was disabled specifically by the non-router-mode synchronization before restoring it.

- **Medium — `extension/portforward/portforward.js:446`:** Initial synchronization runs outside `LOCK_SHARED`. A concurrent `Mode:Applied` handler can interleave with this pass and leave policies reflecting a stale mode. Run the initial synchronization under the same lock and determine the mode inside the critical section.

### 💡 Suggestions

- Skip policy persistence and reenforcement when the desired state already matches the current state. Startup currently rewrites every DMZ policy, changes `updatedTime`, and emits reenforcement events even when nothing changed.
- Add tests covering user-disabled rules, paused rules with `idleTs`, repeated startup, and rapid router/non-router transitions.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9517  
Head SHA: `40008f666bb5c59eb20434bf566621df3416789f`  
Checked at: 2026-09-01 05:02:21 CST

rtrappman-dev:
I see a potentially serious state-preservation issue in `_syncDMZAllowRules()`.

When `routerMode` becomes true, the current code constructs every policy with:

```js
disabled: routerMode ? "0" : "1",
idleTs: ""
```

That means a DMZ rule that was already disabled by the user can be re-enabled when router mode is restored. Likewise, an existing `idleTs` pause can be cleared even though that pause was unrelated to router mode.

I think the synchronization needs to distinguish between:

```text
disabled because non-router mode
```

and:

```text
disabled by the user / another policy mechanism
```

so that restoring router mode only reverses the state change introduced by this synchronization.

There also appears to be a smaller concurrency issue: the initial `_syncDMZAllowRules(await Mode.isRouterModeOn())` runs outside the existing `LOCK_SHARED`, while mode-change synchronization runs inside it. A mode transition during startup could therefore race with the initial pass.

Could the implementation preserve the original disabled/paused state and run the initial synchronization under the same lock? Regression coverage for user-disabled rules, existing `idleTs`, repeated startup, and rapid router/non-router transitions would be useful.

jasonlyc (approved):

PR #9520, opened 2026-09-01
Add MemorySwapMax to brofish.service so MemoryMax is enforceable

### LINKED ISSUE #9519: brofish (zeek) memory pressure can take down dnsmasq box-wide — MemoryMax without MemorySwapMax converts a contained OOM into swap thrash
**Environment:** Firewalla Gold Plus (4GB RAM), base image 3.0929, production branch, zeek 5.0.1, router mode, ~14 monitored VLANs.

**What happened:** A zeek worker (the main LAN trunk interface) grew to 2.2GB total-vm / 571MB rss. `brofish.service` sets `MemoryMax=1000M` but no `MemorySwapMax` — so at the cap, the cgroup's pages went to swap instead of triggering an in-cgroup OOM kill. The box spent ~5 minutes swap-thrashing (15-min load 27): **dnsmasq (`firerouter_dns`) stopped answering for every VLAN** — a total DNS outage from the LAN's perspective — until the *global* OOM killer finally killed zeek. Recovery after that was flawless: the worker watchdog noticed the dead worker and restarted brofish within seconds.

```text
16:13–16:18  dnsmasq query log silent; dhcpcd "route socket overflowed";
             intelproxy "Fail to connect to redis."
16:18:43     kernel: Out of memory: Killed process (zeek)
             total-vm:2201104kB, anon-rss:571332kB  [global_oom,
             task_memcg=/system.slice/brofish.service]
16:18:59     brofish ping failed(HB:OK, CPU:OK, RSS:OK, Worker:fail),
             restart brofish now
```

**This is chronic, not one-off:** node-level OOM-kill counters on this box incremented on 3 separate days in the last 30 (only the last one hit DNS hard enough to be noticed).

**Why the existing guardrails missed it:**

1. `MemoryMax=1000M` alone doesn't contain — with swap available, hitting the cap produces box-wide thrash rather than a clean kill. Note the log line above: the watchdog reported `RSS:OK` while the box was already unusable.
2. `brofish-ping.sh` requires `RSS >= 800MB` **AND** `free mem <= 60MB` simultaneously, and RSS is measured per-process, not cgroup-total. A single worker at 571MB (cgroup ~1GB total) satisfies neither, so no proactive restart ever fires in this failure mode.

**Suggestions:**

- Add `MemorySwapMax=` (0 or small) to `brofish.service` alongside `MemoryMax` — this one line converts the failure from "router-wide DNS outage" to "zeek restarts itself," and the existing worker watchdog already handles the recovery beautifully. — PR submitted alongside this issue
- Optionally add `MemoryHigh=` for graceful pressure/reclaim before the kill.
- Consider making the watchdog RSS check cgroup-total (`/sys/fs/cgroup/system.slice/brofish.service/memory.current`) and OR-ing the conditions, or adding a swap-activity signal — the AND-gate can't fire during the exact scenario it exists for.
- Protect `firerouter_dns` with `OOMScoreAdjust`/`MemoryMin` so core DNS wins against an analytics component under global pressure.

We've applied the `MemorySwapMax`/`MemoryHigh` drop-in locally via `post_main.d` and it fully contains the failure. Happy to share more logs or test a fix. And genuinely: the `local.bro` extension hook in `before_bro.sh` and the post_main.d persistence mechanism made a clean local mitigation possible — nice design.

*Written with AI assistance (Claude); the full contents were reviewed and approved by a human. Incident data and testing are from the submitter's own device.*

sp4ir:
## What

Adds `MemorySwapMax=64M` directly after the existing `MemoryMax=` line in all
11 `platform/*/files/brofish.service` files (blue, gold, goldplus2, goldpro,
gse, navy, orange, pse, purple, red, ubt). One line per file; nothing else
changed.

## Why

`MemoryMax=` (cgroup v2 `memory.max`) alone does not actually bound a service
on a box that has swap: when the cgroup hits the cap, the kernel pushes its
anonymous pages to swap instead of OOM-killing inside the cgroup. The service
keeps growing "into" swap while staying nominally under its memory cap.

Observed failure sequence on a 4GB Gold in production:

1. A zeek worker leaked and reached the brofish `MemoryMax=1000M` cap.
2. Instead of a contained OOM kill, the cgroup started swapping heavily.
3. The resulting box-wide swap thrash starved everything else — dnsmasq
   became unresponsive on all networks (total DNS outage for every client)
   for roughly five minutes.
4. The outage ended only when the kernel's **global** OOM killer finally
   intervened.

`MemorySwapMax=64M` (cgroup v2 `memory.swap.max`) closes that escape hatch:
once the cgroup has used its small swap allowance, hitting `MemoryMax`
triggers an in-cgroup OOM kill of the zeek process. Zeek dies alone, the rest
of the box is unaffected, and the existing brofish watchdog restarts the
service cleanly — the recovery path already works perfectly once the kill is
contained; the only missing piece was making the kill happen inside the
cgroup.

## Tested

- Production Firewalla Gold Plus running 3.0929 (zeek 5.0.1).
- Deployed as a systemd drop-in with exactly these values
  (`MemoryMax=1000M` + `MemorySwapMax=64M`).
- Verified: on reaching the cap, zeek is OOM-killed inside the brofish
  cgroup, no box-wide impact, and the brofish watchdog restarts it cleanly.
- Other platforms are untested by the author; they carry the same
  `MemoryMax` mechanism, so the same swap-escape applies, but I have only
  validated the Gold Plus values.

## Scope / follow-up

This intentionally does **not** add `MemoryHigh=` (which would give earlier
kernel back-pressure before the hard cap) to keep the change minimal and easy
to review; that could be a follow-up if maintainers want it.

Fixes #9519

---

This change was developed with AI assistance (Claude) and was reviewed and
tested by the submitter.

### DIFF
--- a/platform/blue/files/brofish.service
+++ b/platform/blue/files/brofish.service
@@ -14,6 +14,7 @@ RestartSec=10
 TimeoutStartSec=250
 TimeoutStopSec=8
 MemoryMax=600M
+MemorySwapMax=64M
 
 [Install]
 WantedBy=multi-user.target
--- a/platform/gold/files/brofish.service
+++ b/platform/gold/files/brofish.service
@@ -15,6 +15,7 @@ RestartSec=10
 TimeoutStartSec=250
 TimeoutStopSec=8
 MemoryMax=1000M
+MemorySwapMax=64M
 Nice=5
 
 [Install]
--- a/platform/goldplus2/files/brofish.service
+++ b/platform/goldplus2/files/brofish.service
@@ -13,6 +13,7 @@ RestartSec=10
 TimeoutStartSec=250
 TimeoutStopSec=8
 MemoryMax=1000M
+MemorySwapMax=64M
 Nice=5
 
 [Install]
--- a/platform/goldpro/files/brofish.service
+++ b/platform/goldpro/files/brofish.service
@@ -15,6 +15,7 @@ RestartSec=10
 TimeoutStartSec=250
 TimeoutStopSec=8
 MemoryMax=1000M
+MemorySwapMax=64M
 Nice=5
 
 [Install]
--- a/platform/gse/files/brofish.service
+++ b/platform/gse/files/brofish.service
@@ -13,6 +13,7 @@ RestartSec=10
 TimeoutStartSec=250
 TimeoutStopSec=8
 MemoryMax=1000M
+MemorySwapMax=64M
 Nice=5
 
 [Install]
--- a/platform/navy/files/brofish.service
+++ b/platform/navy/files/brofish.service
@@ -15,6 +15,7 @@ RestartSec=10
 TimeoutStartSec=250
 TimeoutStopSec=8
 MemoryMax=600M
+MemorySwapMax=64M
 Nice=5
 
 [Install]
--- a/platform/orange/files/brofish.service
+++ b/platform/orange/files/brofish.service
@@ -14,6 +14,7 @@ RestartSec=10
 TimeoutStartSec=250
 TimeoutStopSec=8
 MemoryMax=600M
+MemorySwapMax=64M
 Nice=5
 
 [Install]
--- a/platform/pse/files/brofish.service
+++ b/platform/pse/files/brofish.service
@@ -13,6 +13,7 @@ RestartSec=10
 TimeoutStartSec=250
 TimeoutStopSec=8
 MemoryMax=600M
+MemorySwapMax=64M
 Nice=5
 
 [Install]
--- a/platform/purple/files/brofish.service
+++ b/platform/purple/files/brofish.service
@@ -14,6 +14,7 @@ RestartSec=10
 TimeoutStartSec=250
 TimeoutStopSec=8
 MemoryMax=600M
+MemorySwapMax=64M
 Nice=5
 
 [Install]
--- a/platform/red/files/brofish.service
+++ b/platform/red/files/brofish.service
@@ -14,6 +14,7 @@ RestartSec=10
 TimeoutStartSec=250
 TimeoutStopSec=8
 MemoryMax=100M
+MemorySwapMax=64M
 
 [Install]
 WantedBy=multi-user.target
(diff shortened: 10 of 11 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The change is minimal and consistently applies `MemorySwapMax=64M` to all 11 `brofish.service` definitions that already specify `MemoryMax`.
- The directive is correctly placed in the systemd resource-control section.
- No unrelated changes, security concerns, concurrency issues, or maintainability regressions were found.
- Linked issue #9519 coverage:
  - **Met:** Bound brofish/Zeek swap usage to prevent unbounded swap growth.
  - **Met:** Preserve the existing watchdog-based recovery path.
  - **Met:** Apply the mitigation across every platform carrying a brofish memory cap.
  - `MemoryHigh`, watchdog changes, and DNS OOM protection were presented as optional follow-ups, so their omission is appropriate.

### ⚠️ Issues found

- **Low — Platform compatibility is not demonstrated.** `MemorySwapMax` is effective with the unified cgroup v2 hierarchy; on systems using cgroup v1, it may have no effect. Since only Gold Plus was tested, the PR does not establish that the other ten platforms run a compatible systemd/cgroup configuration. This is especially relevant because the PR description claims the mitigation applies to all platforms.

### 💡 Suggestions

- Before merging, verify representative images—particularly older platforms such as Red—using:
  - `systemd-analyze verify brofish.service`
  - `systemctl show brofish -p MemorySwapMax`
  - confirmation that `/sys/fs/cgroup` is mounted as `cgroup2`
- Document platforms where the setting is intentionally best-effort or ineffective under cgroup v1.
- Add a lightweight service-file validation check covering these platform variants if the repository has an appropriate CI path.
- Reviewed 11 of 11 changed files and generated one cross-cutting comment; no line-specific defect requires an inline comment.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9520  
Head SHA: `416f4edb7c684ca0b4d19b5b890f7cc1807bce0c`  
Checked at: 2026-09-01 08:48:14 CST

rtrappman-dev (approved):

PR #9529, opened 2026-09-01
Speed up SysInfo.getSysInfo() on the app init path

jasonlyc:
Dedupe concurrent getSysInfo() calls, cache proc version, and rework getTop10RSSProcesses to avoid per-process sudo spawns and blocking the event loop.

### DIFF
--- a/extension/sysinfo/SysInfo.js
+++ b/extension/sysinfo/SysInfo.js
@@ -319,6 +319,40 @@ async function getKernelVersion() {
   return kernelVersion;
 }
 
+// Wraps an async producer with a TTL cache and in-flight de-dup, same pattern as the
+// usbInfoPromise guard on getUsbInfo() below: getSysInfo() can be called concurrently
+// (e.g. by both basicDataForInit() and extensionDataForInit() in the same init request),
+// so without de-dup a slow/expensive producer would otherwise run once per caller.
+// A falsy/null result is never cached, so a failed lookup is retried on the next call.
+// ttlMs === Infinity caches the first successful result forever.
+function cachedAsync(producer, ttlMs) {
+  let value, ts = 0, inflight = null;
+  return async function() {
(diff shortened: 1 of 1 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Replaces blocking `execSync` calls on the initialization path with asynchronous operations.
- Correctly de-duplicates concurrent cache misses through a shared in-flight promise.
- Batches privileged `/proc/<pid>/exe` lookups into one `sudo` invocation.
- Validates parsed PIDs before shell interpolation, avoiding command injection through that path.
- Runs independent system-information lookups concurrently while preserving the returned schema.
- No linked issue found.

### ⚠️ Issues found

- **Low — misleading cache contract:** The comment says falsy results are never cached, but `cachedAsync()` caches every value except `null` and `undefined`, including `false`, `0`, and `""`. Update the comment or change the condition so future callers do not rely on incorrect behavior.
- **Low — no automated coverage:** The new caching and concurrency behavior is untested. Regressions involving rejected producers, concurrent callers, TTL expiration, and retry-after-null would be difficult to detect.

### 💡 Suggestions

- Add focused tests verifying:
  - concurrent callers invoke the producer only once;
  - successful values remain cached for the configured TTL;
  - `null` and rejected results are retried;
  - the in-flight state clears after both resolution and rejection.
- Consider moving `inflight = null` into `.finally()` to express the cleanup invariant once and reduce duplicated state management.
- Consider using `fs.promises.readFile('/proc/version', 'utf8')` instead of spawning `cat`, since `fs` is already available.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9529  
Head SHA: `bbd6b611a9b449f83a408691798f47d6e9852662`  
Checked at: 2026-09-01 17:48:36 CST

MelvinTo (approved):

PR #9536, opened 2026-09-02
sync gp sfp 1.984 staging

jasonlyc:


MelvinTo (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Replaces active port probing with kernel socket-table inspection.
- Reduces blocking subprocess work and deduplicates concurrent system-info requests.
- Block-stat buckets are indexed, bounded, and self-describing.
- Platform-specific `wpa_cli` resolution improves compatibility.
- No linked issue found.

### ⚠️ Issues found

- **High — `alarm/AppTimeUsageManager.js:196`: deregistration race can still throw.**  
  The code explicitly permits deregistration while `enforcePolicy()` is pending. `_deregisterPolicy()` deletes `this.enforcedPolicies[pid]`, but `applyPolicy()` later unconditionally writes `this.enforcedPolicies[pid][uid]`. This can raise `TypeError` after the rule was already removed. Before setting `POLICY_STATE_DOMAIN_ONLY`, verify both the original registered policy and the per-policy state object still exist.

- **Medium — `sensor/BlockStatsSensor.js:162`: a Redis failure permanently clears dirty state.**  
  Dirty markers are removed before `multi.execAsync()`, but they are not restored if the transaction fails. Unless another matching flow arrives, the accumulated bucket will never be retried and may eventually be purged from memory. Restore the cleared markers in a `catch` block, or only acknowledge a captured generation after a successful write.

- **Medium — `sensor/BlockStatsSensor.js:125`: cloud configuration is used without validation.**  
  A zero, negative, non-numeric, or extremely small `slotSecs` produces invalid bucket timestamps or excessive bucket creation. Malformed settings can also throw when `recordKeys` is absent or non-array. Validate and bound the cloud configuration before replacing the known-good local defaults.

### 💡 Suggestions

- Add failure-path tests covering Redis transaction rejection and retry.
- Add a regression test where policy deregistration occurs while enforcement is pending.
- Add configuration tests for invalid `slotSecs`, filters, keys, and `recordKeys`.
- Consider parsing malformed `sys:ept:members` entries individually so one bad JSON member does not prevent lookup of all remaining entries.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9536  
Head SHA: `3ca1aa125699a928f4db28483d0995a2ed59ed23`  
Checked at: 2026-09-02 10:33:06 CST

PR #9550, opened 2026-09-02
Bound WHOIS response buffering

rtrappman-dev:
## Summary

The low-level WHOIS client buffers the entire remote response in memory by appending every received TCP chunk to an array and only concatenating the data after the connection closes.

A malicious or compromised WHOIS server can therefore send an arbitrarily large response and cause the Firewalla process to consume excessive memory.

## Changes

- Add a 1 MiB maximum WHOIS response size.
- Track the cumulative number of bytes received from the TCP connection.
- Reject the lookup when the response exceeds the configured maximum.
- Immediately destroy the socket when the limit is exceeded.
- Prevent the `close` handler from processing an already-rejected request.
- Preserve the existing WHOIS parsing and fallback behavior for valid responses.

## Security / Availability Impact

WHOIS responses are now explicitly bounded before additional buffering and parsing can continue.

This prevents the WHOIS client from retaining an unbounded amount of attacker-controlled network data in process memory and reduces the potential for memory-exhaustion-based denial of service through a malicious WHOIS endpoint.

## Testing

Added `test/test_whois.js` covering:

- rejection of responses larger than 1 MiB;
- normal handling of responses within the configured limit;
- the production TCP buffering path rather than a mocked socket implementation.

Full repository test execution was not performed in this environment.

### DIFF
--- a/lib/whois/index.js
+++ b/lib/whois/index.js
@@ -26,6 +26,7 @@ async function lookup (query, options) {
     let content = [];
     let contentSize = 0;
     let settled = false;
+    let rejectPromise;
 
     client.on('data', function(data){
         if (settled) {
@@ -51,8 +52,6 @@ async function lookup (query, options) {
         tryIP = true;
       }
     });
-
-    let rejectPromise;
 
     return new Promise((resolve, reject) => {
         rejectPromise = reject;
--- a/lib/whois/index.js
+++ b/lib/whois/index.js
@@ -10,6 +10,8 @@ const defaultOptions = {
   raw: false
 };
 
+const MAX_RESPONSE_SIZE = 1024 * 1024;
+
 async function lookup (query, options) {
   let tryIP = false;
 
@@ -22,8 +24,23 @@ async function lookup (query, options) {
     });
 
     let content = [];
+    let contentSize = 0;
+    let settled = false;
 
     client.on('data', function(data){
+        if (settled) {
+          return;
+        }
+
+        contentSize += data.length;
+        if (contentSize > MAX_RESPONSE_SIZE) {
+          const err = new Error(`WHOIS response exceeds maximum size of ${MAX_RESPONSE_SIZE} bytes`);
+          settled = true;
+          client.destroy();
+          rejectPromise(err);
+          return;
+        }
+
         content.push(data);
     });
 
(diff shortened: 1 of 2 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly caps buffered WHOIS data at 1 MiB, mitigating memory-exhaustion attacks.
- Uses byte length rather than string length.
- Destroys the socket immediately after exceeding the limit.
- Prevents the `close` handler from resolving or rejecting the promise twice.
- Preserves existing parsing and fallback behavior for valid responses.
- No linked issue found.

### ⚠️ Issues found

- **Low — Boundary behavior is not actually tested** (`test/test_whois.js:71`): The test claims to cover responses “at or below the limit,” but sends only `"ok"`. A regression from `>` to `>=` would go undetected. Add a test containing exactly 1 MiB and assert that it succeeds.

- **Low — Socket closure is not explicitly verified** (`test/test_whois.js:52`): The test title says the connection is closed, but it only verifies promise rejection. Track the server-side socket’s `close` event to confirm the resource-cleanup behavior.

### 💡 Suggestions

- Consider exposing the response-size limit through options while retaining 1 MiB as the default, if some legitimate WHOIS providers can return larger datasets.
- A single test using multiple writes would better exercise cumulative accounting across TCP chunks.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9550  
Head SHA: `69fd78cf80ca64fc5e3635f721b15b13bb6124c2`  
Checked at: 2026-09-03 03:01:42 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Enforces a strict 1 MiB cumulative byte limit before buffering additional WHOIS data.
- Destroys the socket immediately when the limit is exceeded and rejects with a clear error.
- Guards the `close` handler against settling the promise again.
- Correctly accepts a response exactly at the limit.
- Tests cover rejection, boundary acceptance, and server-observed connection closure.
- No linked issue found.

### ⚠️ Issues found

None.

### 💡 Suggestions

- Consider adding a test that sends the oversized response across multiple writes to explicitly verify cumulative accounting across TCP chunks.
- Consider making the limit configurable if legitimate WHOIS providers are later found to return responses larger than 1 MiB.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9550  
Head SHA: `536ac4e73ed786d6031e2f5a6e86269afc64b498`  
Checked at: 2026-09-03 03:17:19 CST

PR #9555, opened 2026-09-02
fix: bound deferred DNS TTL state and drop excess refreshes at capacity

rtrappman-dev:
## Summary

Bound DNSTool's retained deferred DNS TTL state to 50,000 map entries across pending, overflow, retry, active, and active-update state.

The previous deferred queue could grow without bound. Subsequent changes also left conflicting implementations interleaved in the production module and tests, preventing both files from parsing. This revision removes those remnants and enforces the combined capacity limit.

## Changes

- Restore a single synchronous `tryRefreshDnsTTL()` implementation and matching caller behavior.
- Count all five deferred-state containers, including an active key's separately retained update.
- Admit ordinary deferred entries below 49,000 total entries; reserve the remaining capacity for overflow or active updates. Independently cap the overflow map at 1,000 entries.
- Coalesce updates to already queued entries without evicting unrelated work.
- Serialize drains, process one initial batch per invocation, and retain rejected batches for later retry.
- Preserve newer updates across drain completion and failure. Clear stale unsent state when a leading-edge refresh supersedes it or a DNS mapping is removed.
- Replace malformed, interleaved tests with executable coverage and add explicit syntax checks for the production and test files to the existing CI workflow.

## Overload policy and compatibility

Normal leading-edge refreshes still request inline EXPIRE; throttled refreshes are deferred while capacity remains.

When admission capacity is exhausted, refreshes requiring another retained entry are intentionally dropped. Producers do not wait for Redis, and saturation does not trigger extra inline Redis operations. Already queued entries remain retained and can be updated in place. An update to an in-flight command requires a separate slot and can also be dropped when all 50,000 slots are occupied.

Drops are counted and reported in an aggregate warning when a drain runs. This is an explicit best-effort overload policy: it does **not** guarantee delivery of every refresh or completion within one timer period. Retry work, active updates, and pending work take priority over overflow, so sustained higher-priority work can delay overflow processing.

The 50,000-entry limit applies to retained deferred map entries. The last-refresh LRU is separately capped at 50,000; this is not a byte-level process-memory limit or a global Redis write-rate limit. Under overload, dropped or delayed TTL updates can leave older expiration behavior in effect.

## Validation

Executed locally against the repaired files:

- Node 16.20.2 syntax checks for `net2/DNSTool.js` and `test/test_dns_tool.js`: passed.
- ESLint 8.57.1 with the repository configuration on both files: passed.
- Mocha 9.2.2 focused suite: **13 passing**, including both public TTL-writing callers, 60,000 admissions, aggregate capacity, overflow draining, stalled Redis, overlapping drains, active updates, retry failure/recovery, coalescing, and removal cleanup.
- Restoring either overflow-accounting omission or active-update undercounting causes its targeted test to fail.

The focused tests use a stubbed Redis boundary and real production DNSTool methods. Full repository tests, GitHub CI results, live Redis integration, and appliance behavior have not been verified in this development run.


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The pending map now has a deterministic memory bound.
- Existing-key updates do not evict unrelated entries.
- The interval handle is retained so tests can clean it up.
- Tests cover capacity enforcement and update-at-capacity behavior.
- No linked issue found.

### ⚠️ Issues found

- **Medium — `net2/DNSTool.js:71`: Eviction drops a required TTL refresh.**  
  Removing the oldest pending entry means `_drainDnsTTL()` will never refresh that key unless another update happens. The key can therefore expire despite receiving recent DNS data. This contradicts the existing contract that throttled refreshes are “deferred (not dropped)” and flushed within one period.

  When the queue is full, preserve existing pending entries and make the new refresh execute inline—for example, update its throttle timestamp and return `true`—instead of silently discarding another key’s refresh.

- **Medium — `net2/DNSTool.js:73`: Capacity pressure can generate an unbounded warning stream.**  
  Once the map reaches 50,000 entries, every new unique key emits a warning. The same high-cardinality workload this change is intended to contain could produce heavy log I/O and disk usage. Rate-limit or aggregate this warning, such as logging once per drain period with a dropped/overflow count.

- **Low — `test/test_dns_tool.js:34`: Tests encode eviction but do not verify TTL correctness.**  
  The tests confirm map cardinality and which entry disappears, but not whether every accepted DNS update still receives an `EXPIRE`. Add a test using a stubbed Redis client and `_drainDnsTTL()` to verify the overflow path preserves TTL refresh semantics.

### 💡 Suggestions

- Define the capacity once and reuse it for both `dnsExpireTs` and `dnsExpirePending`, since the PR describes these as matching bounds.
- Update the nearby comments if dropping refreshes is intentionally accepted; they currently promise the opposite behavior.
- CI reports no checks for this branch, and the PR states that the full test suite was not run. Run the targeted test and relevant repository checks before merging.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `b5f97aff1a70e91decec84d3442ebc005a7751ba`  
Checked at: 2026-09-03 06:17:10 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds the deferred refresh map at 50,000 entries.
- Preserves already queued refreshes instead of evicting them.
- Aggregates overflow warnings to avoid log flooding.
- Tests cover capacity, existing-key updates, draining, and warning aggregation.
- No linked issue found.

### ⚠️ Issues found

- **Medium — `net2/DNSTool.js:74`: repeated updates for an overflow key bypass throttling.** Once the queue is full, every subsequent update for a key not already pending returns `true`, even after its timestamp was refreshed. A hot key can therefore issue `EXPIRE` on every update until the next drain, creating avoidable Redis load precisely during high-cardinality pressure.

  Track overflow-handled keys for the throttle period, trigger an early queue drain, or otherwise ensure only the first overflow update executes inline. Add a test that calls `tryRefreshDnsTTL()` repeatedly for the same overflow key and verifies only one inline refresh occurs.

### 💡 Suggestions

- Update the PR description: it still says the oldest pending refresh is evicted, which no longer matches the implementation.
- Reuse `MAX_DNS_EXPIRE_PENDING` for the `dnsExpireTs` LRU capacity to keep the shared bound defined in one place.
- Run the targeted test suite before merging; GitHub currently reports no checks for this branch.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `e9e5f81f8e9b476c123f0882cf48753fa58ae101`  
Checked at: 2026-09-03 06:33:00 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds `dnsExpirePending` without discarding already-deferred refreshes.
- Existing pending keys can still update their expiry value at capacity.
- Overflow refreshes are throttled, warning logs are aggregated, and removal clears all related state.
- The interval handle is retained so tests can clean it up.
- No linked issue found.

### ⚠️ Issues found

- **High — New overflow tests do not enter the overflow branch and will fail.**  
  [test/test_dns_tool.js:58]  
  `overflowKey` is absent from `dnsExpireTs`, so `tryRefreshDnsTTL()` returns from the normal first-refresh branch before checking queue capacity. Consequently, `dnsExpireOverflowCount` remains `0`, contradicting the assertion at line 64.

- **High — The repeated-refresh test exercises normal throttling rather than overflow throttling.**  
  [test/test_dns_tool.js:104]  
  The first call for `overflowKey` again takes the normal first-refresh path and never inserts the key into `dnsExpireOverflowTs`. The test therefore does not validate the newly introduced overflow suppression behavior, and its overflow-count assertion will fail.

- **High — The warning aggregation test cannot produce the expected warning.**  
  [test/test_dns_tool.js:126]  
  Keys `50000`–`50002` have no prior entries in `dnsExpireTs`, so all three calls are ordinary first-time refreshes. No overflow is recorded and `_drainDnsTTL()` will not emit the asserted warning.

### 💡 Suggestions

- Seed each overflow key in `dnsExpireTs` with a recent timestamp before calling `tryRefreshDnsTTL()`. This forces execution through the capacity check and genuinely tests the overflow path.
- Add an assertion that `dnsExpireOverflowTs.has(overflowKey)` becomes true after overflow handling.
- Update the PR description: it currently says the oldest deferred refresh is evicted, but the implementation preserves the pending queue and refreshes the new key inline.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `a54fdca997a6761c50e3bba59aa8a42d20acf0ab`  
Checked at: 2026-09-03 06:48:36 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds the deferred DNS TTL queue at 50,000 entries.
- Preserves queued refreshes by handling overflow keys inline.
- Suppresses repeated inline refreshes during the throttle period.
- Correctly retains updates to existing queued keys at capacity.
- Clears all throttle state when reverse-DNS entries are removed.
- Tests cover capacity, update, overflow throttling, and warning aggregation.
- No linked issue found.

### ⚠️ Issues found

No blocking correctness, security, concurrency, or resource-management issues found in the two changed files.

### 💡 Suggestions

- Update the PR description: it still says the oldest deferred refresh is evicted, but the final implementation preserves existing entries and refreshes the new key inline.
- Add an explicit assertion that `dnsExpireOverflowTs` contains the overflow key, strengthening coverage of the new suppression mechanism.
- Consider using the shared capacity constant in tests instead of repeating `50000`, if exporting or injecting it can be done without complicating production code.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `c033936e1eb911eea4fef904d5a3c20e612d3a83`  
Checked at: 2026-09-03 08:21:01 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Caps `dnsExpirePending` at 50,000 entries.
- Preserves existing queued refreshes when capacity is reached.
- Aggregates overflow warnings instead of flooding logs.
- Clears overflow throttle state when DNS records are removed.
- Adds focused coverage for capacity, repeated overflow, and warning behavior.
- No linked issue found.

### ⚠️ Issues found

- **Medium — `net2/DNSTool.js:71`: overflow suppression can drop a later TTL refresh after the queue drains.**

  An overflow key is immediately rejected while it remains in `dnsExpireOverflowTs`. If the periodic drain clears `dnsExpirePending` shortly after that key was refreshed inline, a subsequent update is neither refreshed inline nor added to the now-available deferred queue.

  Consequently, the TTL reflects the earlier update rather than the latest one, violating the method’s documented guarantee that throttled refreshes are deferred. This is especially relevant when the later call supplies a different `expr`.

  Only suppress the repeat while the queue remains full. Once capacity is available, enqueue the overflow key and clear its overflow marker.

### 💡 Suggestions

- Add a regression test covering: fill queue → overflow key refreshes inline → drain queue → update the same overflow key → verify it enters `dnsExpirePending`.
- Update the PR description, which still says the oldest deferred entry is evicted; the current implementation preserves queued entries and refreshes new overflow keys inline.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `3e2390a69a14909322dfe4a637a8c647a800c3a5`  
Checked at: 2026-09-04 01:17:28 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds the deferred TTL refresh queue at 50,000 entries.
- Preserves existing deferred refreshes by handling overflow keys inline.
- Suppresses repeated inline refreshes during the throttle period.
- Re-queues overflow keys once capacity becomes available.
- Aggregates overflow warnings to avoid log flooding.
- Clears throttle and overflow state when DNS records are removed.
- Tests cover capacity, existing-key updates, overflow throttling, post-drain behavior, and warning aggregation.
- No linked issue found.

### ⚠️ Issues found

No blocking correctness, security, concurrency, or resource-management issues found in the two changed files.

### 💡 Suggestions

- Update the PR description. It still states that the oldest deferred refresh is evicted, while the implementation now preserves pending entries and refreshes overflow keys inline.
- Use `MAX_DNS_EXPIRE_PENDING` for the `dnsExpireTs` LRU capacity as well, keeping the shared limit defined in one place.
- Run the targeted test and relevant repository checks before merging; GitHub currently reports no checks for this branch.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `c02a15cb073f969c2db25fe4961e8d2d8b6dccb9`  
Checked at: 2026-09-04 01:31:34 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds `dnsExpirePending` to prevent unbounded memory growth.
- Preserves updated TTL values for keys already queued.
- Aggregates overflow warnings instead of logging per request.
- Adds focused tests around capacity, overflow throttling, draining, and warning behavior.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:72`:** `dnsExpireOverflowTs.delete(key)` is incompatible with the repository’s `lru-cache` v4 dependency, which exposes `del(key)`, not `delete(key)`. Once an overflow key is revisited after queue capacity becomes available, this throws `TypeError: this.dnsExpireOverflowTs.delete is not a function`, potentially crashing the caller. Replace it with `dnsExpireOverflowTs.del(key)`. The added “defers a later refresh” test should expose this failure when executed.

### 💡 Suggestions

- Run the new test suite before merging; the PR currently reports no CI checks.
- Update the PR description: it still says the oldest pending refresh is evicted, while the current implementation retains pending entries and refreshes overflow keys inline.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `f27714a552a0e6ab4fbdf120b00191ccab477a7c`  
Checked at: 2026-09-05 01:36:48 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds `dnsExpirePending` at 50,000 entries without discarding already-deferred refreshes.
- Preserves updated expiry values for existing pending keys.
- Handles overflow inline and suppresses repeated refreshes for cached overflow keys.
- Aggregates warnings instead of logging for every overflow event.
- Clears all associated throttle state when records are removed.
- Tests cover capacity, repeated overflow, post-drain deferral, and warning aggregation.
- No linked issue found.

### ⚠️ Issues found

- **Medium — `net2/DNSTool.js:86`: overflow throttling degrades under sustained high cardinality.** `dnsExpireOverflowTs` holds only 50,000 keys. Once more than 50,000 distinct keys overflow, older markers are evicted. Because `dnsExpireTs` has the same capacity, revisiting those keys can execute `EXPIRE` inline again within the throttle period. An adversarial workload can therefore generate sustained Redis traffic despite the stated suppression behavior. Consider a globally rate-limited overflow path, bounded batch draining, or explicitly document that suppression is best-effort.

### 💡 Suggestions

- Update the PR description: it still says the oldest pending refresh is evicted, while the final implementation preserves pending entries and refreshes overflow keys inline.
- Reuse `MAX_DNS_EXPIRE_PENDING` for the `dnsExpireTs` capacity instead of duplicating `50000`.
- Run the targeted test suite before merging; GitHub currently reports no checks for the branch.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `41b1035bc55d13d55ce5559e88fb1af02cd06489`  
Checked at: 2026-09-05 03:35:12 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds `dnsExpirePending` to prevent unbounded memory growth.
- Preserves updates for keys already queued.
- Retains the interval handle so tests can clean it up.
- Adds focused coverage for capacity, overflow throttling, and drain behavior.
- No linked issue found.

### ⚠️ Issues found

- **High — New Redis keys can be left without expiration**  
  `[net2/DNSTool.js:71]`  
  The global overflow check runs before determining whether the key has ever received an expiration. While overflow suppression is active and the pending queue remains full, a newly created key returns `false` without being queued. Since `addDns()` and `addReverseDns()` perform `ZADD` first and call `EXPIRE` only when this method returns `true`, that key can remain permanently TTL-less unless another update happens later. This recreates an unbounded Redis-storage risk during sustained high-cardinality traffic. Check the leading-edge case before applying overflow suppression, ensuring every previously unseen key either receives an inline `EXPIRE` or is retained for draining.

- **Medium — Added test has an impossible final expectation**  
  `[test/test_dns_tool.js:167]`  
  After the first drain, `dnsExpireOverflowCount` is reset to zero and `dnsExpirePending` is empty. The second drain therefore returns immediately and cannot replace the warning with `refreshed inline: 3`. This test should expect the original warning to remain unchanged, or generate three new overflow events before asserting another aggregated warning.

### 💡 Suggestions

- Add a regression test that activates overflow suppression, calls `tryRefreshDnsTTL()` with a key absent from `dnsExpireTs`, and verifies that the caller is instructed to expire it inline.
- Update the PR description: it still says the oldest pending refresh is evicted, but the final implementation instead performs one inline refresh and globally suppresses subsequent overflow refreshes.
- Run the new test file in CI; GitHub currently reports no checks for this branch.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `4ea25b81125b33835ebcfbfa9368d38a8339deb5`  
Checked at: 2026-09-05 03:51:11 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds `dnsExpirePending` at 50,000 entries.
- Ensures previously unseen Redis keys still receive an inline expiration.
- Preserves the original queue during drain and aggregates overflow warnings.
- Adds focused tests for capacity, suppression, draining, and warning behavior.
- Reviewed both changed files.
- No linked issue found.

### ⚠️ Issues found

- **Medium — Updates to existing pending entries are discarded during overflow suppression**  
  `[net2/DNSTool.js:77]`  
  When overflow suppression is active and the queue remains full, this branch returns before checking whether `key` already exists in `dnsExpirePending`. Consequently, a later call with a different `expr` does not update that entry, and the drain applies the stale TTL. This also contradicts the PR’s stated requirement that existing pending entries remain updateable. Handle existing keys before the overflow-suppression branch, since replacing a value does not increase map cardinality.

### 💡 Suggestions

- Add a regression test that activates overflow suppression, updates an existing pending key with a different TTL, and verifies that the new TTL is retained.
- Update the PR description: it says the oldest deferred refresh is evicted, while the implementation preserves the queue, refreshes one overflow entry inline, and suppresses subsequent overflow entries globally.
- Review performed using the [github_pr_review skill](/Users/tt/.openclaw/workspace/skills/github_pr_review/SKILL.md).

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `002a6f359a8e548099c17846e7317de28be19b8e`  
Checked at: 2026-09-05 04:21:38 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds `dnsExpirePending` at 50,000 entries.
- Preserves updated TTL values for existing pending keys.
- Ensures previously unseen Redis keys receive an inline expiration.
- Aggregates overflow warnings to prevent log flooding.
- Adds focused tests for capacity, suppression, draining, and warning behavior.
- No linked issue found.

### ⚠️ Issues found

- **High — Throttled refreshes are dropped when the queue is full**  
  `[net2/DNSTool.js:86]`

  While global overflow suppression is active, a recently seen key that is not already pending returns `false` without being added to `dnsExpirePending` or refreshed inline. If no later update arrives after the queue drains, Redis never receives the required `EXPIRE`, despite the documented guarantee that throttled refreshes are deferred rather than dropped.

  This can cause active reverse-DNS records to expire during sustained high-cardinality traffic. The test at `test/test_dns_tool.js:113` currently codifies this loss by expecting only the first of three overflow keys to be refreshed.

  Preserve every accepted refresh—for example, trigger an immediate bounded drain before enqueueing the overflow key, or use another bounded mechanism that does not silently discard refresh requests. Add a regression test verifying that every overflow key is eventually included in an `EXPIRE` operation.

### 💡 Suggestions

- Update the PR description: it still says the oldest deferred entry is evicted, whereas the implementation preserves the queue and globally suppresses most overflow refreshes.
- Use `MAX_DNS_EXPIRE_PENDING` for the `dnsExpireTs` LRU capacity instead of duplicating `50000`.
- Run the targeted test and relevant repository checks before merging.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `4014e6d67352cd74afea9213720a35b953c2756c`  
Checked at: 2026-09-05 04:48:30 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Caps each active deferred queue at 50,000 entries.
- Preserves updated TTL values for existing pending keys.
- Ensures overflow refreshes are eventually processed instead of silently dropped.
- Aggregates overflow warnings to avoid per-update log flooding.
- Adds focused tests for capacity, overflow suppression, draining, and TTL updates.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:86`: asynchronous drains can accumulate without bound.**

  When the queue is full, `_drainDnsTTL()` swaps out the map and starts a Redis operation, but the returned promise is not awaited or tracked. The newly empty queue can fill and trigger another drain while earlier batches remain in flight.

  Under sustained high-cardinality traffic or a slow Redis connection, an unlimited number of 50,000-entry maps and Redis transactions can coexist. This moves the original memory-growth problem from `dnsExpirePending` into untracked in-flight drains and can also overload Redis.

  Serialize drains or enforce a bounded in-flight queue. New refreshes should be merged into a bounded pending structure while a drain is active, with retry/backpressure behavior defined for Redis failures. Add a test using an unresolved `execAsync()` promise to verify that repeated capacity events cannot launch overlapping batches.

### 💡 Suggestions

- Update the PR description: it still says the oldest deferred refresh is evicted, while the implementation now preserves entries and starts an early drain.
- Reuse `MAX_DNS_EXPIRE_PENDING` for the `dnsExpireTs` LRU capacity instead of duplicating `50000`.
- Add failure-path coverage for rejected Redis operations and confirm deferred refreshes are retried or intentionally dropped.
- Run the targeted test suite and relevant repository checks before merging.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `2708a0fa62453c5a85214a030edc7ecf639347c4`  
Checked at: 2026-09-05 05:18:31 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds deferred and retry queues under high-cardinality DNS traffic.
- Serializes drains to prevent overlapping 50,000-command Redis batches.
- Preserves pending entries and updated expiry values at capacity.
- Aggregates overflow warnings to avoid log flooding.
- Adds focused coverage for capacity, throttling, draining, and concurrency.
- No linked issue found.

### ⚠️ Issues found

- **Medium — A stale retry can overwrite a newer inline TTL**  
  `[net2/DNSTool.js:78]`

  The immediate-refresh branch removes the key only from `dnsExpirePending`, not `dnsExpireRetry`. After a failed drain, high cardinality can evict that key from `dnsExpireTs`. A subsequent update then executes `EXPIRE` inline, but the older retry remains queued and can later apply its stale `expr`, overwriting the newer TTL.

  Remove or update matching retry entries when performing an immediate refresh. The active drain batch may also need tracking so an in-flight stale operation cannot race with the newer inline expiration.

### 💡 Suggestions

- Add a regression test covering: failed drain → retry retained → timestamp evicted/reset → same key refreshed with a different TTL → verify the old retry cannot overwrite it.
- Add direct coverage for rejected `expireAsync()` and `multi.execAsync()` calls; retry behavior is substantial new logic but currently untested.
- Update the PR description: it still claims the oldest deferred refresh is evicted, while the implementation now preserves queued entries and uses inline refresh/early draining.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `ffb7b23aa2ed6b7f8fd9ec14acf9fce178373e53`  
Checked at: 2026-09-05 06:48:49 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds each deferred DNS TTL queue to 50,000 entries.
- Serializes Redis drains, preventing unbounded concurrent batches.
- Preserves updates to queued keys and avoids stale retries after inline refreshes.
- Retries failed batches and aggregates overflow warnings.
- Adds focused tests for capacity, overflow, failure, and concurrency behavior.
- No linked issue found.

### ⚠️ Issues found

- **Medium — Newer pending refreshes can be delayed for two drain periods.**  
  `[net2/DNSTool.js:104]`

  When `dnsExpireRetry` is non-empty, `_drainDnsTTL()` processes only that retry batch. Refreshes accumulated in `dnsExpirePending` remain untouched until the next 30-minute interval—even when the retry succeeds.

  A refresh queued just before the first drain can therefore wait almost 60 minutes, contradicting the documented guarantee that deferred refreshes are flushed within one period. After a successful retry, immediately schedule or continue with the pending batch while retaining serialized execution and bounded memory.

### 💡 Suggestions

- Add a regression test with both `dnsExpireRetry` and `dnsExpirePending` populated, then verify one drain cycle eventually processes both batches without overlapping Redis operations.
- Update the PR description: it still says the oldest deferred refresh is evicted, while the current implementation preserves queued entries and uses early draining or inline refreshes.
- Use `MAX_DNS_EXPIRE_PENDING` for the LRU capacity instead of duplicating `50000`.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `ad9900af1576ce44c4991dadb89cda68a2443905`  
Checked at: 2026-09-05 07:20:01 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds deferred and retry queues under high-cardinality DNS traffic.
- Serializes drains to prevent overlapping Redis batches.
- Preserves updated expiry values for pending keys.
- Retries failed batches and drains newer pending work in the same cycle.
- Aggregates overflow warnings to avoid log flooding.
- Adds focused capacity, failure, and concurrency tests.
- No linked issue found.

### ⚠️ Issues found

- **Medium — `net2/DNSTool.js:82`: removing a key from the active map cannot cancel an in-flight Redis operation.**

  `tryRefreshDnsTTL()` deletes a newly refreshed key from `dnsExpireActive`, but `_drainDnsTTL()` has already submitted that batch to Redis before yielding. If the caller performs a newer inline `EXPIRE` while the earlier `execAsync()` or `expireAsync()` remains unresolved, the older operation may complete afterward and overwrite the newer TTL.

  This can either prematurely expire an active record or retain it longer than requested, depending on the two expiry values. The existing stale-retry test covers a rejected batch, but not a successfully completing in-flight batch.

  Prevent inline refreshes from racing active operations—for example, version queued entries and reapply the latest value after the active batch completes, or defer same-key updates until the serialized drain owns them.

### 💡 Suggestions

- Add a regression test that leaves the first Redis operation unresolved, refreshes the same key inline with a different TTL, then completes the older operation and verifies the final TTL remains the newer value.
- Update the PR description: it still describes evicting the oldest deferred entry, while the current implementation preserves entries and uses early draining or inline refreshes.
- Run the targeted test suite and relevant repository checks before merging.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `14eaf1edfb644bfeac1303ef935989727d2831e4`  
Checked at: 2026-09-05 07:48:58 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds deferred TTL state and serializes Redis drain batches.
- Preserves newer values for pending keys and retries failed batches.
- Aggregates overflow warnings to avoid log flooding.
- Adds focused capacity, retry, and concurrency tests.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:101`: overflow handling can race an active refresh for the same key.**

  When a drain is active and `dnsExpirePending` is full, this branch returns `true` without checking whether `key` belongs to `dnsExpireActive`. The caller can consequently issue a newer inline `EXPIRE` while the older batched operation remains unresolved. If the older operation completes afterward, it overwrites the newer TTL.

  The `dnsExpireActiveUpdates` protection at lines 81–83 only covers the leading-edge branch, so throttled overflow updates remain vulnerable.

  Before returning `true` from either capacity-overflow path, detect active keys and place their latest `expr` in `dnsExpireActiveUpdates` for serialized processing.

### 💡 Suggestions

- Add a regression test with an unresolved active batch, a full pending queue, and a newer TTL for a key in the active batch. Verify the method returns `false` and the newer TTL executes only after the older operation completes.
- Run the targeted test suite and relevant repository checks before merging; the PR notes that full test/CI validation was unavailable.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `d5f8110bfe9d5132e9b32e4a91388153cadb370f`  
Checked at: 2026-09-05 08:04:56 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Caps deferred DNS TTL state at 50,000 entries per queue.
- Serializes drain operations and preserves newer updates for active keys.
- Retains failed batches for retry without allowing stale retries to overwrite newer refreshes.
- Aggregates overflow warnings and adds extensive concurrency/failure-path tests.
- No linked issue found.

### ⚠️ Issues found

- **High — overflow suppression fails while a drain is active**  
  `[net2/DNSTool.js:104]`

  When `dnsExpirePending` is full and `dnsExpireDrainPromise` is active, every non-active key returns `true`, even while global overflow suppression is in effect. The timestamp written before returning does not prevent subsequent calls for that key—or other keys—from taking the same branch.

  With slow Redis and sustained high-cardinality traffic, this permits an unbounded stream of inline `EXPIRE` calls, contradicting the PR’s stated goal of preventing overflow from creating excessive Redis load.

  Keep overflow refreshes bounded while a drain is active—for example, reserve bounded follow-up capacity, coalesce them into a bounded structure, or apply explicit backpressure instead of returning `true` for every arrival.

### 💡 Suggestions

- Add a regression test that holds `execAsync()` unresolved, fills the replacement pending queue, then submits many distinct and repeated overflow keys. Assert that inline expiration calls remain bounded.
- The focused test could not be executed in a clean checkout because development dependencies such as `chai` were unavailable; CI should validate it before merge.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `306a4bb78d08b92d725c2a1ecd4ca960ecc001fb`  
Checked at: 2026-09-05 08:20:56 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Serializes drains to prevent overlapping Redis operations.
- Preserves newer TTL values for keys already in an active batch.
- Adds strong regression coverage for capacity, retry, overflow, and concurrency behavior.
- Aggregates overflow warnings to avoid log flooding.
- No linked issue found.

### ⚠️ Issues found

- **High — Deferred state can exceed the intended 50,000-entry bound.**  
  [`net2/DNSTool.js:175`] During an active drain, `dnsExpirePending` can contain 50,000 keys and `dnsExpireActiveUpdates` can independently contain another 50,000. Merging them without checking the combined cardinality can create a 100,000-entry batch. Together with `dnsExpireActive`, memory consumption can be substantially higher still. This contradicts the PR’s deterministic memory-bound objective. Enforce one shared capacity across all deferred states or process active updates in bounded chunks.

- **High — TTL refreshes are silently dropped after `dnsExpireActiveUpdates` reaches capacity.**  
  [`net2/DNSTool.js:105`] and [`net2/DNSTool.js:121`] The method updates `dnsExpireTs` and returns `false` even when the capacity check prevents inserting the refresh. The caller therefore does not issue an inline `EXPIRE`, and no deferred operation retains it. This violates the stated guarantee that throttled refreshes are never discarded. When the update cannot be queued, return `true` for inline handling or retain it through another bounded mechanism.

### 💡 Suggestions

- Add a stress test with an unresolved active drain, 50,000 pending keys, and more than 50,000 distinct overflow updates. Assert both:
  - total deferred cardinality never exceeds the documented bound;
  - every accepted refresh is either queued or explicitly handled inline.
- Centralize queue admission into one helper so all pending, retry, and active-update paths enforce identical capacity and fallback semantics.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `b9cdc2f507e941c2b509eddb8dc97908cc2ba6fd`  
Checked at: 2026-09-05 08:55:29 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds aggregate deferred state across pending, retry, active, and active-update queues.
- Serializes drains and preserves newer updates during in-flight Redis operations.
- Retains failed batches for retry and avoids replaying stale values after newer refreshes.
- Adds substantial regression coverage for capacity, retries, concurrency, and timer cleanup.
- Reviewed both changed files.
- No linked issue found.

### ⚠️ Issues found

- **High — overflow suppression fails during an active full drain**  
  [`net2/DNSTool.js:110`](https://github.com/firewalla/firewalla/blob/4e162faf62f9bfe0d09a1c2aa7f9c0646cd8d45e/net2/DNSTool.js#L110)

  When all 50,000 slots remain occupied by the active batch and other deferred state, `queueActiveUpdate()` returns `true` for every new key. Concurrent callers therefore execute `EXPIRE` inline while Redis is already unresolved. A high-cardinality burst can produce an effectively unbounded number of Redis operations, contradicting the PR’s stated overflow-suppression and availability guarantees.

  The test named `does not start overlapping drains while Redis is unresolved` actually codifies this behavior by expecting 100 consecutive overflow calls to return `true`.

  Please defer overflow work until the active drain releases capacity, or introduce another bounded/coalesced mechanism. The current boolean API may need to become asynchronous so callers can wait for drain capacity without dropping refreshes or issuing unlimited inline commands.

### 💡 Suggestions

- Add a concurrency test that starts a blocked 50,000-entry drain, invokes many distinct overflow refreshes without yielding between calls, and asserts that inline Redis operations remain bounded.
- Include inline overflow operations in `dnsExpireOverflowCount`; currently only the first overflow is counted, so the aggregate warning materially underreports saturation.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `4e162faf62f9bfe0d09a1c2aa7f9c0646cd8d45e`  
Checked at: 2026-09-08 03:56:04 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Serializes Redis drains, preventing overlapping large batches.
- Preserves newer TTL values for keys already active or pending.
- Retries failed batches without replaying stale entries.
- Aggregates overflow warnings to avoid log flooding.
- Adds substantial coverage for capacity, retry, and concurrency behavior.
- Reviewed both changed files.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:142`: the capacity-waiter queue reintroduces unbounded memory growth.**

  `dnsExpireCapacityWaiters` is not included in `_dnsExpireDeferredSize()` and has no independent capacity limit. While Redis is slow or unavailable and the 50,000 deferred slots remain occupied, every distinct incoming key adds another map entry, Promise, resolver, and suspended caller. Repeated updates for one key also append unlimited resolvers at line 146.

  A sustained high-cardinality workload can therefore grow memory without bound—the exact failure mode this PR intends to eliminate. Include waiting work in a strict aggregate bound and define a bounded fallback once that capacity is exhausted. Coalescing should also avoid retaining one resolver per update.

### 💡 Suggestions

- Add a stress test that blocks `execAsync()`, fills deferred capacity, then submits substantially more than 50,000 distinct overflow keys and repeated updates for one key. Assert that total retained state remains bounded.
- Run the targeted tests and relevant repository checks before merging; the PR currently reports no CI checks.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `f15f473fdfd726431623073267306fa355246e7d`  
Checked at: 2026-09-08 04:19:31 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Deferred state is explicitly bounded.
- Existing queued refreshes are preserved at capacity.
- Drains are serialized and failed batches retained for retry.
- Updates for pending keys are coalesced.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:119`: Overflow suppression performs another inline refresh.**  
  When the queue is full and suppression is active without an existing drain, this branch starts `_drainDnsTTL()` and returns `true`. The caller consequently executes `EXPIRE` inline for the second overflow key, contradicting the stated “one inline refresh, suppress/defer subsequent refreshes” behavior.

  This also appears to make these new tests fail:

  - `rate-limits distinct overflow keys globally within the throttle period`
  - `aggregates overflow warnings until the pending queue drains`

  Both expect only the first overflow key to be refreshed inline, but the second call follows this branch and also returns `true`.

- **Medium — `net2/DNSTool.js:88`: An update to an active key can race with the older in-flight refresh.**  
  At aggregate capacity, `_waitForDnsExpireCapacity()` immediately returns `true`. The caller then performs the newer `EXPIRE` concurrently with the active older refresh. If the older operation completes last, it can overwrite the newer TTL. Since the update is for a key already counted in `dnsExpireActive`, coalescing it into `dnsExpireActiveUpdates` would not increase unique-key cardinality and would preserve ordering.

### 💡 Suggestions

- After starting the drain at line 119, queue the overflow refresh behind that drain and return `false` or a capacity-wait promise instead of requesting another inline `EXPIRE`.
- Track capacity by unique keys across active and update maps, or explicitly allow an active key’s coalesced update without consuming another capacity slot.
- Add a regression test where an active batch is at full capacity and a newer TTL for the same key must execute strictly after the older TTL.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `6ee6935e3ec01d349fcccf58a603d6695aaa2d96`  
Checked at: 2026-09-08 04:33:22 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Serializes drain operations to prevent overlapping Redis batches.
- Preserves newer updates arriving while a batch is active.
- Retains failed batches for retry.
- Adds substantial coverage for concurrency, retries, overflow behavior, and timer cleanup.
- No linked issue found.

### ⚠️ Issues found

- **High — The deferred queue can still exceed 50,000 entries.**  
  `[net2/DNSTool.js:113]`  
  While overflow suppression is active and no drain is running, the full-queue branch unconditionally inserts the new key into `dnsExpirePending`. Every subsequent unique key can therefore grow the map beyond `MAX_DNS_EXPIRE_PENDING`. The added “rate-limits distinct overflow keys globally” test explicitly expects the map to accept keys 50,001 and 50,002, contradicting the PR’s deterministic-memory-bound objective. At capacity, these refreshes must remain inline, be dropped under an explicitly documented policy, or enter storage that is genuinely included within the fixed bound.

### 💡 Suggestions

- Add an invariant assertion covering the aggregate state after every overflow path:
  `dnsExpirePending + dnsExpireRetry + dnsExpireActive + active updates + waiters <= MAX_DNS_EXPIRE_PENDING`.
- Rework or remove the capacity-waiter mechanism. `_waitForDnsExpireCapacity()` rejects new waiters whenever aggregate size is already at the limit, which is precisely when current production call paths invoke it; this makes most of the mechanism effectively unreachable and adds considerable complexity.
- Add a sustained-overflow regression test that submits many unique keys without starting a drain and verifies the aggregate deferred state never exceeds 50,000.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `9925019f6f0e42f971bfcc68a19ddf0d445826f5`  
Checked at: 2026-09-08 05:01:51 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Serializes drains and retains failed batches for retry.
- Bounds deferred state and coalesces repeated updates by key.
- Aggregates overflow warnings to avoid log flooding.
- Updates callers to await the newly asynchronous refresh decision.
- Adds useful capacity, failure, and in-flight drain coverage.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:112`: overflow behavior contradicts the new test and will fail CI.** After the queue drains, `dnsExpireOverflowTs` remains active. A subsequent refresh for the same recently refreshed overflow key enters this branch; because capacity is now available, it clears the overflow timestamp, queues the key, and returns `false`. However, `test/test_dns_tool.js` expects that call to return `true` and remain inline. Either correct the implementation or update the test and document the intended semantics.

- **Medium — `net2/DNSTool.js:137`: the capacity-waiter path is effectively unreachable from production callers.** Every call to `_waitForDnsExpireCapacity()` occurs only after the aggregate deferred size is already at the maximum, while the method immediately returns `true` at that size. Consequently, no waiter is registered and the advertised backpressure behavior is never used. Refactor the capacity calculation/conditions or remove the waiter machinery and describe the actual inline fallback behavior.

### 💡 Suggestions

- Add a focused test asserting the intended behavior for repeated overflow refreshes both before and after queue capacity becomes available.
- Run the targeted test suite before merging; the current “keeps later overflow refreshes inline after the queue drains” test should expose the first issue.
- Update the PR description, which still describes oldest-entry eviction even though the current implementation uses serialized draining, retry, and inline overflow handling.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `03610e0ca481af56e788715f4081bb98e6cdf79e`  
Checked at: 2026-09-08 06:02:32 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Aggregate deferred state is capped across pending, retry, active, and active-update maps.
- Drain execution is serialized, preventing overlapping Redis batches.
- Updates to in-flight keys are coalesced and processed afterward.
- Failed batches are retained for retry.
- Overflow warnings are aggregated.
- Tests cover capacity, retry, concurrency, and cleanup paths.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:96`: overflow “suppression” still executes every refresh inline.**

  While `dnsExpireOverflowTs` is active, this branch updates the timestamp and returns `true`. Consequently, every throttled key causes an immediate Redis `EXPIRE`; sustained traffic continually extends the suppression window while producing an unbounded stream of Redis operations.

  This contradicts the PR description’s promise to suppress repeated overflow refreshes. The tests also codify the incorrect behavior by expecting every distinct overflow key—and the same key after draining—to execute inline.

  Coalesce or defer subsequent overflow work within the fixed capacity, or apply backpressure until the active drain releases capacity. Add assertions on actual Redis call counts during sustained overflow.

- **High — `net2/DNSTool.js:88`: throttled updates do not account for keys awaiting retry.**

  After a failed drain, a key remains in `dnsExpireRetry`, but `tryRefreshDnsTTL()` checks only pending and active batches. A subsequent throttled update can therefore execute inline while the stale retry remains queued. On the next drain, that older expiry value can run after the newer inline update and overwrite it.

  Check `dnsExpireRetry` before overflow handling and replace its stored `expr` with the latest value, or otherwise remove the stale retry before returning `true`. Add a regression test using a normal throttled update after failure; the existing test deletes `dnsExpireTs`, so it exercises only the leading-edge cleanup path.

### 💡 Suggestions

- Rename or rewrite the overflow tests so they validate the stated suppression contract rather than expecting every overflow request to run inline.
- Remove the unnecessary `await` on synchronous `tryRefreshDnsTTL()` calls unless the API is intentionally becoming asynchronous.
- Run the targeted test suite and relevant repository checks before merging; the PR states that full validation was unavailable.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `ac39e6f93b7ce0efde6be9aad17a8369b49c50d6`  
Checked at: 2026-09-08 06:47:39 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds aggregate deferred state across pending, retry, active, and coalesced-update maps.
- Serializes Redis drains and preserves failed batches for retry.
- Coalesces newer updates for active and retrying keys.
- Aggregates overflow warnings to prevent log flooding.
- Adds broad coverage for capacity, failure, and in-flight update paths.
- Reviewed both changed files.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:101`: TTL refreshes remain suppressed after queue capacity becomes available.**

  `dnsExpireOverflowTs` is global and checked before current deferred capacity. Once one overflow occurs, every eligible key absent from pending, active, and retry state is rejected for 30 minutes—even immediately after the queue drains.

  This silently drops later TTL updates and contradicts both the documented deferred-refresh contract and the PR description’s claim that normal deferral resumes once capacity becomes available. It is particularly incorrect when a later call supplies a different `expr`.

  Only apply overflow suppression while deferred capacity is still exhausted. When capacity becomes available, enqueue the refresh and clear or otherwise end the overflow window.

- **Medium — `test/test_dns_tool.js:163`: the regression test codifies the incorrect post-drain behavior.**

  “keeps later overflow refreshes suppressed after the queue drains” expects a later refresh to be discarded despite available capacity. Change the test to verify that the refresh enters `dnsExpirePending`, is drained, and uses the latest expiry value.

### 💡 Suggestions

- Add an invariant test confirming aggregate deferred state never exceeds 50,000 during active drains, retries, and sustained overflow.
- Replace repeated test literals such as `50000` with a shared or injectable capacity where practical.
- Remove the duplicated `dnsExpireActiveUpdates.size` assertion in the overlapping-drain test.
- Run the targeted test suite and relevant repository checks before merging; the PR reports that full validation was unavailable.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `362a2308c1af88a06e3cb56021dff8b7cf698ae4`  
Checked at: 2026-09-08 07:06:52 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Aggregate deferred state is capped across pending, retry, active, and coalesced-update maps.
- Redis drains are serialized, avoiding overlapping batches.
- Updates to active and retrying keys are coalesced with the latest expiry.
- Failed batches are retained for retry without replaying stale values.
- Overflow warnings are aggregated.
- Tests cover capacity, failures, concurrent updates, and timer cleanup.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:102`: overflow suppression silently drops TTL refreshes.**

  Once the deferred state reaches 50,000 entries, only the first overflow key is refreshed inline. Every subsequent eligible key during the 30-minute global suppression window returns `false`, but is neither queued nor refreshed inline.

  This means an actively updated DNS key can expire even though new data was recorded. It also contradicts the PR’s stated contract that throttled refreshes are deferred rather than discarded. The global timestamp makes this worse because one overflowing key suppresses unrelated keys.

  Use bounded backpressure, trigger/drain capacity before accepting additional refreshes, or explicitly preserve each suppressed key through a bounded mechanism. If dropping refreshes is intentional, the comments and PR description must document that weaker guarantee and demonstrate that every supported `expr` exceeds the maximum suppression interval.

### 💡 Suggestions

- Add a regression test with a blocked full drain and multiple distinct overflow keys, then verify every accepted update is either refreshed inline or retained for later processing.
- Assert the TTL behavior using an expiry shorter than `RDNS_TTL_REFRESH_PERIOD`; this exposes the data-loss consequence directly.
- Run the targeted suite and relevant repository checks before merging; the PR currently has no reported CI checks.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `d2ee40aeef1be5862f35e91154a288ce6b9f054e`  
Checked at: 2026-09-08 07:48:07 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Bounds aggregate deferred state across pending, retry, active, and coalesced-update maps.
- Serializes Redis drains to prevent overlapping batches.
- Preserves newer expiry values for active and retrying keys.
- Retains failed batches for retry.
- Aggregates overflow warnings.
- Adds extensive capacity, failure, concurrency, and timer-cleanup coverage.
- No linked issue found.

### ⚠️ Issues found

- **High — Overflow traffic can generate unbounded Redis operations.**  
  When deferred capacity is exhausted, every distinct throttled key returns `true`, causing callers to execute `EXPIRE` inline. Repeated calls for the same overflow key also return `true` because no overflow state is retained. A high-cardinality burst while Redis is slow can therefore create an unrestricted stream of commands, contradicting the PR description’s overflow-suppression and availability claims.

  Add bounded backpressure or coalescing so sustained overflow cannot create unlimited inline operations. If inline execution for every overflow is intentional, update the PR description and explicitly document the Redis-load tradeoff.

- **High — The aggregate-capacity regression test will fail.**  
  In `keeps aggregate deferred state within capacity during active overflow`, the test expects only the first of 1,000 overflow calls to return `true`. The implementation returns `true` for all 1,000 because deferred size remains at capacity and there is no suppression mechanism.

  Align the implementation with the intended one-inline-refresh behavior, or change the assertion if every overflow refresh is intentionally inline. Also assert actual Redis call counts under sustained overflow.

### 💡 Suggestions

- Add a stress test with a blocked 50,000-entry drain and many distinct overflow keys, verifying both retained memory and Redis operations remain bounded.
- Avoid globally ignoring `package-lock.json`; lockfiles are normally important for reproducible Node.js dependency resolution. Scope the rule if only a specific generated lockfile should be excluded.
- Run the targeted test suite before merging; the PR currently reports no CI checks.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `b7af23bc158e6c44260516fde20c74fa1c5cf0a1`  
Checked at: 2026-09-08 08:10:34 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Aggregate deferred state is capped across pending, retry, active, and coalesced-update maps.
- Redis drains are serialized, preventing overlapping batches.
- Updates for active or retrying keys retain the latest expiry value.
- Failed batches are preserved for retry.
- Overflow warnings are aggregated rather than emitted per request.
- Tests cover capacity, retry behavior, concurrent updates, and timer cleanup.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:102`: overflow suppression silently drops TTL refreshes.**

  When deferred capacity is exhausted, only the first overflow key is refreshed inline. Every subsequent throttled key during the global 30-minute suppression period returns `false`, but is neither queued nor refreshed.

  This can let a DNS key expire despite receiving new data, especially because callers may supply an expiry shorter than the suppression period. It also contradicts the documented contract that throttled refreshes are deferred rather than dropped.

  Use bounded backpressure or another bounded mechanism that retains each suppressed refresh. If dropping is intentional, explicitly document the weaker guarantee and validate supported expiry values against the suppression period.

- **High — `test/test_dns_tool.js:216`: the warning aggregation test deterministically fails.**

  The first drain logs `refreshed inline: 1` and resets `dnsExpireOverflowCount` to zero. The second drain has no pending or retry entries, so it immediately returns without producing another warning. Therefore the expectation at line 217 for a single `refreshed inline: 3` warning cannot pass.

  Correct the expected warning history or generate another overflow cycle before asserting a second warning.

### 💡 Suggestions

- Add a regression test with a blocked, full drain and multiple distinct overflow keys, verifying that every refresh is either executed or retained.
- Test with an expiry shorter than `RDNS_TTL_REFRESH_PERIOD` to expose dropped-refresh behavior.
- Avoid globally ignoring `package-lock.json`; lockfiles are generally needed for reproducible dependency resolution. Scope the ignore rule to the specific generated location if necessary.
- Remove the unnecessary `await` around synchronous `tryRefreshDnsTTL()` calls, or make the method asynchronous only if that is intentional.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `6291f8e06468747e463789a87a9df4adb7265d6b`  
Checked at: 2026-09-08 08:20:50 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Serializes drain operations and prevents overlapping Redis batches.
- Coalesces updates for pending, active, overflow, and retry states.
- Preserves newer refreshes when a Redis batch fails.
- Adds substantial regression coverage for capacity, retries, and in-flight updates.
- No linked issue found.

### ⚠️ Issues found

- **High — [net2/DNSTool.js:109]** Once `dnsExpireOverflow` reaches 1,000 entries, every additional unique key returns `true`, causing an inline Redis `EXPIRE`. Under sustained high-cardinality traffic, this becomes an unbounded stream of Redis operations—the exact overload scenario the PR claims to suppress. The included test covers only 1,002 overflow calls, which produces two inline calls; extending it beyond 1,002 would expose the issue. When both queues are full, refreshes should remain rate-limited or trigger a controlled drain instead of allowing every subsequent key inline.

- **Medium — [net2/DNSTool.js:116]** `_dnsExpireDeferredSize()` excludes `dnsExpireOverflow`, so the implementation can retain 51,000 deferred keys despite the stated 50,000 deterministic bound. Either include overflow entries in the shared capacity calculation or document and enforce 51,000 as the actual aggregate bound.

### 💡 Suggestions

- Add a sustained-overflow test with significantly more than `MAX_DNS_EXPIRE_OVERFLOW` unique keys and assert that inline Redis calls remain bounded.
- Add an assertion for total retained deferred state across pending, retry, active updates, and overflow—not only `_dnsExpireDeferredSize()`.
- Reconsider ignoring `package-lock.json` repository-wide; lockfiles are generally valuable for reproducible development and CI dependencies.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `9a36474eb5d8aa2068f08952f40a16d9688c29fc`  
Checked at: 2026-09-08 08:36:24 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Serializes drain operations to prevent overlapping Redis batches.
- Coalesces updates for pending, active, overflow, and retry entries.
- Preserves failed batches for retry.
- Adds extensive regression coverage around saturation and failure handling.
- No linked issue found.

### ⚠️ Issues found

- **High — overflow capacity reintroduces unbounded Redis traffic**  
  [`net2/DNSTool.js:120`](https://github.com/firewalla/firewalla/pull/9555/files#diff-7b95b734abfd45164a954305b280dc40d668a62a6cf84b3026f894ec436faca0R120)

  Once `dnsExpireOverflow` reaches 1,000 entries, every additional distinct key returns `true`, causing an inline `EXPIRE`. This bypasses `dnsExpireOverflowTs`, so a sustained high-cardinality workload can generate an unbounded stream of Redis calls—the exact behavior the PR intends to prevent.

  The new test at `test/test_dns_tool.js:582` expects at most two inline calls for 3,000 overflow keys, but the implementation will return `true` for approximately 2,000 of them while the original drain remains unresolved. This test should fail as written.

  The overflow-full branch needs genuine rate limiting or backpressure instead of unconditionally returning `true`. If refreshes cannot be dropped, the design needs another bounded delivery mechanism that does not translate every excess key into an immediate Redis operation.

### 💡 Suggestions

- Add an assertion that counts actual `tryRefreshDnsTTL()` `true` results after overflow capacity is exhausted; this will isolate the faulty branch directly.
- Reconcile the documented 50,000-entry bound with the implementation, which retains up to 51,000 deferred entries through the additional overflow map.
- Consider whether globally ignoring `package-lock.json` is intended; this unrelated change could hide dependency lockfiles that should be committed.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `70078c1997571a417eed86507625d2f9e5a06ced`  
Checked at: 2026-09-08 08:47:55 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Aggregate deferred state is bounded across pending, retry, active, and coalesced-update maps.
- Drain operations are serialized, preventing overlapping Redis batches.
- Newer expiry values are preserved for active and retrying keys.
- Failed Redis batches are retained for retry.
- Overflow warnings are aggregated.
- Tests cover capacity, failures, concurrent updates, and timer cleanup.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:102`: overflow suppression silently drops TTL refreshes.**

  Once deferred capacity is exhausted, only the first overflow key is refreshed inline. Every subsequent eligible key during the global 30-minute suppression window returns `false`, but is neither queued nor refreshed.

  This violates the documented contract that throttled updates are deferred rather than dropped. A DNS entry can expire despite receiving new data, particularly when `expr` is shorter than `RDNS_TTL_REFRESH_PERIOD`. The global timestamp also lets one overflowing key suppress unrelated keys.

  Use bounded backpressure or another delivery mechanism that retains each accepted refresh. If dropping is intentional overload behavior, explicitly document that weaker guarantee and validate that supported TTLs cannot expire during suppression.

- **Medium — `net2/DNSTool.js:127`: a drain can run indefinitely under sustained updates.**

  The drain repeatedly consumes `dnsExpireActiveUpdates` and `dnsExpirePending` until both are empty. If updates arrive continuously, the promise may never settle and a single timer invocation can issue an unbounded sequence of Redis batches.

  Limit each drain invocation to a fixed number of batches or entries, then leave remaining work for the next scheduled drain. This would preserve serialization while placing an actual bound on Redis work per cycle.

### 💡 Suggestions

- Add a regression test with a blocked full drain and multiple distinct overflow keys, verifying that every refresh is either executed or retained.
- Test overflow behavior with an expiry shorter than 30 minutes to expose the dropped-refresh consequence.
- Add a sustained-producer test proving `_drainDnsTTL()` terminates and Redis work per invocation remains bounded.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `b1917edb2d04e2a58e77105d554aa1c79bfc5144`  
Checked at: 2026-09-08 09:37:51 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Coalesces repeated TTL refreshes for pending, active, and retry batches.
- Serializes drains to prevent overlapping Redis operations.
- Preserves failed batches for retry and includes focused regression tests.
- Handles updates arriving during an active drain.
- No linked issue found.

### ⚠️ Issues found

- **High — deferred memory remains unbounded**  
  [`net2/DNSTool.js:96`] Each overflow call enters `_deferDnsTTLWhenCapacityAvailable()` and retains its `key`, `expr`, promise, and async continuation while waiting. Although the maps are capped at 50,000 entries, the number of waiting producers is unlimited. A high-cardinality workload can therefore still cause unbounded memory growth—the retained state has merely moved outside the maps. The test that creates 1,000 waiters demonstrates this behavior but does not assert a bound on waiter count. Use a bounded overflow mechanism, reject/drop according to an explicit policy, or reserve queue capacity synchronously before creating a waiter.

- **Medium — Redis failure can indefinitely block DNS update callers**  
  [`net2/DNSTool.js:105`] When the queue is full and Redis remains unavailable, each drain restores the full retry batch and releases/reawakens all waiters. Those callers loop and wait again indefinitely. Because `addDns()` and `addReverseDns()` now await this path, unrelated DNS updates can accumulate blocked promises and potentially stall upstream processing. Backpressure needs a bounded wait, cancellation/timeout, or a non-blocking overflow policy.

### 💡 Suggestions

- Add a stress test with substantially more overflow producers and assert that total retained work—including waiters—is bounded.
- Add a persistent Redis-failure test verifying that callers terminate or degrade predictably rather than remaining pending forever.
- Consider making `tryRefreshDnsTTL()` consistently asynchronous; its current boolean-or-Promise return type is easy for future callers to misuse without `await`.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `c6543388e76fedf730d578adedc95ac78161d030`  
Checked at: 2026-09-08 10:33:31 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- The design attempts to bound deferred TTL state, serialize drains, coalesce updates, and retain failed batches for retry.
- Tests target saturation, Redis failures, concurrent updates, and timer cleanup.
- No linked GitHub issue was found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:77`: Production code does not parse.**  
  `tryRefreshDnsTTL` is declared twice, with the second declaration beginning inside the first method. There are also unmatched braces and unreachable remnants around lines 119–126. Loading `DNSTool.js` will fail with a syntax error, breaking every consumer of this module.

- **High — `test/test_dns_tool.js:82`: The new test file does not parse.**  
  Several test blocks were accidentally interleaved. For example, a second `it(...)` begins inside the loop opened at line 80, and the object literal beginning at line 160 is never completed before more statements appear. The suite cannot execute and therefore provides no regression coverage.

- **High — `net2/DNSTool.js:177`: Overflow entries are excluded from the capacity calculation.**  
  `_dnsExpireDeferredSize()` counts pending, retry, active updates, and active entries, but omits `dnsExpireOverflow`. Once the main queue reaches 49,000 entries, every additional unique key continues satisfying `deferredSize < MAX_DNS_EXPIRE_PENDING` and is inserted into `dnsExpireOverflow`. This makes the overflow map unbounded and defeats the PR’s primary memory-safety objective. Include `dnsExpireOverflow.size` in the total or enforce its independent 1,000-entry limit directly.

- **High — Requirement coverage is not met.**  
  The PR description requires a deterministic 50,000-entry bound, executable regression coverage, and preserved TTL-refresh behavior. The omitted overflow accounting violates the bound, while syntax errors prevent both implementation and tests from running.

### 💡 Suggestions

- Remove the duplicated implementations and resolve the apparent bad merge before further review.
- Add a syntax/lint check to CI so malformed JavaScript cannot reach review.
- After repairing the files, test the combined cardinality of pending, overflow, retry, active, and active-update state—not each container in isolation.
- Update the PR description after settling the overload policy; it currently describes inline handling and preservation of refreshes, while the submitted implementation intentionally drops excess refreshes.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `29dd5f3249e59576f84833b7ab262d6fd4f85c29`  
Checked at: 2026-09-08 10:53:13 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Enforces a combined 50,000-entry bound across pending, overflow, retry, active, and active-update state.
- Serializes drains and preserves failed batches for retry.
- Correctly coalesces newer TTL values, including updates arriving during an active drain.
- Aggregates overload warnings and avoids saturation-triggered Redis writes.
- Clears deferred state when records are removed or superseded.
- Adds strong focused coverage for capacity, concurrent producers, failures, retries, and cleanup.
- No linked issue found.

### ⚠️ Issues found

No blocking correctness, security, concurrency, performance, or resource-management issues found in the three changed files.

### 💡 Suggestions

- The workflow only runs `node --check`; execute `test/test_dns_tool.js` in CI as well. Syntax validation will not catch regressions in capacity accounting, retry ordering, or active-update handling.
- Consider adding an operational metric for `dnsExpireDroppedCount`. The aggregated warning is useful, but a metric would make sustained TTL-refresh loss easier to detect and alert on.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `31d54a681422cb4b6087553229baaaeea51a7c91`  
Checked at: 2026-09-08 11:19:07 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Enforces a combined 50,000-entry bound across all deferred-state maps.
- Serializes drains and retains failed batches for retry.
- Coalesces newer TTL values, including updates during active drains.
- Aggregates overload warnings.
- Adds comprehensive tests for capacity, failures, concurrent updates, and cleanup.
- No linked issue found.

### ⚠️ Issues found

- **High — `net2/DNSTool.js:82`: LRU eviction can bypass overload protection.**

  `dnsExpireTs` holds only 50,000 timestamps. With more than 50,000 distinct keys, timestamps for still-retained pending, overflow, or retry entries can be evicted. Revisiting such a key makes `last` undefined, so the leading-edge branch removes its deferred entry and returns `true`, causing an inline Redis `EXPIRE`.

  Cycling through a sufficiently large key set can therefore produce unbounded inline Redis traffic during saturation, contradicting the stated overload policy.

  Check retained queue membership before relying on `dnsExpireTs`, or preserve throttle metadata for every retained entry independently of the evicting LRU. Add a regression test that fills the deferred state, churns `dnsExpireTs` past its capacity, and revisits an evicted-but-still-queued key.

### 💡 Suggestions

- Run the focused Mocha suite in CI; `node --check` only detects syntax errors and will not catch capacity, retry, or coalescing regressions.
- Export a metric for dropped refreshes so persistent overload is observable beyond aggregated logs.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `094ae6f280cfec8fbbffe86fd411818e1e89caf9`  
Checked at: 2026-09-08 12:06:17 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Enforces the intended 50,000-entry aggregate bound across pending, overflow, retry, active, and active-update state.
- Prevents LRU eviction from converting retained deferred work into unexpected inline Redis operations.
- Serializes drains and preserves failed batches for retry.
- Coalesces repeated updates while retaining the latest TTL.
- Aggregates dropped-refresh warnings to avoid log flooding.
- Adds thorough coverage for saturation, retries, concurrent producers, LRU churn, and cleanup.
- CI performs syntax, lint, and full coverage-test execution.
- No linked issue found.

### ⚠️ Issues found

No blocking correctness, security, concurrency, performance, or resource-management issues found in the three changed files.

### 💡 Suggestions

- Consider exporting `dnsExpireDroppedCount` through operational metrics; logs alone may make sustained overload difficult to monitor.
- Document the intentional best-effort drop policy near the relevant operational configuration, since dropped refreshes can preserve an older expiration deadline.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `d4bb1fe1abb070ff690a5edc7d98848280e50fe8`  
Checked at: 2026-09-08 12:32:40 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Enforces the intended 50,000-entry aggregate bound across pending, overflow, retry, active, and active-update state.
- Correctly coalesces repeated updates and preserves newer TTL values during active or failed drains.
- Serializes drains and retains failed batches for retry without creating unbounded waiters.
- Handles timestamp-LRU eviction without converting retained work into unexpected inline Redis traffic.
- Aggregates dropped-refresh warnings and explicitly documents the best-effort overload policy.
- Tests cover saturation, concurrency, retries, failures, LRU churn, and cleanup.
- No linked issue found.

### ⚠️ Issues found

- **Medium — `net2/DNSTool.js:195`: A single Redis transaction may contain up to 49,000 commands.**

  `_drainDnsTTL()` adds the entire selected map to one `MULTI/EXEC`. During overload, this can produce a large request and reply allocation and monopolize the shared Redis server while the transaction executes. The new memory bound therefore still permits a significant latency spike.

  Drain in fixed-size chunks—such as several hundred or a few thousand commands—while retaining serialization and the per-invocation work limit.

### 💡 Suggestions

- Add a test proving that large queues are split into bounded Redis transactions.
- Export the dropped-refresh count as an operational metric; aggregate logs alone may be insufficient for alerting on sustained overload.
- Consider documenting why 49,000 main entries and 1,000 overflow entries are appropriate for appliance memory and Redis latency constraints.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `284c2209812663a5c88a4b511ca0e9afcfb6bfe8`  
Checked at: 2026-09-10 01:52:57 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Enforces a strict 50,000-entry bound across pending, overflow, retry, active, and active-update state.
- Serializes drains and retains failed batches for retry without blocking producers.
- Correctly coalesces repeated updates and preserves newer updates when an active batch fails.
- Prevents LRU eviction from accidentally converting retained deferred work into inline Redis traffic.
- Aggregates overload warnings instead of logging once per dropped refresh.
- Adds strong tests for saturation, concurrent producers, retry behavior, removal, and bounded accounting.
- No linked issue found.

### ⚠️ Issues found

No blocking correctness, security, concurrency, or resource-management issues found in the three changed files.

### 💡 Suggestions

- The workflow only performs syntax checks on the new test file; it does not execute the test suite. Add the focused Mocha test to CI so behavioral regressions are detected.
- Consider exposing `dnsExpireDroppedCount` through operational metrics. A periodic warning may be insufficient to detect sustained TTL-refresh loss.
- Document why the 49,000/1,000 capacity split is appropriate for expected appliance memory and Redis latency constraints.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `20d1d1d4edbd2b7874cb6aa12992a1a46ecbc1b3`  
Checked at: 2026-09-11 01:05:28 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Enforces the intended 50,000-entry aggregate bound across pending, overflow, retry, active, and active-update state.
- Serializes drains and preserves failed batches for retry.
- Coalesces repeated updates while retaining the latest TTL.
- Prevents timestamp-LRU eviction from bypassing overload protection.
- Aggregates dropped-refresh warnings and exports the count through metrics.
- Adds comprehensive tests for saturation, concurrent producers, failures, retries, and cleanup.
- No linked issue found.

### ⚠️ Issues found

- **Medium — `net2/DNSTool.js:165`: A drain can submit up to 49,000 commands in one Redis transaction.** Building and executing one `MULTI/EXEC` for the entire main queue can create large request/reply allocations and monopolize the shared Redis server, producing a significant latency spike on constrained appliances. Drain in fixed-size chunks while preserving serialization and bounded work per invocation.

### 💡 Suggestions

- Add a test confirming that a saturated queue is divided into bounded Redis transactions.
- Execute `test/test_dns_tool.js` in CI. The added `node --check` commands validate syntax only; they do not catch regressions in capacity accounting, coalescing, or retry behavior.
- Validate the 49,000/1,000 capacity split and transaction size through appliance profiling before deployment.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9555  
Head SHA: `f176ab903bf6ab61f8e721c39ab4d4816e1a2fa0`  
Checked at: 2026-09-11 09:55:13 CST

PR #9561, opened 2026-09-03
category: make user target list deactivation decide in FireMain

wzjjack:
Deleting the last rule/exception on a user target list (TL-*) still left CategoryUpdateSensor polling its hashset, so a box removed from an MSP kept logging "Fail to fetch category list from cloud" forever.

Both guards of the exception path read FireMain-only in-memory state: hasActivePolicies() reads activeCategoryPolicyMap and deactivateCategory() early-returns on isActivated(), which reads activeCategories. Exception deletes run in FireApi (netbot exception:delete, Guardian.reset), where both are always empty, so the path never emitted Category:Delete and FireMain kept the category active.

The rule path runs in FireMain and works, but its unenforce is dispatched cross-process without awaiting, so when a rule and an exception reference the same target list the two deletes race: whichever runs first sees the other one still present and skips deactivation.

Funnel both paths through ExceptionManager.maybeDeactivateCategory(), which hands the decision to FireMain via a new Category:MaybeDeactivate event when called elsewhere. Both deletes now re-run the same idempotent check, so whichever lands last performs the cleanup regardless of order.

Also:
- Guardian.reset() now sends ExceptionChange like netbot does on exception:delete, otherwise FireMain keeps a stale CategoryMatcher that re-activates the target list on the next UPDATE_CATEGORY_DOMAIN.
- Category:Delete disables the cloudcache item of a user target list; cloudcache's own job refreshes registered items regardless of whether the category is still active.

Fixes firewalla/firecommit#9299

### DIFF
--- a/alarm/ExceptionManager.js
+++ b/alarm/ExceptionManager.js
@@ -120,10 +120,22 @@ module.exports = class {
     return exceptions.some(e => e.getCategory() === category);
   }
 
+  // Whether any rule or exception still needs this category. Must be re-evaluated right before
+  // the destructive cleanup, not just when the deactivation is decided: the two are separated by
+  // an event hop, and a rule/exception can reference the target list again in between.
+  async isCategoryReferenced(category) {
+    if (!category) return false;
+    if (categoryUpdater.hasActivePolicies(category)) return true;
+    if (await this.hasException(category)) return true;
+    // lazy require, PolicyManager2 requires this module at load time
+    const PolicyManager2 = require('./PolicyManager2.js');
+    return new PolicyManager2().hasPersistedCategoryPolicy(category);
+  }
+
   // User target list categories are only kept alive by the rules/exceptions referencing them.
-  // Both checks below read FireMain-only in-memory state (activeCategoryPolicyMap, activeCategories),
-  // so other processes must hand the decision over instead of silently no-op'ing here.
-  // Rule delete and exception delete both funnel through this, so whichever lands last cleans up.
+  // isCategoryReferenced() and deactivateCategory() both read FireMain-only in-memory state, so
+  // other processes must hand the decision over instead of silently no-op'ing here. Rule delete
+  // and exception delete both funnel through this, so whichever lands last cleans up.
   async maybeDeactivateCategory(category) {
     if (!category || !categoryUpdater.isUserTargetList(category)) return;
     if (!firewalla.isMain()) {
@@ -135,8 +147,7 @@ module.exports = class {
       });
       return;
     }
-    if (categoryUpdater.hasActivePolicies(category)) return;
-    if (await this.hasException(category)) return;
+    if (await this.isCategoryReferenced(category)) return;
     await categoryUpdater.deactivateCategory(category);
   }
 
--- a/alarm/ExceptionManager.js
+++ b/alarm/ExceptionManager.js
@@ -120,11 +120,21 @@ module.exports = class {
     return exceptions.some(e => e.getCategory() === category);
   }
 
-  // user target list categories are only kept alive by rules/exceptions referencing them;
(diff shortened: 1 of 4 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Centralizes target-list deactivation in FireMain, where the authoritative policy/category state lives.
- Handles both rule and exception deletion ordering.
- Disables the cloud-cache entry when removing a user target list.
- Refreshes exception matchers after MSP exceptions are removed.
- Linked issue firecommit#9299 is mostly covered: MSP cleanup now removes stale target-list polling.

### ⚠️ Issues found

[alarm/ExceptionManager.js:140]

**Medium — Deactivation has a check-then-act race.** After `hasActivePolicies()` and `hasException()` return false, another rule or exception can activate the category before the asynchronous `Category:Delete` handler completes. That handler removes the category without rechecking references, potentially disabling a target list that is already needed again.

Serialize activation/deactivation or revalidate policy and exception references immediately before destructive cleanup.

This makes the linked issue requirement **partially met**: stale polling is removed, but concurrent rule creation can cause incorrect deactivation.

### 💡 Suggestions

  - Rule deleted before exception.
  - Exception deleted before rule.
  - MSP reset removing both.
  - A new rule created while deactivation is pending.
  - Cloud-cache registration removed after the last reference disappears.

- Add semicolons after the new event listeners for consistency.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9561  
Head SHA: `ee4c7addbe2c2dad42a54e8214ae167f3146dafb`  
Checked at: 2026-09-03 16:18:51 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly routes deactivation decisions from FireApi/FireMon to FireMain.
- Rechecks references before destructive cleanup.
- Disables the cloud-cache entry, addressing the repeated fetches reported by linked issue firecommit#9299.
- Sends `ExceptionChange` after Guardian removes MSP exceptions.
- Issue requirement coverage: **Partially met**. The primary removal flow is addressed, but multi-target rules and a remaining check/cleanup race can still cause incorrect deactivation.

### ⚠️ Issues found

[alarm/PolicyManager2.js:634]  
**High:** `hasPersistedCategoryPolicy()` checks only `p.target`. Category policies also support multiple categories through `p.targets`. During the enforcement window this persisted lookup is the only protection before `activeCategoryPolicyMap` is updated, so a newly persisted multi-target rule can be missed and its category data flushed. Check both `p.target === target` and `p.targets?.includes(target)`.

[sensor/CategoryUpdateSensor.js:653]  
**Medium:** The reference check and cleanup remain non-atomic. An exception or policy can be persisted after `isCategoryReferenced()` completes but before `disableCache()` and the Redis flushes execute. Because the category is still marked active during that interval, concurrent `activateCategory()` may return early without scheduling a reload, after which this handler removes the active state and data. Serialize category activation/deactivation with a per-category lock or generation/version check covering the final validation and state transition.

### 💡 Suggestions

- Add regression tests for:
  - A newly created multi-target category policy racing with deletion.
  - Exception and rule edits that delete and immediately recreate the reference.
  - Activation occurring during `Category:Delete`.
  - Removal of the final MSP rule and exception confirming cloud-cache polling stops.
- Catch errors around the complete `Category:Delete` async handler to avoid unhandled promise rejections from the new Redis-backed validation.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9561  
Head SHA: `35151dab906a5e7ce2abe0eacb57ac8ad5165e1b`  
Checked at: 2026-09-03 16:51:26 CST

jasonlyc (approved):

PR #9576, opened 2026-09-04
sync purple 1.984 staging

jasonlyc:


MelvinTo (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly fixes global-rule detection by using `policy.tag`; firecommit#9361 is met and covered by a regression test.
- User target-list cleanup is centralized in FireMain and addresses the rule/exception deletion race described in firecommit#9299.
- AP live statistics satisfy firecommit#9381: aggregate throughput, Ethernet ports, wireless bands, and uplink RSSI are returned.
- Port probing avoids unsupported `ss` filter syntax while preserving per-protocol failure handling.
- Block-stat range semantics are clearly documented as `[begin, end)`.

### ⚠️ Issues found

- **High — paired-app notifications can be permanently lost**

  [net2/PairedAppEventTool.js:67]

  `claimPending()` deletes the pending record before `emitPairedEvent()` completes. If Redis event persistence or another operation inside `emitPairedEvent()` fails, the caller reports an error but the backstop sensor can no longer retry, so the `phone_paired` event is permanently lost.

  Claim the record atomically into an in-flight state and delete it only after successful emission, or restore/requeue it when emission fails. A test should simulate `emitPairedEvent()` failure and verify that a later flush retries the event.

- **Medium — the claim is not atomic with replacement of the same EID**

  [net2/PairedAppEventTool.js:63]

  The separate `HGET` and `HDEL` permit this sequence: a caller reads an old record, another pairing overwrites that EID, and the first caller deletes the new record before emitting the old payload. Use a Lua script or transaction that conditionally removes the exact value that was read.

- **Low — block-stat range validation accepts non-finite and reversed ranges**

  [controllers/netbot.js:1471]

  `isNaN()` accepts values such as `Infinity`, and there is no validation that `begin < end`. Invalid bounds can reach Redis, where `_getBlockStatsBuckets()` catches the resulting error and returns an indistinguishable empty result. Convert first, require `Number.isFinite(begin/end)`, and reject `begin >= end`.

### 💡 Suggestions

- Add focused tests for concurrent `claimPending()` calls, emission failure/retry, malformed pending records, and unpair/re-pair reuse of the same EID.
- Add AP status and aggregation tests covering unavailable `fwapc`, stale timestamps, non-numeric rates, and an existing 6 GHz `uplink`.
- The PR contains 29 changed files and several unrelated fixes. Splitting future synchronization PRs by feature would reduce regression risk and make requirement verification easier.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9576  
Checked at: 2026-09-04 19:17:56 CST

PR #9578, opened 2026-09-04
sync gse 1.984 staging

jasonlyc:


MelvinTo (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Reviewed all 29 changed files.
- firecommit#9299: Met — user target-list deactivation is delegated to FireMain, and cloud-cache polling is disabled during category deletion.
- firecommit#9361: Met — global-rule filtering now checks the actual `policy.tag` field and includes regression coverage.
- firecommit#9381: Partially met — AP live statistics aggregate Ethernet and Wi-Fi rates and expose uplink RSSI, but the offline response schema is incomplete.
- Network-sensitive changes correctly encode URL path parameters and use daemon-backed country ipset updates.
- The port checker preserves valid per-protocol results when the other probe fails.

### ⚠️ Issues found

[net2/PairedAppEventTool.js:70]

**Medium — The pending event is deleted before durable publication.** After `HDEL`, a process exit, Redis publish failure, or exception before `emitPairedEvent()` completes permanently loses the `phone_paired` event. `EventRequestApi.addActionEvent()` also catches publication failures internally, so `claimPending()` can report success despite the event never being stored. This contradicts the documented exactly-once behavior. Keep the record until publication is acknowledged, or use an atomic state transition/outbox that supports retry without duplicate delivery.

[sensor/LiveStatsPlugin.js:212]

**Medium — Offline AP responses omit the required `ports` and `bands` fields.** Online and live-stats-error paths return both arrays, while an AP considered offline returns only `{ target, type, tx, rx }`. This makes the API schema state-dependent and does not fully satisfy firecommit#9381’s required `ports: []` / `bands: []` format. Return empty arrays for offline APs as well.

[controllers/netbot.js:1471]

**Low — `blockStats` accepts invalid ranges.** `isNaN()` allows empty strings, booleans, infinities, negative timestamps, fractional values, and `begin >= end`. These inputs can produce surprising Redis score queries instead of a clear client error. Convert first, require finite non-negative integer timestamps, and require `begin < end`.

### 💡 Suggestions

- Add concurrency and failure-path tests for `claimPending()`, including simultaneous claims, publication failure, and interruption after claiming.
- Add an offline-AP test that asserts the complete stable response shape.
- Add boundary tests for `blockStats`, especially an exclusive `end`, invalid numeric inputs, and reversed ranges.
- Avoid mutating fwapc’s response objects in `formatAPThroughput()` when attaching uplink data; return copied entries to prevent side effects if the response is reused.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9578  
Checked at: 2026-09-04 19:13:02 CST

PR #9581, opened 2026-09-04
sync gp sfp 1.984 staging

jasonlyc:


MelvinTo (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Correctly fixes global-rule filtering for firecommit#9361 using `policy.tag`.
- AP live-stats aggregation cleanly separates Ethernet ports and Wi-Fi bands.
- User target-list deactivation is delegated to FireMain and cloud-cache polling is disabled during cleanup.
- Port checking now tolerates failures on one protocol.
- Block-stat range queries correctly use `[begin, end)` Redis bounds.
- Added focused regression tests for RuleStats, netbot notifications, and AP formatting.

### ⚠️ Issues found

**Medium — phone-paired events can be permanently lost**

[net2/PairedAppEventTool.js:67]

`claimPending()` deletes the only pending record before publishing the action event. `EventRequestApi.addActionEvent()` catches Redis publish failures internally, so this method still returns success and the sensor cannot retry. A transient Redis/publisher failure therefore loses the event permanently while the notification may still be sent.

Use a recoverable processing state or an atomic claim/ack workflow, deleting the record only after confirmed publication. The publishing API must also propagate failures.

**Medium — wireless uplink data is dropped unless the backhaul happens to use 6 GHz**

[sensor/LiveStatsPlugin.js:449]

`formatAPThroughput()` attaches the top-level uplink only to a `6g` interface. APs using 5 GHz wireless backhaul—or responses without a 6 GHz entry—lose the required uplink RSSI entirely. This only partially satisfies firecommit#9381’s requirement to expose wireless uplink RSSI.

Associate the uplink with the actual backhaul interface/band, or preserve it at the AP result level when the interface cannot be determined.

**Low — offline AP response violates the new response shape**

[sensor/LiveStatsPlugin.js:214]

The offline path omits `ports` and `bands`, while online and unavailable-live-stats paths return both arrays. Clients consuming the documented firecommit#9381 schema would need an unnecessary special case. Return `ports: []` and `bands: []` here as well.

### 💡 Suggestions

- Add concurrency and publish-failure tests for `PairedAppEventTool.claimPending()`.
- Add AP live-stats tests for 5 GHz backhaul, missing 6 GHz interfaces, offline APs, and failed fwapc requests.
- Issue requirement coverage:
  - firecommit#9361: **Met**.
  - firecommit#9299: **Met by implementation**, though automated coverage for cross-process deletion ordering would reduce regression risk.
  - firecommit#9381: **Partially met** because uplink RSSI is hardcoded to 6 GHz and the offline response omits required arrays.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9581  
Checked at: 2026-09-04 19:04:31 CST

PR #9584, opened 2026-09-04
suspend dmz allow-source rules in non-router model

hhirokazu:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- DMZ policy state is synchronized during initialization and mode transitions.
- `dmzModeSuspended` distinguishes automatically suspended rules from rules already disabled beforehand.
- Policy updates occur under the existing shared lock.
- Reviewed 1 of 1 changed files.
- No linked issue found.

### ⚠️ Issues found

**Medium — An explicit user disable can be undone by a later router-mode transition.**

[extension/portforward/portforward.js:473]

When non-router mode automatically disables a rule, it retains `dmzModeSuspended="1"`. If the user subsequently explicitly disables that already-disabled rule, `disablePolicy()` returns early and does not clear the marker. Returning to router mode therefore re-enables a rule the user intended to keep disabled.

The automatic suspension marker should be cleared when an explicit policy update/disable occurs, or the original enabled state should be tracked separately from the normal `disabled` field.

### 💡 Suggestions

- Add tests covering:
  - Enabled rule → non-router → router.
  - Initially disabled rule across both transitions.
  - Rule explicitly disabled while automatically suspended.
  - Repeated/idempotent mode transitions.
- Consider using the standard policy enable/disable operations or a dedicated temporary-suspension mechanism so policy lifecycle side effects remain centralized.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9584  
Head SHA: `3e5e1ae0e4efdb31b99c44a665ce642ede60efd1`  
Checked at: 2026-09-05 07:17:31 CST

jasonlyc (approved):

PR #9585, opened 2026-09-07
Close CBC padding oracle in unauthenticated encrypted-message API

MelvinTo:
Fixes firewalla/firecommit#9400.

The external `/v1/encipher/message/:gid` endpoint distinguished CBC padding failures (412) from valid-padding-but-invalid-JSON (400), giving an unauthenticated client a padding oracle against the group key's CBC envelopes.

### Changes
- **Uniform errors**: every failure in `decryptRequest` (bad envelope, bad padding, GCM auth failure, oversized input, post-decrypt JSON parse failure) rejects with the same opaque `decrypt_error`; the middleware returns one uniform `412 {error: "decrypt_error"}` with details logged locally only. Guardian's web-message path picks up the same uniform behavior automatically.
- **Anti-downgrade (TOFU), gated by the `gcm_downgrade_protection` feature (default off = monitor-only)**: on the unauthenticated local API path, a group's first successful GCM request marks it migrated (`sys:ept:gcmGids`, in-memory + redis). With the feature **off** (default), a CBC/legacy request for a migrated group is logged as a would-be rejection but still accepted — safe for mixed groups during migration. With the feature **on**, it is rejected. Roll back a group with `redis-cli hdel sys:ept:gcmGids <gid>`. The Guardian/MSP path does not enforce this since other legitimate clients of the same group may still use CBC.
- **Pre-decryption limit**: 5MB cap on encrypted input (matches the body-parser limit) before any crypto/parse work.
- **Streaming**: SSE frames now mirror the request scheme instead of forcing legacy zero-IV CBC; a GCM client gets GCM envelope frames (legacy frames are bare base64 and never start with `{`, so clients can distinguish).

### Compatibility
- CBC-only app versions keep working: enforcement only starts after the group's first successful GCM request.
- Mixed groups (one updated GCM app + one old CBC app on the same gid): unaffected by default (monitor-only). Once `gcm_downgrade_protection` is turned on, the old app is rejected after the new one's first GCM request — the intended downgrade protection; the redis flag is the escape hatch.

### Validation
- `node --check` on all touched files.
- Self-contained harness (based on the issue's verification script, extended with rclient/Constants stubs) run against this checkout: uniform 412 + identical bodies for bad-padding vs valid-padding/bad-JSON fixtures, valid CBC & GCM accepted pre-migration, CBC/legacy rejected + GCM accepted post-migration, flag persisted, oversize rejected, Guardian path still accepts CBC, streaming mirrors GCM.
- Regression tests added in `test/test_encipher_oracle.js` (mocha; needs a box/dev env with node_modules + redis).

### DIFF
--- a/api/lib/Encryption.js
+++ b/api/lib/Encryption.js
@@ -56,9 +56,11 @@ module.exports = class {
       }
     }
 
-    // decryptRequest parses the { iv, message } envelope once and returns usedIv;
-    // a request that carried an iv gets its reply mirrored with a fresh iv.
-    cloudWrapper.getCloud().decryptRequest(gid, message).then(({ decrypted: decryptedMessage, scheme }) => {
+    // decryptRequest parses the { iv, message } envelope once and reports the
+    // request scheme so the reply mirrors it. enforceGcmPolicy applies the
+    // anti-downgrade rule on this unauthenticated path: once a group has made a
+    // successful GCM request, CBC/legacy requests for it are rejected.
+    cloudWrapper.getCloud().decryptRequest(gid, message, { enforceGcmPolicy: true }).then(({ decrypted: decryptedMessage, scheme }) => {
       req.reqScheme = scheme;
       decryptedMessage.mtype = decryptedMessage.message.mtype;
       req.body = decryptedMessage;
@@ -66,11 +68,11 @@ module.exports = class {
       log.debug(req.id, 'Message decrypted')
       next();
     }).catch((err) => {
-      if(err && err.message === "decrypt_error") {
-        res.status(412).json({"error" : err});
-      } else {
-        res.status(400).json({"error" : err});
-      }
+      // All decrypt/parse failures must be externally indistinguishable — a
+      // status or body that varies with padding validity is a CBC padding
+      // oracle. Log details locally, return one uniform error.
+      log.error('Failed to decrypt request:', err && err.message);
+      res.status(412).json({"error" : "decrypt_error"});
     });
   }
 
@@ -88,9 +90,11 @@ module.exports = class {
       return;
     }
 
-    // Mirror the request scheme (gcm / cbc-iv / legacy) on the reply. Streaming
-    // stays on the legacy zero IV for now (its SSE frame carries no envelope).
-    const scheme = streaming ? 'legacy' : (req.reqScheme || 'legacy');
+    // Mirror the request scheme (gcm / cbc-iv / legacy) on the reply, streaming
+    // included: a client that sent a GCM request gets GCM SSE frames (the data
+    // field carries the JSON envelope; legacy frames were bare base64, which
+    // never starts with '{', so clients can tell them apart).
+    const scheme = req.reqScheme || 'legacy';
 
     // log.info('Response Data:', JSON.parse(body));
     const time = process.hrtime();
--- a/encipher/lib/encipherio.js
+++ b/encipher/lib/encipherio.js
@@ -971,9 +971,18 @@ let legoEptCloud = class {
           }
           const scheme = this._schemeOf(env);
           if (opts.enforceGcmPolicy && scheme !== 'gcm' && await this.isGcmMigrated(gid)) {
-            log.error(`Rejecting ${scheme} request for GCM-migrated group ${gid}`);
-            reject(new Error("decrypt_error"));
-            return;
+            // gcm_downgrade_protection gates enforcement: off (default) is
+            // monitor-only — log the would-be rejection but accept the request,
+            // so mixed groups (an updated GCM app alongside an older CBC-only
+            // app) keep working while the fleet migrates. Turn the feature on
+            // to enforce and actually reject the downgrade.
+            if (config.isFeatureOn("gcm_downgrade_protection")) {
+              log.error(`Rejecting ${scheme} request for GCM-migrated group ${gid}`);
+              reject(new Error("decrypt_error"));
+              return;
+            } else {
+              log.warn(`Monitor: ${scheme} request for GCM-migrated group ${gid} would be rejected if gcm_downgrade_protection were enforced`);
+            }
           }
           let decrypted;
           try {
--- a/encipher/lib/encipherio.js
+++ b/encipher/lib/encipherio.js
@@ -58,6 +58,10 @@ const notificationResendKey = "notification:resend";
 const notificationResendDuration = fConfig.timing['notification.resend.duration'] || 86400
 const notificationResendMaxCount = fConfig.timing['notification.resend.maxcount'] || 50
 
+// pre-decryption cap on unauthenticated encrypted-message input (chars of
+// base64); the API body parser allows 5mb, this bounds crypto/parse work.
+const MAX_ENCRYPTED_MSG_LEN = 5 * 1024 * 1024;
+
 function getUserHome() {
   return process.env[(process.platform == 'win32') ? 'USERPROFILE' : 'HOME'];
 }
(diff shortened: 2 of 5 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Decrypt, padding, authentication, envelope, and JSON failures now produce the same external response, closing the observable padding oracle.
- GCM responses, including SSE streaming frames, mirror the request scheme.
- Migration state is persisted in Redis and compatibility behavior is documented.
- Linked issue `firewalla/firecommit#9400` coverage:
  - **Met:** Uniform external decrypt/parse failures.
  - **Met:** Valid GCM requests and GCM streaming responses.
  - **Met:** Compatibility behavior is documented.
  - **Partially met:** Anti-downgrade enforcement is implemented but disabled by default.
  - **Partially met:** The size limit does not cover every accepted envelope representation.

### ⚠️ Issues found

- **High — Anti-downgrade protection remains inactive by default.**  
  `[net2/config.json:1171]`  
  `gcm_downgrade_protection` defaults to `false`, so migrated groups continue accepting unauthenticated CBC indefinitely unless deployment configuration changes. This does not fully satisfy the linked issue’s requirement to reject CBC for migrated clients. Enable enforcement by default after a defined migration window, or document and implement an explicit activation/sunset mechanism.

- **Medium — Migration persistence can fail after the request succeeds.**  
  `[encipher/lib/encipherio.js:930]`  
  `markGcmMigrated()` updates memory and starts the Redis write without awaiting it. If persistence fails, the GCM request still succeeds, but a restart loses the migration state and re-enables CBC downgrade. Make persistence part of the successful migration transaction—preferably await the Redis write and fail closed when enforcement is enabled.

- **Medium — The supplied regression test references undefined variables.**  
  `[test/test_encipher_oracle.js:144]`  
  `ept2` and `gid2` are not declared in the supplied diff, causing a `ReferenceError` and failing the test suite. Construct a separate unmigrated instance/group before asserting that failed GCM authentication does not mark it migrated.

- **Low — The pre-decryption limit only measures the ciphertext field for object input.**  
  `[encipher/lib/encipherio.js:960]`  
  An already-parsed envelope can contain large `iv`, `tag`, `alg`, or unrelated properties while `approxLen` remains below the limit. Validate the complete serialized envelope size—or enforce bounded lengths for every accepted field—to ensure malformed envelopes are consistently constrained.

### 💡 Suggestions

- Add tests for Redis persistence failure and migration behavior after creating a fresh `EptCloud` instance.
- Test the actual HTTP middleware response body and status for every failure class, not only `decryptRequest()` rejection messages.
- Add an operational rollout plan specifying when enforcement becomes default and when CBC support will be removed.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9585  
Head SHA: `53f34092db3ee41930107b2a5a30b3dbcc27f242`  
Checked at: 2026-09-07 11:22:04 CST

MelvinTo:
## Testbed validation — goldplus (Gold Plus, x86_64 jammy, node 12.14)

Tested on the goldplus testbed: switched to `master` (44ca5494a), applied this PR's diff, restarted FireApi, and ran live requests against the running service on port 8833 with the box's real group key.

### Results

| Test | Result |
|---|---|
| Oracle probe A: valid padding + non-JSON plaintext | `412 {"error":"decrypt_error"}` |
| Oracle probe B: invalid padding (corrupt IV, same one-block fixture) | `412 {"error":"decrypt_error"}` — byte-identical to probe A, **oracle closed** |
| Valid CBC (random-IV envelope) | `200`, ~12ms, reply mirrored as fresh-IV `{iv, message}` envelope |
| Valid legacy (zero-IV bare base64) | `200`, reply mirrored as legacy |
| Valid GCM | `200`, reply mirrored as GCM envelope; `sys:ept:gcmGids` flag set |
| CBC after GCM, monitor mode (feature off, default) | accepted (`200`) + WARN logged: `Monitor: cbc-iv request for GCM-migrated group … would be rejected if gcm_downgrade_protection were enforced` |
| CBC/legacy after GCM, enforced (`hset sys:features gcm_downgrade_protection 1`) | `412 {"error":"decrypt_error"}` |
| GCM with enforcement on | `200`, full E2E round-trip: decrypted reply `{"replyid":"final","code":200,"data":{"uptime":71.5,…}}` |
| Oversized body (>5MB) | `413` from body-parser before decryption |

### Notes

- The box was cleaned up afterward: feature + migration flags cleared, FireApi restarted and sanity-probed (garbage message → 412).
- Two pre-existing master behaviors observed, unrelated to this PR: a decrypted message with no `message.mtype` never gets an HTTP response (request hangs until client timeout), and `ping` is only a `cmd` item, not `get`.

MelvinTo:
Addressed the review findings in d547a08a4:

- **Migration persistence (Medium)**: `markGcmMigrated()` is now persist-then-cache — the redis `hset` is awaited and the in-memory flag is only set after it succeeds, so a failed write is retried on the next GCM request instead of silently degrading to memory-only state that a restart would lose. With `gcm_downgrade_protection` **on**, a persistence failure fails closed (the GCM request is rejected with the uniform `decrypt_error`); in monitor mode (default) the request is accepted and the write retries later.
- **Test variables (Medium)**: `gid2`/`ept2` were declared in the committed file (the diff view may have clipped them), but the test now constructs a genuinely separate `EptCloud` instance + group as suggested, and additionally asserts `isGcmMigrated()` stays false.
- **Envelope field bounds (Low)**: every accepted envelope field is now bounded before any crypto work — `ct` must be a string ≤ the 5MB cap, `iv`/`tag` ≤ 64 chars, `alg` ≤ 16 — closing the gap where an already-parsed object envelope could carry oversized fields while `approxLen` (ciphertext only) stayed under the limit. Out-of-bounds envelopes get the same uniform `decrypt_error`.

New regression tests cover all three (oversized iv/tag/alg, non-string ct, persistence-failure in both modes), and the vm verification harness passes with the added checks: fail-closed under enforcement, monitor-mode acceptance without memory-marking, successful retry persisting the flag afterward.

(The High finding is intentionally not addressed in this PR per discussion.)

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Decrypt, padding, authentication, envelope, size, and JSON failures now return the same external error.
- Input bounds are applied before cryptographic processing.
- GCM migration state is persisted before being cached.
- Streaming responses preserve the authenticated GCM scheme.
- Regression tests cover CBC/GCM behavior, malformed envelopes, persistence failures, and migration policy.
- Linked issue #9400: uniform errors, resource limits, streaming support, documentation, and most regression coverage are met.

### ⚠️ Issues found

- **High — Issue requirement not enabled by default**  
  `[net2/config.json:1171]`  
  `gcm_downgrade_protection` defaults to `false`, so deployed systems continue accepting unauthenticated CBC after successful GCM migration. This leaves the linked issue’s core authenticated-encryption and downgrade-prevention requirement only partially met while the PR declares `Fixes firewalla/firecommit#9400`. Either enable enforcement by default, provide a concrete rollout that enables it before closing the issue, or avoid closing the issue until enforcement is active.

- **High — Redis read failures silently disable downgrade protection**  
  `[encipher/lib/encipherio.js:916]`  
  `isGcmMigrated()` converts every Redis error into “not migrated.” After a restart or on an uncached group, a temporary Redis failure therefore permits CBC even when enforcement is enabled. This contradicts the fail-closed handling implemented for migration writes. Propagate the read error and reject CBC/legacy requests while enforcement is enabled; monitor mode may continue logging and accepting if compatibility requires it.

### 💡 Suggestions

- Add a regression test where `hgetAsync()` fails with enforcement enabled and verify that CBC/legacy requests fail closed.
- Add an integration test for actual SSE framing and client decoding, since current tests primarily exercise `decryptRequest()` rather than the full streaming route.
- Document the operational rollout and rollback criteria for enabling `gcm_downgrade_protection`.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9585  
Head SHA: `d547a08a4d0530ee50bc73616160d4dba997a6e6`  
Checked at: 2026-09-07 12:17:14 CST

PR #9587, opened 2026-09-07
sync purple 1.984 staging

jasonlyc:


MelvinTo (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- DMZ allow rules are synchronized during startup and mode transitions.
- `dmzModeSuspended` distinguishes rules disabled by mode handling from rules already disabled beforehand.
- Policy persistence occurs before enforcement is requested.
- No linked issue found.

### ⚠️ Issues found

- **Medium — User-disabled rules can be unintentionally re-enabled**  
  `[extension/portforward/portforward.js:481]`  
  After this code marks a rule with `dmzModeSuspended`, normal policy updates do not clear that field. If the user explicitly disables the rule while the box remains outside router mode, returning to router mode still interprets it as mode-suspended and enables it. Clear the marker whenever an explicit user action changes the rule, or store suspension provenance separately so user intent takes precedence.

- **Medium — Rules created while already outside router mode remain active**  
  `[extension/portforward/portforward.js:429]`  
  Synchronization only runs at startup and on `Mode:Applied`. A DMZ allow rule created or enabled afterward while the box remains in bridge/non-router mode can be enforced until another restart or mode transition. The router-mode restriction should also be applied in the policy creation/update path, or DMZ policy changes should trigger `_syncDMZAllowRules()`.

### 💡 Suggestions

- Add tests covering:
  - Rules created while already in non-router mode.
  - Rules disabled by the user while mode-suspended.
  - Rules that were disabled before leaving router mode.
  - Repeated router/non-router transitions.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9587  
Checked at: 2026-09-07 15:03:46 CST

PR #9588, opened 2026-09-07
sync purple se 1.984 staging

jasonlyc:


MelvinTo (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- DMZ allow rules are synchronized during both initialization and mode changes.
- The `dmzModeSuspended` marker avoids restoring rules that were already disabled before leaving router mode.
- Unchanged policies are skipped, avoiding unnecessary persistence and reenforcement.
- Startup synchronization runs under the existing shared lock.
- No linked issue found.

### ⚠️ Issues found

- **Medium — `extension/portforward/portforward.js:478`: stale ownership marker can later re-enable a manually disabled rule.**  
  When a marked rule becomes enabled through another policy update, the router-mode branch skips it without clearing `dmzModeSuspended`. If the user subsequently disables that rule while still in router mode, a later startup or mode synchronization sees both `disabled === "1"` and the stale marker, then enables it unexpectedly. Clear the marker whenever router mode observes a marked-but-enabled rule, or use state-transition logic that cannot leave ownership metadata behind.

### 💡 Suggestions

- Add regression tests for:
  - preserving rules disabled before entering non-router mode;
  - repeated mode transitions;
  - external enable/disable changes while `dmzModeSuspended` is set;
  - restart while already in router or non-router mode;
  - partial failures while synchronizing multiple policies.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9588  
Checked at: 2026-09-07 14:59:44 CST

PR #9589, opened 2026-09-07
sync gold se 1.984 staging

jasonlyc:


MelvinTo (approved):

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Synchronizes DMZ allow rules during startup and mode transitions.
- Runs startup and mode-change synchronization under `LOCK_SHARED`.
- Avoids restoring rules that were disabled before entering non-router mode.
- Skips unnecessary persistence and enforcement when state already matches.
- No linked issue found.

### ⚠️ Issues found

- **High — `extension/portforward/portforward.js:480`:** `dmzModeSuspended` does not reliably preserve ownership once another actor modifies the rule. For example, after this code suspends a rule, `disablePolicy()` immediately returns because it is already disabled and therefore cannot clear the marker. Returning to router mode will then re-enable the rule despite the newer disable intent. Similarly, explicitly enabling the rule in non-router mode leaves the marker behind and temporarily activates a rule that should remain suspended. The marker must be integrated into all enable/disable paths, or suspension state should be stored separately from the user-controlled `disabled` field.

### 💡 Suggestions

  - Rules disabled before entering non-router mode.
  - Disable and enable requests while a rule is mode-suspended.
  - Repeated router/non-router transitions.
  - Partial policy-update or enforcement failures.
  - Startup synchronization racing with policy changes.

- Consider isolating each rule’s failure so one malformed or failed policy does not prevent synchronization of all remaining DMZ rules.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9589  
Checked at: 2026-09-07 14:57:56 CST

PR #9593, opened 2026-09-08
Prevent Command Injection In Probe Targets

rtrappman-dev:
## Summary

Prevent command injection in network-monitor ping, DNS, and HTTP probes by replacing shell command construction with direct `execFile` calls.

## Root cause

Network-monitor targets and related policy values were interpolated into shell command strings used to invoke `ping`, `dig`, and `curl`.

Because these values can come from network-monitor policy configuration, shell metacharacters in a target or DNS lookup name could be interpreted as additional shell commands. The ping path was particularly sensitive because the generated command invoked `sudo ping`.

## Changes

- Add a dedicated network-monitor command helper that:
  - invokes `sudo ping`, `dig`, and `curl` with `execFile`;
  - passes policy-controlled values as discrete arguments rather than shell syntax;
  - parses ping latency and DNS query-time output directly in JavaScript;
  - discards HTTP response bodies while retaining curl timing output;
  - uses curl's `--` delimiter so URL values cannot be interpreted as command-line options.
- Update `NetworkMonitorSensor` to use the safe helper while preserving its existing sampling and error-handling behavior.
- Add regression tests with shell-metacharacter payloads to verify that targets and lookup names remain literal process arguments.

## Security impact

This closes a command-injection path at the subprocess boundary. Policy-controlled values are no longer evaluated by a shell, so characters such as `;`, quotes, and `#` cannot create additional commands or pipelines.

No authentication or authorization behavior is changed by this PR.

## Compatibility

The existing network-monitor interfaces and stored metric formats remain unchanged. Ping, DNS, and HTTP timing values are still returned to the sensor as numeric samples.

HTTP probes now explicitly discard the response body with `curl -o /dev/null`; only the timing value is needed by the network-monitor feature.

## Validation

- `npx mocha --exit test/test_network_monitor_security.js`
  - 2 passing
- `node -c sensor/NetworkMonitorSensor.js`
- `node -c util/NetworkMonitorCommand.js`
- `git diff --check`

ESLint was not run because an ESLint executable was not available in the installed repository dependencies.

### DIFF
--- a/sensor/NetworkMonitorSensor.js
+++ b/sensor/NetworkMonitorSensor.js
@@ -18,7 +18,7 @@ const log = require('../net2/logger.js')(__filename);
 
 const Sensor = require('./Sensor.js').Sensor;
 
-const exec = require('child-process-promise').exec;
+const networkMonitorCommand = require('../util/NetworkMonitorCommand.js');
 const f = require('../net2/Firewalla.js');
 const fc = require('../net2/config.js');
 const extensionManager = require('./ExtensionManager.js')
@@ -334,11 +334,10 @@ class NetworkMonitorSensor extends Sensor {
 
         rtid = intf.rtid;
       }
-      const result = await exec(`sudo ping -i ${cfg.sampleTick} ${rtid ? `-m ${rtid}` : ""} -c ${cfg.sampleCount} -W 1 -4 -n ${target}| awk '/time=/ && !/DUP!/ {print $7}' | cut -d= -f2`).catch((err) => {
+      const data = await networkMonitorCommand.ping(target, cfg.sampleTick, cfg.sampleCount, rtid).catch((err) => {
         log.error(`ping failed on ${target}:`,err.message);
-        return null;
+        return [];
       } );
-      const data = (result && result.stdout) ?  result.stdout.trim().split(/\n/).map(e => parseFloat(e)) : [];
       return { "status": "OK", "data": await this.recordSampleDataInRedis(MONITOR_PING, target, timeSlot, data, cfg, opts)};
     } catch (err) {
       log.error("failed to sample PING:",err.message);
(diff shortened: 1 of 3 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Replaces shell-interpreted commands with `execFile` and discrete arguments, closing the reported command-injection paths.
- Uses curl’s `--` delimiter to prevent URL option injection.
- Preserves numeric sample formats and existing failure behavior.
- Parses command output directly in JavaScript with appropriate filtering.
- Regression tests verify hostile values remain single literal arguments.
- Reviewed all 3 changed files.
- No linked issue found.

### ⚠️ Issues found

No blocking correctness, security, concurrency, or resource-management issues found.

### 💡 Suggestions

[util/NetworkMonitorCommand.js:37]  
`Number('')` evaluates to `0`, so unexpectedly empty curl output would be recorded as a successful zero-second request. Return `null` when trimmed output is empty or non-finite, and add a test for malformed/empty output.

[util/NetworkMonitorCommand.js:15]  
Consider inserting `--` before the ping target if the deployed `ping` implementation supports it. This prevents a target beginning with `-` from being interpreted as an option. This is argument injection rather than shell command injection, but boundary hardening would be worthwhile.

[test/test_network_monitor_security.js:31]  
Add edge-case coverage for ping output using `time<1 ms`, duplicate responses, malformed output, and empty curl output. Those parsing paths replaced the prior shell pipeline and deserve direct regression coverage.

### Verdict

APPROVE

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9593  
Head SHA: `cbda0e2c138723e211f8fcb9374567e08d02efc0`  
Checked at: 2026-09-08 10:31:56 CST

PR #9596, opened 2026-09-08
add various input checks

drahnieR:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Centralizes validation for policy fields, domains, hashes, routing-table names, ipset names, and VPN profile IDs.
- Replaces several shell-interpolated operations with `execFile`.
- Adds substantial regression coverage for control-character and command-injection inputs.
- No linked issue found.

### ⚠️ Issues found

- **High — `controllers/netbot.js:3484`**  
  `deleteVpnProfile` removed the local `type` declaration, but the running-profile error still references `${type}`. If the profile is active, this path throws `ReferenceError: type is not defined` instead of the intended HTTP 400 response. Preserve the normalized type in a local variable or obtain it from the resolved client.

- **High — `extension/routing/routing.js:65`**  
  Exceptions inside the async lock callback still do not call `done()`. In particular, a rejected `execFile()` leaves the outer promise pending until the lock timeout, despite the new comment claiming every failure is routed through `done(err)`. Wrap the entire callback in `try/catch` and call `done(err)` from the catch path.

- **High — `control/Block.js:1862`**  
  Invalid `tlsHost` input merely drops the TLS match while continuing to emit the rule. That can turn a host-specific block/allow/delete operation into a much broader rule. Reject or skip the complete rule instead of failing open.

- **Medium — `extension/routing/routing.js:76`**  
  Failure to read `/etc/iproute2/rt_tables` is logged but processing continues with an empty table. The allocator can consequently select an already-used ID and corrupt routing-table mappings. Treat the read failure as fatal and complete the lock with an error.

- **Medium — `test/test_netbot.js:1137`**  
  The new tests do not match `getVPNClient()` behavior: invalid IDs now produce `"invalid profileId"`, while tests expect the previous detailed message; unsupported types are converted to a thrown `{code, msg}` object, while `passesCheck()` expects an `Error` containing `"Unrecognized VPN client type"`. These tests should fail before validating the intended behavior.

### 💡 Suggestions

- Add explicit tests for `execFile()` and `readFile()` rejection paths in routing-table creation.
- Add a regression test for deleting a currently running VPN profile.
- Make the domain validators enforce the intended semantic format, not merely a shell-safe character set, if downstream consumers require valid DNS labels.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9596  
Head SHA: `c1d2cf78d00b4983de129fded334b619a32792b2`  
Checked at: 2026-09-08 17:46:41 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Centralizes validation for policy fields, domains, hashes, VPN profile IDs, routing-table names, and ipset names.
- Adds defense-in-depth at both API ingress and downstream command/config sinks.
- Includes substantial negative testing for control-character and command-injection payloads.
- Replaces several shell-interpolated operations with `execFile`.
- Reviewed all 15 changed files.

### ⚠️ Issues found

**High — Invalid TLS hosts produce dangerously broadened firewall rules**

[control/Block.js:1859]

When `tlsHost` is invalid, the code logs and omits only the `--tls-host` matcher, but continues emitting the rule. This can transform a host-specific block/allow/delete into a rule matching all traffic covered by its remaining predicates. Reject the entire operation or return before the rule is applied. A security validation failure must not widen a privileged firewall rule.

**High — Routing-table removal treats the validated name as a regular expression**

[extension/routing/routing.js:49]

`tableName` is interpolated into a `sed` regular expression, while `isValidTableName()` permits `.`. A request to remove `foo.bar` therefore also matches `fooXbar`, potentially deleting an unrelated routing-table entry. Escape regex metacharacters before constructing the expression, or parse and rewrite the file without regex interpolation.

**Medium — Deleting an active VPN profile references an undefined variable**

[controllers/netbot.js:3484]

The refactor removes the local `type` declaration, but the active-profile error still interpolates `${type}`. If `vpnClient.status()` is truthy, this path throws `ReferenceError` instead of the intended structured 400 response. Retain the resolved type or use `value.type || "openvpn"`.

**Medium — New netbot validation tests do not match the refactored error contract**

[test/test_netbot.js:1138]

The tests expect profile validation messages beginning with `'profileId' should only contain`, but `getVPNClient()` now returns `invalid profileId`. They also expect an `Error` containing `Unrecognized VPN client type`, while the helper catches that error and throws a plain `{code: 400, msg: "Unsupported VPN client type..."}` object. These cases will fail before validating the intended behavior. Update the tests or preserve the established handler error contract.

### 💡 Suggestions

- Add regression tests proving an invalid TLS host emits no firewall operation.
- Test removal of `foo.bar` while `fooXbar` exists to prevent regex-collision regressions.
- Add an active-profile deletion test covering the structured 400 response.
- Wrap the routing-table lock callback in `try/catch` and explicitly call `done(err)`, ensuring rejected `readFile`/`execFile` operations cannot leave the outer promise pending.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9596  
Head SHA: `eec7ff952536968338bb07f12c318a77224ab5e1`  
Checked at: 2026-09-08 18:02:12 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Centralizes VPN client input validation and preserves appropriate 400/404 responses.
- Adds defense-in-depth checks at multiple command/configuration boundaries.
- Uses `execFile` for routing operations where shell interpretation is unnecessary.
- Includes broad regression coverage for malformed policy, domain, category, VPN, and routing inputs.
- No linked issue found.

### ⚠️ Issues found

[extension/routing/routing.js:49]  
**High — Routing table names are embedded in a `sed` regular expression without escaping.** `isValidTableName()` permits `.`, which `sed` interprets as a wildcard. Removing a valid name such as `foo.bar` can therefore delete a different entry such as `fooXbar`. Escape BRE metacharacters before constructing the expression, or parse and rewrite `rt_tables` without regex interpolation.

[util/util.js:235]  
**Medium — The `notes` exemption applies to every nested property named `notes`.** The JSON replacer checks only the current key, so a structure such as `{ dnsmasq: { notes: "...\n..." } }` bypasses control-character validation even though only the top-level policy note is documented as safe multiline text. Track traversal depth/path and allow CR/LF exclusively for the root policy’s `notes` field.

### 💡 Suggestions

- Add regression tests for routing names containing literal dots and for nested `notes` values containing newlines.
- Consider bounding routing-table name length to the actual downstream kernel/iproute2 constraint.
- Add a concurrent or multi-process routing-table allocation test; the JavaScript lock does not coordinate separate processes during the read/select phase.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9596  
Head SHA: `08f56650afb6aea73e561622806808149340142a`  
Checked at: 2026-09-08 19:02:31 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Replaces several shell-interpolated operations with argument-based `execFile` calls.
- Adds centralized validation for VPN profile IDs, domains, hashes, ipset names, routing-table names, and policy control characters.
- Preserves host-specific TLS matching by dropping unsafe rules entirely instead of accidentally creating broader rules.
- Adds substantial regression coverage for command injection, control characters, routing-table lifecycle, and API validation.
- CI build passes.
- No linked issue found.

### ⚠️ Issues found

- **Medium — `control/CategoryUpdater.js:768`**  
  `addCategoryData()` dereferences `obj.type` without verifying that each cloud/MSP-provided entry is a non-null object. A malformed entry such as `null` causes a `TypeError`, aborting the entire category update instead of being filtered. Since this validation is explicitly placed at an untrusted ingestion boundary, malformed values should be dropped safely:

  ```text
  [control/CategoryUpdater.js:768]
  Please validate that `obj` is a non-null object before reading `obj.type` or `obj.id`. Consider making `filterValid()` catch validator failures as well, so one malformed cloud entry cannot abort the complete category refresh.
  ```

- **Medium — `controllers/netbot.js:3273`**  
  `updateIncludedElements` validates contents only when `elements` is an array. A non-array value bypasses validation and is silently accepted by `CategoryUpdater.updateIncludedElements()`, which returns without updating anything. This produces a successful API response for an invalid request and conflicts with the stated boundary-validation behavior.

  ```text
  [controllers/netbot.js:3273]
  Reject non-array `elements` explicitly. The condition should require `_.isArray(elements)` before checking members; otherwise malformed payloads succeed as no-ops and callers cannot distinguish invalid input from a completed update.
  ```

- **Medium — `net2/Ipset.js:40`**  
  `REGEX_SETNAME` permits names beginning with `-`, while `read()` still interpolates the name into a shell command. Such a value can be interpreted as an `ipset list` option rather than a set name. This is argument injection even though shell metacharacters are blocked.

  ```text
  [net2/Ipset.js:40]
  Require the first character to be alphanumeric/underscore, or migrate this invocation to `execFile` and use an end-of-options delimiter if `ipset` supports it. Add regression coverage for values such as `-t` and `-output`.
  ```

### 💡 Suggestions

- Add tests covering `null`, primitives, and malformed objects passed to `addCategoryData()`.
- Add a negative API test for non-array `updateIncludedElements` payloads.
- Consider imposing a practical routing-table-name length limit; the current validator intentionally accepts arbitrarily long names, which downstream tooling may reject.
- Fix the `safty` typo in the `PolicyManager` comment.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9596  
Head SHA: `583fb590f475da8afebf65fa85de6cff8fd68211`  
Checked at: 2026-09-09 09:33:39 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Centralized validation reduces duplicated security checks.
- `execFile` and strict allowlists substantially reduce shell-injection exposure.
- Validation covers nested policy values and preserves retry behavior.
- Tests cover command injection, control characters, routing-table lifecycle, and API validation.
- No linked issue found.

### ⚠️ Issues found

[control/CategoryUpdater.js:841]  
**Medium:** Regex entries containing control characters still pass validation. `CategoryEntry.isValidDomainRE()` accepts strings containing CR/LF because they are valid JavaScript regexes. These values later become line-oriented dnsmasq `re-match` directives. Depending on `writeConfig`, this can either inject directives or reject the entire category mapping file. Explicitly reject control characters before persisting regex entries and add newline/carriage-return tests.

[extension/routing/routing.js:80]  
**Medium:** Routing-table ID selection occurs before acquiring the cross-process `flock`. The process-local `AsyncLock` does not coordinate multiple Firewalla processes, so two processes can read the same available ID and subsequently append different table names using that ID. Move the read, allocation, and update into one `flock`-protected operation.

### 💡 Suggestions

- Add integration coverage for concurrent routing-table creation from separate processes.
- Add negative regex tests for CR, LF, NUL, DEL, and catastrophic-backtracking expressions.
- Fix the trailing whitespace in `controllers/netbot.js:1816`.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9596  
Head SHA: `ebb273819e4614f5503dcda0f57415a432dfb37b`  
Checked at: 2026-09-09 13:48:49 CST

j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Centralizes validation for policy fields, domains, hashes, VPN profile IDs, routing-table names, and ipset names.
- Uses `execFile` and strict allowlists to reduce command-injection exposure.
- Invalid TLS hosts now cause the entire rule operation to be skipped, avoiding accidentally broadened firewall rules.
- Validation correctly handles nested policy data and preserves retry behavior for rejected policies.
- Adds extensive regression coverage for control characters, malformed API input, and routing-table operations.
- No linked issue found.

### ⚠️ Issues found

**High — Routing-table ID allocation is not protected across processes**

[extension/routing/routing.js:65]

The cross-process `flock` protects only the final file update. Reading `rt_tables` and selecting an unused ID happen beforehand under a process-local `AsyncLock`.

Two Firewalla processes can therefore read the same state, select the same free ID, and then sequentially append different names with that ID. The resulting duplicate mapping can send routes to the wrong table.

Move the complete read/select/write transaction under the same `flock`. A small helper executed while holding the lock, or atomic file parsing and rewriting inside the locked process, would ensure IDs are allocated from the latest state.

### 💡 Suggestions

- Add an integration test that launches routing-table creation from two separate Node processes and verifies unique IDs.
- Add direct tests for malformed non-domain objects passed to `addCategoryData()`, including `null`, missing IDs, and invalid address entries.
- Consider imposing a practical maximum length on routing-table names so accepted input matches downstream iproute2 constraints.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9596  
Head SHA: `52ea3dd5a5cbd6201851bd8e6b30e9a4786d05b0`  
Checked at: 2026-09-09 17:47:03 CST

j-sallyjin:
## PR Review Summary

Reviewed all 19 files in the supplied diff. Static review only; appliance tests were not run.

### ✅ What looks good

- Centralizes validation across VPN handlers and category ingestion.
- Uses `execFile` for routing operations and correctly escapes dots in the deletion pattern.
- Ensures routing failures release the process-local lock.
- Preserves enforcement order and avoids caching rejected policy values as applied.

### ⚠️ Issues found

**Medium — Validation exceptions can abort all policy enforcement**  
[net2/PolicyManager.js:403]

`hasControlChar()` uses `JSON.stringify()`, which can throw for excessively nested data, circular references, or a throwing `toJSON()` method. This new validation loop runs outside the existing per-policy exception handler, so one such value rejects `execute()` before any clean policies are applied. Catch validation errors per key and mark that key invalid. JavaScript invokes `toJSON()` before the replacer can reject it. [ECMAScript specification](https://tc39.es/ecma262/multipage/structured-data.html#sec-serializejsonproperty)

**Medium — Delete validation prevents cleanup of legacy entries**  
[controllers/netbot.js:3209]

Both removal handlers now reject domains that fail the new insertion rules. Earlier versions could persist those values, and this PR does not migrate existing include/exclude sets. After upgrading, users therefore cannot remove affected entries through these APIs. Permit exact-member deletion through Redis with appropriate type validation, or provide migration and cleanup before enforcing the stricter deletion checks. The same issue applies to `removeExcludeDomain`.

**Medium — Routing tests confuse name resolution with kernel table existence**  
[test/routing.js:39]

`createCustomizedRoutingTable()` registers a name in `rt_tables`; it does not create a kernel routing table. On modern Linux/iproute2, querying a newly registered table without routes can return a nonzero status even though its name resolved correctly. Consequently, several lifecycle assertions can fail on a clean appliance. Verify the file mapping separately, then establish a temporary route before asserting successful kernel queries. [Linux routing implementation](https://github.com/torvalds/linux/blob/master/net/ipv4/fib_frontend.c)

### 💡 Suggestions

- Add regression coverage for validation exceptions alongside clean policies, and deletion of pre-upgrade include/exclude entries.
- Run the routing lifecycle tests on supported appliance versions after correcting their table-existence assumptions.
- Track cross-process routing allocation separately: the race predates this diff.

**Issue requirement coverage:** No linked issue found on the accessible [PR conversation](https://github.com/firewalla/firewalla/pull/9596). Minimized comments could not be inspected.

### Verdict

**REQUEST_CHANGES** — Address the enforcement error isolation, legacy cleanup regression, and routing test assumptions before merging.

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9596  
Head SHA: `215763b175e39fece98964da4e68d3f796b8a54d`  
Checked at: 2026-09-14 12:43:08 CST

jasonlyc (approved):

PR #9603, opened 2026-09-10
send mDNS queries from port 5353 to stop draining avahi reflection slots

jasonlyc:
BonjourSensor bound only the first of its per-interface bonjour instances to 0.0.0.0:5353; the rest used bind:false, so their queries left from ephemeral ports. avahi treats those as legacy unicast and holds one of its 100 reflection slots for each, which they exhausted.

Use one socket bound to 0.0.0.0:5353 and fan queries out by moving IP_MULTICAST_IF before each send, serialized because dgram.send() only reaches the kernel a tick later.

Verified on a box with 5 bridges: legacy unicast from the box went to zero, queries spread evenly across bridges, discovery unaffected.

### DIFF
--- a/sensor/BonjourSensor.js
+++ b/sensor/BonjourSensor.js
@@ -15,6 +15,7 @@
 'use strict';
 
 const net = require('net')
+const dgram = require('dgram')
 
 const log = require('../net2/logger.js')(__filename);
 
@@ -35,6 +36,8 @@ const hostManager = new HostManager();
 const _ = require('lodash')
 
 const lastProcessTimeMap = {};
+
+const MDNS_MULTICAST_ADDR = '224.0.0.251';
 
 // BonjourSensor is used to two purposes:
 // 1. Discover new device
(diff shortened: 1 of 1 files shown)


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Uses a single socket bound to UDP/5353, directly addressing BonjourSensor’s accidental legacy-unicast queries.
- Serializes per-interface sends so `IP_MULTICAST_IF` remains stable until each send completes.
- Deduplicates interface addresses and preserves VPN-interface exclusions.
- Handles membership and synchronous send failures without crashing.
- Verification confirms queries originate from port 5353 and discovery still works.

### ⚠️ Issues found

- **Medium — Linked issue only partially covered.** Issue `firecommit#9462` also identifies duplicate reflection across parent/VLAN interfaces and possible loops involving neighboring reflectors. This PR fixes BonjourSensor’s ephemeral source ports, but does not address those additional sources of reflector load. The issue should remain open unless post-change measurements confirm the slot-exhaustion warnings and dropped queries are eliminated.
- **Low — No automated coverage for the new queue/socket behavior.** The implementation introduces stateful asynchronous logic without tests for multi-interface fan-out, send failures, membership failures, and teardown with pending jobs.

### 💡 Suggestions

- Add focused tests verifying one query per unique interface, strict serialization, continued draining after errors, and no sends before the socket is listening.
- Record before/after rates for `No slot available for legacy unicast reflection` on XCrystal203. Source-port validation alone does not prove that reflection failures are resolved.
- Document the remaining parent/VLAN duplication and reflector-loop investigation in the linked issue.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9603  
Head SHA: `4d5e7e824309661ac5b4c55718245b363d001377`  
Checked at: 2026-09-10 12:04:04 CST

MelvinTo (approved):

PR #9607, opened 2026-09-10
sync purple 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- DNS rules now default to `address-ip-only`, preventing non-A/AAAA queries from leaking upstream.
- `ipOnly` is preserved during policy serialization and comparison.
- Docker 29’s missing `GraphDriver` metadata is handled with a Docker-root fallback.
- MSP trace summarization preserves the full payload in action history.
- Tests cover the primary DNS, Docker, and trace-log compatibility paths.
- No linked issue found in the PR title, body, or comments.

### ⚠️ Issues found

- **Medium — `lib/ControllerBot.js:235`: Changed fields can be reported as an empty list.**

  `_.isEqual(next, prev)` correctly detects all value changes, but `difference()` intentionally treats `false`, `null`, `undefined`, and `""` as equivalent defaults. Consequently, transitions such as `null → false` or `"" → null` produce a trace entry with `changed: []`, despite the payload being different.

  This contradicts the method’s stated purpose of recording which fields changed and makes the resulting audit entry misleading. Derive the top-level changed keys using strict deep equality—for example, compare the union of keys with `_.isEqual(next[key], prev[key])`—or enhance `difference()` with strict semantics for this caller. Add a regression test covering default-like value transitions.

### 💡 Suggestions

- Add focused tests for `Policy.ipOnly` parsing and equality behavior, including absent, `"true"`, `"false"`, and malformed persisted values.
- Add tests confirming `address-ip-only` generation and the expanded literal-address search regex across `address-high`, `address-uhigh`, and `address-ip-only`.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9607  
Head SHA: `cf33e92cdd6d273d0f4458c9418234c1de96a2eb`  
Checked at: 2026-09-10 14:06:04 CST

MelvinTo (approved):

PR #9629, opened 2026-09-14
Rename the IDS flow-engine feature to pcap_suricata_fleet

MelvinTo:
Follow-up to #9606. The feature that makes fleet replace **suricata** was named `pcap_zeek_suricata`, which reads as a zeek feature. It becomes `pcap_suricata_fleet`, matching `pcap_zeek_fleet`:

| feature | on |
|---|---|
| `pcap_zeek_fleet` | fleet runs as `brofish.service` and writes the zeek logs |
| `pcap_suricata_fleet` | fleet evaluates the suricata rule set and writes `eve.json` |

Renamed in `net2/Constants.js`, the `userFeatures` defaults in `net2/config.json`, the PSE platform default, `platform/platform.sh`'s accessor, the effective-features file `FleetEnginePlugin` publishes for the shell, and the docs and tests.

No migration: #9606 merged yesterday, both features default to false everywhere except PSE (not shipping yet), so no box has the old key set except two test boxes, which I have cleared by hand.

PR #9637, opened 2026-09-14
One fleet process when it owns both roles

MelvinTo:
Builds on #9629. With both features on, a box ran **two** fleet processes: one as `brofish.service` for the zeek logs and one under `suricata.service` for the IDS. This folds them into one.

**Why they were split.** `--zeekctl-compat` installs zeek's `restrict_filters` as the capture BPF, and those deliberately drop and sample traffic (the TLS sampling filter, NTP, port 8080, the box's own host). Sharing that socket with the rule engine silently cost IDS coverage.

**What changed.** fleet 0.1.34 keeps the two views apart inside one process:

- the capture socket takes suricata's own `bpf-filter`;
- zeek's filters are compiled once (`pcap_open_dead` + `pcap_compile`) and evaluated per packet (`pcap_offline_filter`);
- only packets zeek would have captured reach the zeek pipeline (tracking, byte accounting, parsers, signatures, notices, logs), so conn.log stays byte-for-byte what a filtered zeek produced;
- every packet reaches the rule engine, and a flow zeek never saw is tracked for the IDS but writes no conn.log record.

So when fleet owns both roles and the box wants both, `brofish` runs without `--no-suricata` and the suricata unit is held off again. The split arrangement stays for the mixed case (zeek for flows, fleet for the IDS).

**Safety**

- The fold happens only if the installed binary says it can: `fleet --capabilities` must list `shared-roles` (0.1.34+). An older asset, including one an upgrade has not downloaded yet, keeps the two services apart.
- Folding stops the IDS-only fleet first, so two processes cannot both write `eve.json`.
- fleet decides shared mode from the role, not from the rules having loaded, so the `suricata.yaml` Firewalla writes after startup gets full coverage immediately.
- The watchdog and the suricata cron template follow: one role to check, `crontab.fleet` covers the IDS.

**Tested** on a Gold Pro: 101 sandbox checks pass, including a fleet without the capability keeping the two services apart and a capable one folding them. Live with both features on: one process, `brofish` active, `suricata` inactive, API reporting `suricata=true` with 7,213 rules loaded, watchdog green. fleet-side unit tests drive two synthetic flows through a worker holding a zeek filter: both are tracked, only the zeek-visible one reaches conn.log.

fleet releases: https://github.com/firewalla/fleet/releases/tag/v0.1.34


j-sallyjin:
## PR Review Summary

Reviewed all five supplied file diffs. An isolated shell check confirmed the watchdog issue below; full repository tests were unavailable.

### ✅ What looks good

- Service installation checks binary capabilities before enabling shared roles.
- The transition explicitly stops the separate IDS service to prevent duplicate processing.
- Added tests cover installation with both capable and older fleet binaries.

### ⚠️ Issues found

**[High] The older-binary fallback loses IDS watchdog coverage**

Locations: `scripts/fleet-ping.sh:46` and `net2/SuricataControl.js:70–72`.

When both engines are fleet and both roles are enabled, a binary without `shared-roles` correctly gets two separate services. However, the watchdog and cron selection assume a shared process without checking that capability.

Consequently, `fleet-ping.sh all` probes only brofish, `fleet-ping.sh suricata` exits without probing anything, and SuricataControl omits `crontab.fleet-ids`. An unresponsive separate IDS process can therefore remain undetected by the fleet watchdog.

**Fix:** Make service installation, cron selection, and watchdog routing consume the same applied shared-role state. Preserve the IDS cron entry and port `8928` probe whenever the services remain separate.

### 💡 Suggestions

- Extend the older-binary regression test to verify cron selection and watchdog targets; inspecting drop-ins alone misses this regression.
- Resolve shared-role capability once per apply so separate probes cannot produce inconsistent service configurations.

**Issue requirement coverage:** No linked issue found in the supplied diff. GitHub PR metadata and comments could not be retrieved, so linked-issue requirements remain unverified.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9637  
Head SHA: `22e3f88753677568cd7624fe9ff787cd68780e1e`  
Checked at: 2026-09-14 12:32:40 CST

j-sallyjin:
## PR Review Summary

Reviewed all five files in the supplied diff. Found two issues requiring changes.

### ✅ What looks good

- Installation checks binary capabilities before enabling shared roles.
- The transition explicitly stops the separate IDS service and checks for stop failures.
- Tests cover drop-in selection for older and capable fleet binaries.

### ⚠️ Issues found

1. **High — Older fleet binaries lose IDS watchdog coverage**  
   **[scripts/fleet-ping.sh:46]**, **[net2/SuricataControl.js:69–71]**

   With both roles enabled and both engines set to fleet, a binary without `shared-roles` correctly retains the separate IDS service. However, these two paths assume sharing without checking capability: `SuricataControl` removes the fleet IDS cron entry, and `fleet-ping.sh` skips the IDS even when explicitly invoked for `suricata`.

   A stalled IDS consequently loses its fleet health check and recovery path. Make both decisions depend on the successfully applied shared/split mode.

2. **Medium — Repeated capability probes can produce conflicting service configurations**  
   **[scripts/fleet-engine.sh:93–95]**

   `shared_roles()` executes a fresh capability probe for each rendering and lifecycle decision. If the first probe fails or times out and the second succeeds, the script renders brofish with `--no-suricata` while installing the drop-in that disables suricata. Neither service then evaluates IDS rules. The reverse sequence configures two IDS processes.

   Resolve shared mode once per apply and reuse that result throughout rendering, stopping, verification, and restarting. An isolated shell reproduction confirmed the conflicting rendering when probe results change.

### 💡 Suggestions

- Extend the older-binary test to verify cron selection and actual IDS watchdog execution.
- Add a capability-failure test asserting that one apply always produces a consistent service arrangement.
- Validate rule updates and restarts in shared mode, including propagation to the process under `brofish.service`.

**Issue requirement coverage:** No linked issue found in the supplied material. GitHub retrieval failed, so PR body/comments and linked requirements could not be verified.

**Validation:** Reviewed the supplied diff and ran isolated decision-logic checks. The full PR test suite was not run.

### Verdict

**REQUEST_CHANGES** — Fix watchdog coverage and ensure consistent mode selection before merging.

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9637  
Head SHA: `7942b225ac71593a0b2330cbbb3148373aff8243`  
Checked at: 2026-09-15 11:38:28 CST

jasonlyc (approved):

PR #9642, opened 2026-09-14
event summary: allow record keys on top-level event fields

jasonlyc:
Refs firewalla/firecommit#9550

## What changed

A setting gains a sibling key list, `eventKeys`, resolved against the **top-level event object**,
alongside the existing `labelKeys`, which stay resolved against **`event.labels`**. A setting can
now group by `state_key` (or `state_type`, `state_value`, ...) — which was previously impossible,
because record identity came exclusively from labels. The first real use case is a state type whose
labels never carry the thing you want to group on: `wan_state` puts the interface name in
`state_key` and only link status in the labels.

Two commits, and the split matters for review:

- `2797b0c5a` — pure refactor, no behaviour change. The `(key list, source object, nil→null)`
  resolution was written twice: once to build the joined record key, once to build the record's
  visible fields. Two loops in two functions that had to agree, or a stored record would
  misrepresent its own group. Both now derive from one `_resolveRecordFields()` returning an
  ordered `[name, value]` list.
- `cb2e4d59a` — the feature: one `.map()` over `eventKeys` inside that helper, the validator
  changes, the comment rewrite, and the tests.

The refactor lands first on purpose. It is what makes the back-compat claim below checkable by
reading one function rather than by reasoning about two, and it stops `eventKeys` from turning two
parallel loops into four.

## Design decisions

**Two sibling arrays, not a prefixed `recordKeys` list.** A qualified syntax (`"labels.mac"` vs
`"state_key"`) would invent path-parsing vocabulary that exists nowhere else here — `matchFilter`
is a flat `obj[field] === value` — and would need an escaping story for a label whose name contains
a dot. It also collides with the neighbour: `recordKeys` already means *top-level event fields* in
`BlockStatsSensor`, so reusing the word for a mixed list would make one config token mean two
things in two adjacent files. Two arrays need no parser and keep each key honest about its source
object by construction.

**The event is not flattened.** A label literally named `state_key` can never shadow the top-level
`state_key`, and vice versa. There is a test pinning exactly that.

**A name appearing in both lists makes the setting invalid**, and it is dropped with an error log.
The record is a flat map, so it can hold only one value per name; allowing the collision would
produce a record whose stored fields cannot reproduce its own `_k`. The alternative — qualifying
the field name in the stored record — changes the app-visible record shape and needs app-side
agreement we do not have.

**A setting is valid with `eventKeys` only.** The validator's rule is now "at least one of the two
lists is non-empty". Requiring a dummy `labelKeys` to unlock outer-level grouping would be a wart.

**A missing outer field is `null`**, exactly as a missing label already is, and it groups
distinctly rather than collapsing. No warning log — `_onEvent` is a per-event hot path, and
`state_key` is a required field of every state event, so the case is a producer bug that the `null`
in the app payload makes visible anyway.

**An `eventKeys` value that resolves to a non-primitive is coerced to `null`** (e.g. pointing
`eventKeys` at `labels` itself). `JSON.stringify`'s key order for an object is insertion-dependent,
which would make `_k` an unstable record identity. `labelKeys` values are deliberately left as-is,
untouched, so existing settings' keys stay byte-identical.

**Caps are unchanged.** `MAX_RECORDS_PER_KEY` (400) already bounds the redis payload per setting
per bucket, and the cardinality multiplier is entirely under the control of whoever authors the
setting. Raising it speculatively, with no measured `state_key` cardinality for any shipped
setting, would loosen a bound that is currently doing its job.

## Stored data, and the operational rule that falls out of it

Buckets live in redis for `RETENTION_SECS` (7 days) and each record carries its `_k` inside it,
matched with `===` on every subsequent event. So the key encoding is frozen:

The record key tuple is `[...eventKeys values, ...labelKeys values]`. For a label-only setting
`eventKeys` is `[]`, so an empty prefix concatenated onto the same array yields the same
`JSON.stringify` output — existing cloud and local settings produce byte-identical keys and
records, and buckets already in redis keep matching. `net2/config.json` is deliberately **not**
touched: the shipped `event::dap_state_change` setting keeps its label-only composition.

The corollary, now stated in the file's header comment: **a setting that changes its key
composition must be given a new `key`, not be edited in place.** Adding `eventKeys` to a live
setting appends duplicate records against its existing buckets for up to 7 days. A new `key` lands
in a different bucket and sidesteps it. Worth flagging to whoever authors the next cloud push,
since nothing in the code can prevent it.

No reader change was needed: `HostManager.summarizeEventSummaryBucket` strips exactly
`_k`/`_firstTs`/`_lastTs` and forwards every other record field, so `state_key` reaches the app
additively.

## A note on tightened validation

Element validation is now stricter than before: a `labelKeys` entry that is not a non-empty string
drops the whole setting, where previously it was accepted and produced a record field literally
named `"null"`. Dropping is the better failure, but it is a behaviour change to a
currently-accepted input.

## Verification

Exercised on real hardware by publishing synthetic `EVENT_GENERATED` messages through the sensor's
own `sem` listener and reading the resulting redis buckets directly: two events differing only in
`state_key` produced two distinct records under an `eventKeys` setting, while a label-only sibling
setting fed by the same two events produced one record with `cnt: 2` and a key string identical to
the pre-change format. A `state_key`-less event grouped separately with `state_key: null`, and a
setting naming the same key in both lists was rejected with the expected log line while its sibling
kept accumulating.

Not covered: no real state-transition producer was driven, and the new unit tests in
`test/test_event_summary.js` were not run in this workspace — the suite needs a live redis and
mocha/chai, and master itself is not green there. They are left for CI.


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Centralized field resolution keeps grouping keys and visible fields consistent.
- Label-only settings preserve their existing serialized identity.
- Added tests cover mixed grouping, missing fields, and separation of top-level fields from labels.
- Reviewed both supplied file diffs.

### ⚠️ Issues found

**Medium — Reject grouping keys that collide with record metadata**  
[sensor/EventSummarySensor.js:269]

The new validation accepts `eventKeys: ['_k', 'state_key']`. For an ordinary event without `_k`, field resolution produces `null`, and this assignment overwrites the computed `record._k`. The persisted identity therefore differs from the grouping key, undermining subsequent record matching and potentially creating duplicate records.

Reject `_k` and other reserved summary fields during settings validation. Apply this protection to both key lists.

An isolated JavaScript probe reproduced the identity overwrite. The repository test suite was unavailable.

### 💡 Suggestions

- Add a regression test rejecting reserved field names, plus a repeated-event test verifying that persistence preserves one record per grouping key.
- Extend tests to cover object-valued top-level fields and invalid key-list entries, exercising the new normalization and validation branches.

**Issue requirement coverage:** No linked issue found in the supplied diff. GitHub PR details and comments could not be retrieved, so linked-issue discovery and requirement coverage remain unverified.

### Verdict

**REQUEST_CHANGES** — Protect the persisted record identity before merging.

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9642  
Head SHA: `cb2e4d59a078ab4c879d5273de81b0fb7c1cdc02`  
Checked at: 2026-09-14 17:03:21 CST

MelvinTo (approved):

PR #9648, opened 2026-09-14
sync gold se 1.984 staging

jasonlyc:


MelvinTo (approved):

j-sallyjin:
## PR Review Summary

Reviewed all 14 changed files in the supplied diff. Found two correctness issues requiring changes.

### ✅ What looks good

- WAN-aware alarm deduplication includes coverage for system-wide and per-WAN alarms.
- Domain-only recycling preserves resolution for static and port-specific entries.
- Event summaries use bucket-level locking and millisecond timestamps to handle concurrent and out-of-order updates.
- Tests cover grouping, DST boundaries, retention, and record limits.

### ⚠️ Issues found

**1. Medium — Initial `no_error` events are still discarded**

[event/EventRequestHandler.js:258]

The comment promises initial-state emission, but the branch still checks only `if (isError)`. Since `isStateEventError()` now always returns `false` for `no_error` events, their initial state is cached without being logged or summarized. Repeated observations remain suppressed until the value changes.

Emit the initial event when `isError || this.isNoErrorStateEvent(eventRequest)`, while keeping error-cache updates conditional on `isError`. The newly added initial-state assertion currently fails.

**2. Medium — A failed Redis read can erase an existing summary bucket**

[sensor/EventSummarySensor.js:247]

The catch handles both JSON parsing failures and Redis GET failures by returning `null`. If a transient read failure occurs and the subsequent write succeeds, `_upsertRecord()` creates an empty bucket and overwrites the existing data. This loses every previously accumulated record and count in that bucket.

Let Redis read failures abort or retry the update. Handle malformed JSON separately, and initialize a new bucket only after successfully establishing that the key is absent.

### 💡 Suggestions

- Add a regression test verifying that a failed GET leaves existing bucket data unchanged.
- Run the added `no_error` integration test after correcting initial-state emission.

Both findings were reproduced with isolated in-memory stubs. The full Redis-backed test suite was not run because dependencies were unavailable.

**Issue requirement coverage:** Unverified. The supplied diff contains no issue reference, and GitHub access failed, preventing inspection of PR metadata, comments, and linked issues.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9648  
Head SHA: `716af5393a62dc6f022a35861eeeb8e01d63d2c9`  
Checked at: 2026-09-14 18:54:12 CST

PR #9650, opened 2026-09-14
sync orange 1.984 staging

jasonlyc:


MelvinTo (approved):

j-sallyjin:
## PR Review Summary

Reviewed all 14 files in the supplied diff. Found two correctness issues.

### ✅ What looks good

- WAN-specific alarm deduplication has focused regression coverage.
- Summary updates use per-bucket locking and millisecond timestamps to handle concurrent, out-of-order events.
- The DNS optimization preserves resolution for static and port-specific entries.

### ⚠️ Issues found

1. **[Medium] Initial `no_error` events are never emitted** — [event/EventRequestHandler.js:258]

   `isStateEventError()` now returns false for `no_error` events, but the initial-state branch still sends only when `isError` is true. The first observation is cached without reaching the event log or summary sensor, contradicting the new behavior and test. Include `isNoErrorStateEvent(eventRequest)` in the initial-emission condition while keeping error-cache writes conditional on `isError`.

2. **[Medium] A failed Redis read can erase the entire summary bucket** — [sensor/EventSummarySensor.js:247]

   The catch converts both read failures and JSON parsing failures into `null`. If a transient GET failure is followed by a successful write, the code overwrites the existing bucket with a single-event payload, losing every previous record and count. Abort or retry on read errors; initialize an empty bucket only after a successful read confirms the key is absent.

### 💡 Suggestions

- Add a regression test proving that a rejected GET cannot overwrite existing summary data.
- Both findings were reproduced with isolated checks. The full test suite was not run because dependencies were unavailable.
- **Issue requirement coverage:** Unverified. GitHub access failed, so the target PR’s linked issues and acceptance criteria could not be checked.

### Verdict

**REQUEST_CHANGES** — fix initial-event emission and preserve existing summaries on Redis read failures.

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9650  
Head SHA: `716af5393a62dc6f022a35861eeeb8e01d63d2c9`  
Checked at: 2026-09-14 18:42:14 CST

PR #9664, opened 2026-09-15
sync purple 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

Reviewed all five supplied changed files. Found one confirmed test regression.

### ✅ What looks good

- Extends the existing serialized queue to `switch_*` events while preserving AP handling.
- Defines the phone-paired notification category centrally.
- Updates notification assertions to check the retained `message` field.

### ⚠️ Issues found

**Medium — New category breaks two existing tests**  
[controllers/netbot.js:419]

Adding `payload.category` changes the complete object returned by `getNotifEvent()`. The existing tests at `test/test_netbot.js:732` and `:740` compare that object using `.eql()`, but both expected objects omit `category`. Consequently, both the current-label and legacy-label tests fail.

**Fix:** Add `category: Constants.NOTIF_CATEGORY_PHONE_PAIRED` to both expected objects. Also assert that `_notifyNewEvent()` forwards the category.

An isolated before/after check reproduced both failures using the test file matching the supplied base blob. The full test suite was not run.

**Issue requirement coverage:** Unverified. Cached metadata for the original serialization PR references [proj-ap#758](https://github.com/firewalla/proj-ap/issues/758), but its requirements and the target PR’s metadata could not be retrieved.

### 💡 Suggestions

- **`test/test_event.js:112`:** Add a concurrency regression test that submits overlapping switch events for the same key, deliberately delays the first state update, and verifies processing order, emitted transitions, and final saved state. The added predicate assertions alone cannot detect a broken serialization path.
- Test the final notification payload to verify category forwarding and removal of `body_loc_msg`.

### Verdict

**REQUEST_CHANGES** — Update the two failing payload expectations before merging.

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9664  
Head SHA: `e9fb461f61009e2b3940bad7869c9de8b410ec1a`  
Checked at: 2026-09-15 21:20:12 CST

MelvinTo (approved):

PR #9681, opened 2026-09-17
fix bug, disable udp_tls desn't work in some platforms

caixingcheng:
1. some platforms enabled systemd-pstore service, so need check folder /var/lib/systemd/pstore for these platforms
2. goldv1/goldv2 pstore module was compiled as module and not loaded by default, need install pstore module for these platforms


j-sallyjin:
## PR Review Summary

### ✅ What looks good

Reviewed **1/1 changed files** and the linked issue discussion.

- Uses argument-based `execFile` calls.
- Checks backend registration after `modprobe`.
- Archives and clears the selected source under the existing Redis lock.

### ⚠️ Issues found

1. **Medium — Generated `dmesg.txt` corrupts crash timestamps**  
   [net2/KernelCrashMonitor.js:41](https://github.com/firewalla/firewalla/blob/474af42294188efab08ce35eca7a39b3ac125501/net2/KernelCrashMonitor.js#L41)

   The expanded glob includes `dmesg.txt`, which [systemd generates during archival](https://github.com/systemd/systemd/blob/v249/src/pstore/pstore.c). Its modification time reflects archival rather than the original crash. The downstream maximum-timestamp calculation can therefore bypass `crashPredatesCurrentModule` and disable an upgraded module because of an older crash.

   **Fix:** Derive crash timestamps from the original records; exclude the reconstructed file from timestamp selection and deduplication.

2. **Medium — Separate archived crashes share module evidence**  
   [net2/KernelCrashMonitor.js:309](https://github.com/firewalla/firewalla/blob/474af42294188efab08ce35eca7a39b3ac125501/net2/KernelCrashMonitor.js#L309)

   The recursive scan includes historical crash directories, but the downstream detector concatenates every record and applies the newest panic timestamp to any `xt_udp_tls` match. An old UDP-TLS crash preceding a module upgrade, followed by a newer unrelated panic, incorrectly disables the upgraded module. This remains reproducible independently of the `dmesg.txt` timestamp issue.

   **Fix:** Group records by crash and evaluate panic evidence, module presence, and timestamps within each group.

### 💡 Suggestions

Add regression coverage for both cases above. Isolated tests against the PR source reproduced both incorrect disable decisions; three control scenarios and the syntax check passed. Hardware behavior was not independently tested.

**Issue requirement coverage — [#9661](https://github.com/firewalla/firecommit/issues/9661):**

- **Met:** EFI backend loading precedes scanning, with registration checked afterward.
- **Partially met:** Systemd archive detection works, but historical records can incorrectly disable an upgraded module.
- **Explicitly deferred:** GSE panic/reboot fixes and image changes remain outside this code change, as clarified in the issue discussion.

### Verdict

**REQUEST_CHANGES** — Fix crash timestamp selection and attribution across archived records before merging.

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9681  
Head SHA: `474af42294188efab08ce35eca7a39b3ac125501`  
Checked at: 2026-09-17 17:00:13 CST

jasonlyc (approved):

PR #9682, opened 2026-09-17
fix bug, disable udp_tls desn't work in some platforms

caixingcheng:
1. some platforms enabled systemd-pstore service, so need check folder /var/lib/systemd/pstore for these platforms
2. goldv1/goldv2 pstore module was compiled as module and not loaded by default, need install pstore module for these platforms


j-sallyjin:
## PR Review Summary

Reviewed **1/1 changed files** and the linked issue, including its validation comments. Found three correctness issues, reproduced with mocked subprocess and Redis responses.

### ✅ What looks good

- Uses argument-based `execFile` calls without introducing shell interpolation.
- Checks backend registration after loading `efi_pstore`.
- Preserves the existing lock and copy-before-delete archive flow.
- The [issue’s validation report](https://github.com/firewalla/firecommit/issues/9661#issuecomment-5710416105) records successful hardware tests for both new paths.

### ⚠️ Issues found

1. **High — Live records can hide an archived UDP-TLS crash.**  
   [net2/KernelCrashMonitor.js:313–314](https://github.com/firewalla/firewalla/blob/e7645a18f5e6d081aab03f95a1953462ca8c34fc/net2/KernelCrashMonitor.js#L313)

   Returning the first nonempty source assumes the directories cannot both contain records. However, systemd can archive the previous boot’s UDP-TLS panic, followed by an unrelated Oops creating a new live record before FireMain starts. The monitor then examines only that live record and leaves UDP TLS enabled despite the archived panic. Scan both sources and deduplicate records before deciding whether to reload the module.

2. **Medium — Generated `dmesg.txt` timestamps invalidate crash-age checks.**  
   [net2/KernelCrashMonitor.js:41](https://github.com/firewalla/firewalla/blob/e7645a18f5e6d081aab03f95a1953462ca8c34fc/net2/KernelCrashMonitor.js#L41)

   The broader glob includes reconstructed `dmesg.txt`, which [systemd writes as a new file](https://github.com/systemd/systemd/blob/v249/src/pstore/pstore.c#L175-L205). Its modification time represents reconstruction, not the original crash. Because the detector selects the newest matching timestamp, an old Oops harvested after a module upgrade can incorrectly disable the replacement module. Derive crash timestamps from original records; exclude reconstructed copies from timestamp selection.

3. **Medium — Historical crashes are combined into one attribution decision.**  
   [net2/KernelCrashMonitor.js:308–314](https://github.com/firewalla/firewalla/blob/e7645a18f5e6d081aab03f95a1953462ca8c34fc/net2/KernelCrashMonitor.js#L308)

   Recursive discovery flattens every archived crash into one list. The downstream detector combines the newest panic timestamp with an `xt_udp_tls` match from **any** record. An old UDP-TLS crash predating the installed module plus a newer unrelated panic therefore disables the current module incorrectly. Group records by crash and evaluate each group’s timestamp and module evidence together.

### 💡 Suggestions

- Add regression tests for these three cases. Two control scenarios passed in the focused harness; all three edge cases produced incorrect disable decisions.
- Make existing subprocess mocks distinguish live and archived paths.

**Issue requirement coverage — [#9661](https://github.com/firewalla/firecommit/issues/9661):**

- **Met:** Conditional EFI backend loading addresses the Gold startup path.
- **Partially met:** Archived records are discoverable, but selection and attribution remain incorrect in the cases above.
- The GSE panic hang and unsupported platforms are explicitly outside this PR’s scope in the issue discussion.

### Verdict

**REQUEST_CHANGES**

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9682  
Head SHA: `e7645a18f5e6d081aab03f95a1953462ca8c34fc`  
Checked at: 2026-09-17 16:55:38 CST

jasonlyc (approved):

PR #9686, opened 2026-09-17
crystal bug fix

yijin-png:


jasonlyc (approved):

j-sallyjin:
## PR Review Summary

Reviewed all 8 changed files at `8513009`. Found two correctness issues. JavaScript and shell syntax checks passed; full integration and hardware boot tests were not run.

### ✅ What looks good

- Reset staging now uses a private temporary directory, eliminating the predictable shared runner path.
- Interface mapping reports `jq` failures and brings interfaces down before renaming.
- Clock synchronization bounds network requests and requires agreement between multiple responses.

### ⚠️ Issues found

**Medium — Preserve retries for transient HTTP failures**  
[scripts/update_assets.sh:89](https://github.com/firewalla/firewalla/blob/85130099c997d380af8c3e05d402194e4dd351be/scripts/update_assets.sh#L89-L91)

Curl exit code `22` covers HTTP errors, including temporary `503` responses and `429` throttling. Breaking on every such response causes an otherwise recoverable asset update to be skipped. [Curl documentation](https://curl.se/docs/manpage.html#--fail)

**Verified:** Against a local server returning `503` or `429` followed by a valid hash, the base code succeeded on its second request; this revision stopped after its first.

Capture the HTTP status separately. Stop immediately for permanent failures such as `404`, but retain bounded retries for transient statuses and respect `Retry-After` where applicable.

**Medium — Validate the provisioned MAC before using it**  
[platform/crystal/CrystalPlatform.js:46](https://github.com/firewalla/firewalla/blob/85130099c997d380af8c3e05d402194e4dd351be/platform/crystal/CrystalPlatform.js#L46-L49)

`readProvisionedMac()` returns arbitrary JSON values, but the caller treats every truthy value as a string. A numeric or object-valued `provision.ifmap.eth0` throws at `toUpperCase()` outside the reader’s `try/catch`, bypassing the fallback and breaking board-serial consumers, including licensing. Invalid strings are cached as device identities.

**Verified:** Numeric and object values threw `mac.toUpperCase is not a function`; `"invalid"` was accepted as `"INVALID"`.

Require a string with a valid MAC format, normalize it, and use the existing fallback when validation fails.

### 💡 Suggestions

- Add regression coverage for transient HTTP errors and malformed provisioned MAC values.
- Exercise clock synchronization with unavailable sources, conflicting timestamps, and clocks substantially ahead of or behind real time.

### Verdict

**REQUEST_CHANGES** — Fix the asset retry regression and provisioned-MAC validation before merging.

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9686  
Head SHA: `85130099c997d380af8c3e05d402194e4dd351be`  
Checked at: 2026-09-17 19:03:40 CST

PR #9687, opened 2026-09-17
sync purple 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

Reviewed **1/1 changed files** at `f0dff96`.

### ✅ What looks good

- Backend loading runs before scanning and remains protected by the existing Redis lock.
- Commands use argument arrays, avoiding shell interpolation.
- Archiving copies records before deleting the source.

### ⚠️ Issues found

- **Medium — Generated logs invalidate crash timestamps**  
  [net2/KernelCrashMonitor.js:41](https://github.com/firewalla/firewalla/blob/f0dff9654392c313f1e32036044cbb10de5a23d8/net2/KernelCrashMonitor.js#L41)  
  The `dmesg*` glob includes systemd’s reconstructed `dmesg.txt`, whose modification time reflects reconstruction rather than the original crash. [Systemd creates this file without preserving the records’ timestamps.](https://github.com/systemd/systemd/blob/v245/src/pstore/pstore.c#L175-L201) The downstream maximum-mtime calculation therefore allows a crash predating a module upgrade to disable the newer module when reconstruction happens afterward. Use the original records’ timestamps for crash age and deduplication; associate reconstructed content with those records explicitly.

- **Medium — Historical crashes are incorrectly combined**  
  [net2/KernelCrashMonitor.js:309–314](https://github.com/firewalla/firewalla/blob/f0dff9654392c313f1e32036044cbb10de5a23d8/net2/KernelCrashMonitor.js#L309-L314)  
  Recursive discovery returns every retained crash group in systemd’s archive. The downstream logic takes the newest panic timestamp but searches for `xt_udp_tls` across all returned content. Consequently, an old UDP TLS crash plus a newer unrelated panic disables the current module, even when the old crash predates its installation. Group records by crash event and evaluate each event’s timestamp and module list together.

### 💡 Suggestions

- Add regression tests for reconstructed-log timestamps and multiple historical crash groups. The existing command mocks return identical discovery results for both sources, so they cannot validate fallback behavior.
- Both findings reproduced using the actual module with mocked Redis and process calls. Syntax checks and live/archive detection controls passed.

### Verdict

**REQUEST_CHANGES** — Correct timestamp handling and crash-event grouping before merging.

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9687  
Checked at: 2026-09-17 19:56:50 CST

MelvinTo (approved):

PR #9689, opened 2026-09-17
sync gold se 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Reviewed the single changed file; found two correctness issues.
- Backend loading runs before scanning, within the existing Redis lock.
- Commands use argument arrays, and source files are cleared only after copying succeeds.

### ⚠️ Issues found

1. **Medium — Generated `dmesg.txt` can make an old crash appear newer.**  
   [net2/KernelCrashMonitor.js:41](https://github.com/firewalla/firewalla/blob/f0dff9654392c313f1e32036044cbb10de5a23d8/net2/KernelCrashMonitor.js#L41)

   The expanded glob includes systemd’s reconstructed `dmesg.txt`, whose modification time reflects reconstruction rather than the original crash. [Systemd’s implementation](https://github.com/systemd/systemd/blob/v249/src/pstore/pstore.c#L162) creates and writes this file without restoring the crash timestamp.

   The existing logic takes the newest matching file’s timestamp. If a module replacement falls between the crash and reconstruction, this bypasses the stale-crash check and incorrectly disables the replacement module. Prefer original record timestamps; exclude reconstructed files from timestamp selection or associate them with their original records.

2. **Medium — Historical crashes are combined into one UDP TLS decision.**  
   [net2/KernelCrashMonitor.js:314](https://github.com/firewalla/firewalla/blob/f0dff9654392c313f1e32036044cbb10de5a23d8/net2/KernelCrashMonitor.js#L314)

   The recursive archive scan returns every historical crash together. Downstream code takes the newest panic timestamp but searches all returned content for `xt_udp_tls`. Systemd stores separate crash groups in [separate archive directories](https://github.com/systemd/systemd/blob/v249/src/pstore/pstore.c#L222).

   Consequently, an old UDP TLS crash plus a newer unrelated panic disables the current module—even when the old crash predates its installation. Group records by crash identity and evaluate each group’s timestamp and module list together, while preserving support for split records.

### 💡 Suggestions

- Add regression tests for reconstructed-file timestamps and multiple archived crash groups. Both incorrect disable decisions were reproduced against the exact PR file using mocked dependencies.
- Add source-specific fixtures for live pstore, systemd archives, and backend-loading failures; the existing scan mock does not distinguish sources.

JavaScript syntax validation passed. Device-level startup behavior was not tested.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9689  
Checked at: 2026-09-17 19:47:56 CST

MelvinTo (approved):

PR #9690, opened 2026-09-17
sync gold pro 1.984 staging

jasonlyc:


j-sallyjin:
## PR Review Summary

### ✅ What looks good

- Reviewed **1/1 changed files** at `f0dff96`.
- The systemd archive fallback addresses a real gap in crash discovery.
- Commands use argument arrays, and scanning and cleanup remain protected by the existing Redis lock.
- **Issue requirement coverage:** No linked issue found.

### ⚠️ Issues found

1. **[Medium] Generated `dmesg.txt` timestamps bypass the stale-crash guard** — [net2/KernelCrashMonitor.js:41](https://github.com/firewalla/firewalla/blob/f0dff9654392c313f1e32036044cbb10de5a23d8/net2/KernelCrashMonitor.js#L41)

   Matching `dmesg*` includes systemd’s generated `dmesg.txt`, whose modification time reflects when systemd reconstructed it, rather than when the crash occurred. [Systemd implementation](https://github.com/systemd/systemd/blob/v249/src/pstore/pstore.c#L162-L188).

   The downstream logic takes the newest matching timestamp. Consequently, a crash preceding a module upgrade can appear newer than the installed module and incorrectly disable UDP TLS. Reproduced with crash time `100`, module time `200`, and generated-file time `300`.

   **Fix:** Derive crash timing from the original records; associate reconstructed content with those records instead of using its own modification time.

2. **[Medium] Historical crashes are combined into one attribution decision** — [net2/KernelCrashMonitor.js:307](https://github.com/firewalla/firewalla/blob/f0dff9654392c313f1e32036044cbb10de5a23d8/net2/KernelCrashMonitor.js#L307)

   The recursive archive scan returns records from multiple crashes, but the consumer combines their contents and uses the newest panic timestamp. An old crash containing `xt_udp_tls` can therefore supply the module evidence for a newer, unrelated crash.

   Reproduced without any `dmesg.txt`: a TLS crash at `100`, module installation at `200`, and unrelated crash at `300` incorrectly disable the current module.

   **Fix:** Group records by crash identity and evaluate timestamps and module evidence within each group.

### 💡 Suggestions

- Add regression tests for both cases, with separate fixtures for live pstore and systemd archives.
- Syntax checking passed. Both findings were reproduced against the exact PR file with mocked system commands and Redis; device behavior was not exercised.

### Verdict

**REQUEST_CHANGES** — fix crash timestamp selection and attribution across archived crashes before merging.

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9690  
Checked at: 2026-09-17 19:43:39 CST

MelvinTo (approved):

PR #9694, opened 2026-09-18
add missing node_modules hash pointer

jasonlyc:
Cherry-pick of ca996c494 from `master` onto `beta_25_0`.

`scripts/NODE_MODULES_REVISION.crystal` was missing, so crystal had no
node_modules hash pointer at all. It lands here as a symlink to
`NODE_MODULES_REVISION.gold`, matching what goldpro already does; `.ubt` links to
`.navy` the same way. The rest of the commit is the `upgrade_verify.sh` change
and its test coverage.

## Conflict resolution

`scripts/bootstrap.sha256sum` conflicted, and it is a checksum manifest, so
neither side could be taken wholesale — the file has to match the bytes actually
on this branch. Two lines were in the conflict hunk:

| entry | resolution | why |
| --- | --- | --- |
| `fireupgrade.sh` | kept `beta_25_0`'s `e30e3203…` | the commit does not touch this script; `master`'s `5dd8f8fe…` describes a different version of it that does not exist on this branch |
| `upgrade_verify.sh` | took the commit's `50c10455…` | the commit does modify this script, and it applied cleanly here |

Taking `master`'s side wholesale would have written a `fireupgrade.sh` hash that
does not match the file on this branch, which is the failure mode this manifest
exists to catch — bootstrap verification would reject a correct file.

Verified with `sha256sum -c bootstrap.sha256sum` in `scripts/`: all nine entries
report OK against the working tree. The net diff against `beta_25_0` is a single
line, the `upgrade_verify.sh` hash.

One line outside the hunk resolved itself correctly and is worth noting since it
shows the same pattern: `network_settings.sh` differs between the branches
(`46909ccd…` here vs `52e5e53f…` on master) and git kept this branch's value.

## Testing

Not runtime-tested; the suite here needs a live box and redis. Verification was
the checksum check above plus confirming both new symlinks resolve.

PR #9696, opened 2026-09-18
add missing node_modules hash pointer

jasonlyc:
Cherry-pick of ca996c494 from `master` onto `release_15_0`. Companion to #9694, which
takes the same commit to `beta_25_0`.

`scripts/NODE_MODULES_REVISION.crystal` was missing on this branch, so crystal had
no node_modules hash pointer at all. It lands as a symlink to
`NODE_MODULES_REVISION.gold`, matching what goldpro already does; `.ubt` links to
`.navy` the same way. The rest of the commit is the `upgrade_verify.sh` change and
its test coverage.

## Conflict resolution

`scripts/bootstrap.sha256sum` conflicted, identically to #9694. It is a checksum
manifest, so neither side could be taken wholesale — the file has to match the
bytes actually on this branch. Two lines were in the conflict hunk:

| entry | resolution | why |
| --- | --- | --- |
| `fireupgrade.sh` | kept `release_15_0`'s `e30e3203…` | the commit does not touch this script; `master`'s `5dd8f8fe…` describes a different version of it that does not exist on this branch |
| `upgrade_verify.sh` | took the commit's `50c10455…` | the commit does modify this script, and it applied cleanly here |

Taking `master`'s side wholesale would have written a `fireupgrade.sh` hash that
does not match the file on this branch, which is the failure mode this manifest
exists to catch — bootstrap verification would reject a correct file.

The `upgrade_verify.sh` value was not copied from the other branch on faith; it
was recomputed from the post-cherry-pick file here and happens to match, because
this branch's copy of that script was identical to master's before the change.

Verified with `sha256sum -c bootstrap.sha256sum` in `scripts/`: all nine entries
report OK against the working tree. The net diff against `release_15_0` is a single line,
the `upgrade_verify.sh` hash.

## Testing

Not runtime-tested; the suite here needs a live box and redis. Verification was
the checksum check above plus confirming both new symlinks resolve.


j-sallyjin:
## PR Review Summary

Reviewed all 5 changed files at `8845b5d`.

### ✅ What looks good

- Crystal and UBT symlinks reference pins matching their module repositories.
- All nine bootstrap checksums match the files, including the preserved `fireupgrade.sh` checksum.
- Both shell scripts pass syntax checks.
- Test isolation avoids the device’s real asset keyring and asset updater.

### ⚠️ Issues found

**Low — Platform-pin test depends on the host’s Node version**  
[test/test_upgrade_verify.sh:255](https://github.com/firewalla/firewalla/blob/8845b5d1eda06323ca89166cbeb03c378fd12382/test/test_upgrade_verify.sh#L255)

Overriding `FIREWALLA_PLATFORM` does not override the Node version queried by `uv_node_modules_pin_file`. On a Node 4 host, every platform therefore resolves to a `.node4` filename, although only Red has that variant.

Reproducing the new loop with simulated Node versions yielded **12 passes with Node 12**, but **11 false failures with Node 4**. Conversely, modern hosts never validate Red’s legacy pin. Make version selection deterministic for each supported platform/runtime combination.

### 💡 Suggestions

- Run the complete signature tests with `gpg` and `gpgv` installed; those tools were unavailable locally, so full-suite validation remains outstanding.
- **Issue requirement coverage:** No linked issue found. Reference #9694 is a companion PR.

### Verdict

COMMENT

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9696  
Head SHA: `8845b5d1eda06323ca89166cbeb03c378fd12382`  
Checked at: 2026-09-18 09:40:22 CST

MelvinTo (approved):

PR #9710, opened 2026-09-18
add last_ts to every event in event:log

jasonlyc:
## Design

Every event written to `event:log` now carries `last_ts`: the timestamp of the previous event of
the same *type*. Answering "when did this last happen" previously meant scanning `event:log`
backwards and re-filtering (`EventApi.getLatestEventsByType`), which is both expensive and bounded
by the log's retention and count trimming — a type that last fired a week ago is simply gone.

The hook sits in `EventApi.addEvent`, the single chokepoint for every `event:log` write. A new
`event/EventClassifier.js` maps an event to a type key; `addEvent` reads the previous event of that
key from a new `event:last:cache` hash, stamps `last_ts`, writes the new event back to the hash,
then does the `ZADD`. The hash stores the whole event, so it doubles as a "latest event per type"
lookup and the existing `EventApi.cleanCachedEventsByTime()` expiry works on it unchanged.

Classification is two-tier:

- **`FALLBACK_SETTINGS`**, hardcoded: state events key on `state_type` + `state_key` +
  `state_value`, actions on `action_type`. Not configurable.
- **Configured settings**, shipped in `net2/config.json` and replaceable from the cloud via
  `event_classifier_config`, using the same `filter` + keys shape as `BlockStatsSensor` and
  `EventSummarySensor`. Two ship today: `wan_monitor` (the `*_RTT`/`*_lossrate` actions, keyed per
  WAN) and one keying `weak_password_scan_complete` on a new `found` label.

`classify()` tries the configured settings first and falls through to the fallback.

## Decisions

**The hook is in `addEvent`, not `EventRequestHandler.sendEvent`.** `controllers/netbot.js:804`
adds `firewalla_upgrade` by calling `EventApi` directly and deliberately bypasses the handler, so a
hook in `sendEvent` would leave that event unstamped. This also means `last_ts` does *not* reach
`MSG_EVENT_GENERATED` consumers, since `sendEvent` emits its own copy built before `addEvent` runs.
Making `sendEvent` await `addEvent` would reorder the persist against the fan-out, so it doesn't.

**The fallback is not configurable.** It guarantees every state and every action gets a `last_ts`,
so configuration can only ever refine the grouping, never remove an event from tracking. A bad or
empty cloud push degrades to catch-all grouping rather than to untracked events.

**`last_ts` is clamped to never exceed the event's own `ts`.** Emission order is not `ts` order —
`ap_*`/`switch_*` events go through a bee-queue while the rest are processed inline — so a
late-arriving older event would otherwise report a "previous occurrence" that hasn't happened yet.
Such an event gets `null` and does not roll the cache backwards.

**`setting.key` is part of the cache field name.** Two settings can resolve to the same value tuple
(`wan_monitor` and the `action` catch-all both key on `action_type`), and without the prefix a
config change moving events between them would append to one field with two different meanings. The
cost is that the field name is frozen for the cache's 30-day life: renaming a setting's `key`, or
an event moving between settings, restarts that lineage at `null` and orphans the old field until
it expires.

**`HGET` + `HSET` under a per-key `AsyncLock`, not a Lua script.** An atomic `EVAL` would also close
the FireMain↔FireApi race, but the `event:log` member has to contain `last_ts`, so Lua would have to
rebuild the event JSON — a `cjson` round trip changes key order and can change number formatting for
`state_value`. The race it would fix is between FireMain's event stream and FireApi's single
`firewalla_upgrade` action, which isn't worth that. Consequence: the lookup now precedes the `ZADD`,
so two assertions in `test/test_event.js` that relied on fire-and-forget ordering now `await`
`addEvent`.

## Beyond the subject

- `InternalScanSensor.addScanCompleteEvent` gains a `found` 0/1 label so a clean scan and a scan
  with findings keep separate lineages. It flows to the app notification too, but `netbot.js:442`
  reads labels by name, so the notification text is unchanged.
- New init field `json.previousEventsByType`: the classified types whose most recent event predates
  the 24h `networkMonitorEvents` window. Without it the app cannot distinguish "this hasn't happened
  in a day" from "this has never happened". Capped at 200 entries, newest first — the hash can hold
  5000 whole events and returning it all would add megabytes to init.
- `event:last:cache` is bounded by a 5000-key cap and expires with the existing
  `latestStateEventsExpire` (30d) sweep.

## Testing

Deployed to a test box and driven through redis pub/sub: `last_ts` chaining for state and action
events, flap semantics (`1→0→1` chains to the previous `1`, not the `0`), per-device AP and switch
port lineages, per-WAN separation for all six monitor types, `found` 0/1 separation, cloud config
override and revert, cache survival across restarts, and `previousEventsByType` containing only
stale types with zero overlap against `networkMonitorEvents`. Regression-checked `eventSummary`,
`latestAllStateEvents`, the scan notification text and the `sanity_check.sh` event dump.

`test/test_event_classifier.js` is new and has not been run — mocha and chai are devDependencies and
aren't installed on a box.


j-sallyjin:
## PR Review Summary

Reviewed all 10 changed files at `b03aab5`. No linked issue found.

### ✅ What looks good

- Centralizing stamping in `addEvent()` covers direct callers and preserves caller objects.
- Out-of-order events cannot move the cached timestamp backward.
- Isolated checks confirmed normal chaining and per-WAN separation. Syntax checks passed for all changed files.

### ⚠️ Issues found

1. **High — The expiry test deletes operational cache entries.**  
   [test/test_event_classifier.js:341](https://github.com/firewalla/firewalla/blob/b03aab5ffcfa6ed1671741a3f75c6e62b6b11a85/test/test_event_classifier.js#L341)  
   `cleanCachedEventsByTime(KEY_EVENT_LAST_CACHE, now)` sweeps the entire shared hash. When run on a box, it deletes practically every existing entry, including unrelated production events, resetting their timestamp chains. The test’s field-specific cleanup cannot restore them. Use a dedicated test hash or isolated Redis database.

2. **Medium — Recent event types can disappear from both init collections.**  
   [net2/HostManager.js:1056–1058](https://github.com/firewalla/firewalla/blob/b03aab5ffcfa6ed1671741a3f75c6e62b6b11a85/net2/HostManager.js#L1056-L1058)  
   Excluding all cache entries newer than 24 hours assumes those types appear in `networkMonitorEvents`, but that collection filters event types and keeps only 250 recent events. A reboot 23 hours ago followed by 251 RTT events is absent from both collections; a recent weak-password scan is also absent. Select supplementary entries based on the types actually returned, preserving the payload cap.

3. **Medium — Malformed filters bypass the guaranteed fallback.**  
   [event/EventClassifier.js:126–130](https://github.com/firewalla/firewalla/blob/b03aab5ffcfa6ed1671741a3f75c6e62b6b11a85/event/EventClassifier.js#L126-L130)  
   Validation only checks whether the filter is empty. A setting containing `filter: {and: [true]}` passes validation but throws inside `matchFilter()`. Classification exits before reaching the fallback, so affected events are persisted without `last_ts` and their cache entries stop advancing. Recursively validate filter nodes and isolate failures to the offending setting.

4. **Medium — Concurrent new types bypass the 5,000-entry cap.**  
   [event/EventApi.js:203–207](https://github.com/firewalla/firewalla/blob/b03aab5ffcfa6ed1671741a3f75c6e62b6b11a85/event/EventApi.js#L203-L207)  
   The lock is per classifier key, while `HLEN` checks a shared hash. Different keys can all observe available capacity before inserting. An isolated reproduction starting with 4,999 entries and 64 concurrent new types ended with 5,063 entries. Make capacity checking and insertion atomic across keys; Redis can store the supplied JSON string without re-encoding it.

5. **Medium — Expiry can delete a freshly refreshed cache entry.**  
   [event/EventApi.js:335–336](https://github.com/firewalla/firewalla/blob/b03aab5ffcfa6ed1671741a3f75c6e62b6b11a85/event/EventApi.js#L335-L336)  
   The reused cleanup helper reads entries with `HSCAN`, then unconditionally deletes expired fields. If stamping refreshes a field between the scan and deletion, cleanup removes the fresh event. The next occurrence then receives `last_ts: null`. Use an atomic compare-and-delete against the scanned value, or recheck expiry atomically before deleting.

### 💡 Suggestions

Add regression tests for these cases using isolated Redis. The findings were reproduced with the actual modules and an in-memory Redis substitute; the full Mocha integration suite was not run.

### Verdict

**REQUEST_CHANGES**

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9710  
Head SHA: `b03aab5ffcfa6ed1671741a3f75c6e62b6b11a85`  
Checked at: 2026-09-18 19:04:28 CST

MelvinTo (approved):

PR #9722, opened 2026-09-20
fix OSI boot coverage for hard route rules and rebuild the pool atomically

jasonlyc:
Two independent boot-time holes in OSI's VPN kill-switch coverage, plus the `await`s the second one needs to be correct. No issue carries this, so the reasoning is here.

## Root cause: hard route rules were gated on the wrong flag

`updateOSIPool()` skipped a route rule unless the VPN client or virtual WAN group it points at had `strictVPN` on. That is not what decides kill-switch behaviour for a route rule. `Block.js` sends a `routeType=hard` rule to the PBR *hard* ipset, and both owners of that set populate it independently of `strictVPN` — `VPNClient._prepareRoutes()` and `VirtWanGroup.refreshRT()` — as does the unreachable default route in the VPN routing table that drops still-marked traffic when the tunnel is down.

So the runtime already treats a hard route rule as a kill switch whether or not the client is strict, while OSI declined to cover it:

| route rule | strictVPN | blocks on disconnect at runtime | in the OSI pool, before |
| --- | --- | --- | --- |
| hard | on | yes | yes |
| hard | off | yes | **no — leaks to the default WAN until the rule is enforced** |

The gate now keys off the client being *enabled*, which is the case the original comment was reaching for: with the client disabled nothing marks the traffic, so there is no kill switch to bridge and OSI would only be blocking traffic that is going to flow anyway. `getHighImpactfulRules()` already filters on `isRouteRuleToVPN()`, so every rule reaching this loop is `routeType=hard`.

## Root cause: the pool rebuild was destructive and non-atomic

`osi:active` and `osi:rules:active` are the only OSI state that survives a reboot — firerouter reads them to arm the ipsets before any rule is enforced. `updateOSIPool()` deleted both up front, then repopulated them over a long chain of awaits inside a `try`:

| event during the rebuild | result, before |
| --- | --- |
| power loss or reboot | next boot arms from a truncated pool |
| any throw | `catch` logs it; pool stays truncated until the next 5-minute cycle |

Both fail silently: the box comes up with devices unprotected and nothing says so. The rebuild now writes to `osi:active:tmp` / `osi:rules:active:tmp` and swaps them in with `RENAME` only once every step has succeeded; on failure the previous good pool is left alone and the temp keys are dropped. `RENAME` errors on a missing source, which is the legitimate "nothing qualifies for OSI" case, so that path clears the live key instead.

That swap is also what makes the third change load-bearing: three `processNetwork()` / `processIdentity()` calls in `processRule()` were never awaited. Harmless while the writes went straight to the live key — with the swap they could land after the `RENAME` and be lost.

## Testing

Exercised against a real OpenVPN client with the kill switch both on and off, and with faults injected at two different points in the rebuild, each compared against current behaviour.


j-sallyjin:
## PR Review Summary

Reviewed all 3 changed files at `68cde5a`.

### ✅ What looks good

- Enabled VPN clients now receive hard-route OSI coverage independently of `strictVPN`.
- The added `await`s ensure network and identity writes complete before publication.
- Temporary sets preserve existing coverage when rebuilding fails before publication.

### ⚠️ Issues found

1. **High — Concurrent rebuilds can silently erase live coverage**  
   [sensor/OSIPlugin.js:542](https://github.com/firewalla/firewalla/blob/68cde5a6b8b286bcef2c81ce0f48edb40333a2ec/sensor/OSIPlugin.js#L542)

   `releaseBrake()` schedules updates through both a periodic timer and a policy-event timer, without serializing them. Both invocations share the same temporary keys. If one invocation pauses after writing its rules while another completes, the paused invocation subsequently finds the temporary rules key missing and deletes the newly published live key.

   Reproduced with two identical overlapping rebuilds: the base implementation retained both rules; this implementation removed all rule coverage without logging an error. Serialize the entire rebuild, publication, cleanup, and cache reset.

2. **High — The two pools are not published atomically**  
   [sensor/OSIPlugin.js:643–644](https://github.com/firewalla/firewalla/blob/68cde5a6b8b286bcef2c81ce0f48edb40333a2ec/sensor/OSIPlugin.js#L643-L644)

   The first swap commits before the second begins. If the second fails, the catch block deletes temporary data but cannot restore the first live key, despite logging that the previous pool was preserved.

   For example, when a device moves from VPN assignment coverage to hard-route coverage, the first swap removes its old entry. Failure before publishing the rules pool leaves it absent from both live pools, losing protection on the next boot. This was reproduced with an injected second-swap failure.

   Publish both pools—including empty-pool deletions—in one Redis operation, such as an [atomically executed Lua script](https://redis.io/docs/latest/develop/programmability/eval-intro/).

### 💡 Suggestions

Add regression tests for overlapping updates and failures during publication, including devices moving between pools.

Syntax checks and isolated tests with mocked Redis passed for enabled-client filtering, empty pools, pre-publication failures, and delayed network/identity writes. Appliance integration was not run.

### Verdict

**REQUEST_CHANGES** — Both publication failures can remove boot-time kill-switch coverage.

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9722  
Head SHA: `68cde5a6b8b286bcef2c81ce0f48edb40333a2ec`  
Checked at: 2026-09-20 18:54:09 CST

j-sallyjin:
## PR Review Summary

Reviewed all 3 changed files at `81b51ea`. One high-severity issue remains.

### ✅ What looks good

- Enabled VPN clients receive hard-route OSI coverage regardless of `strictVPN`.
- The new lock fixes overlapping rebuilds and protects the shared temporary keys and cache.
- Added `await`s ensure network and identity writes finish before publication.
- Failures before publication preserve both existing pools.

**Issue requirement coverage:** No linked issue found. The PR’s stated atomic-publication goal remains incomplete.

### ⚠️ Issues found

**High — Publish both OSI pools atomically**  
[sensor/OSIPlugin.js:653–654](https://github.com/firewalla/firewalla/blob/81b51eaab52f9b45e02ef835c19b9e65c3dfb689/sensor/OSIPlugin.js#L653-L654)

The first swap commits before the second begins. If the second operation fails, cleanup deletes the remaining temporary data without restoring the first live pool, despite logging that the previous pool was preserved.

For a device moving from VPN-assignment coverage to hard-route coverage, the first swap removes its old entry. Failure publishing the rules pool then leaves it absent from **both live pools**, losing OSI protection on the next boot.

Reproduced against the current implementation with an injected second-publication failure, both when the first operation renames a populated set and when it deletes an empty pool.

Publish both pools—including empty-pool deletions—in one atomic Redis operation, such as a [Lua script](https://redis.io/docs/latest/develop/programmability/eval-intro/). The process-local lock cannot protect against interruption between Redis commands.

### 💡 Suggestions

Add regression tests for publication failures while devices move between pools, covering empty and populated replacements.

Syntax checks and isolated tests with mocked Redis passed for filtering, serialization, delayed writes, empty pools, and pre-publication failures. Appliance integration was not run.

### Verdict

**REQUEST_CHANGES** — The remaining publication gap can remove boot-time kill-switch coverage.

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9722  
Head SHA: `81b51eaab52f9b45e02ef835c19b9e65c3dfb689`  
Checked at: 2026-09-21 13:15:44 CST

MelvinTo (approved):

PR #9738, opened 2026-09-22
Gold Pro 1.984 Alpha Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

Review targets commit `6a554550` of this 354-file deployment PR, with findings in firewall enforcement, release verification, and DNS interception.

### ✅ What looks good

- Target-list cleanup checks remaining policy and exception references and unregisters cloud-cache polling.
- AP throughput formatting handles missing input and includes focused tests for Ethernet ports, Wi-Fi bands, and uplink information.
- Syntax checks passed for 263 changed JavaScript and shell files.

### ⚠️ Issues found

1. **High — Cached firewall drops survive policy changes.**  
   [control/iptables_common.sh:614](https://github.com/firewalla/firewalla/blob/6a554550c9b1b2517279562e5615cefaccd15473/control/iptables_common.sh#L614)

   Blocking an established connection sets connmark bit `0x200`. Subsequent packets hit this fast-path drop before current policies are evaluated. However, `BlockControl.refreshConnmark()` only clears `0x80000000`, and removing a block does not schedule a refresh. An established ICMP/UDP flow can therefore remain blocked after the blocking rule is removed, until its conntrack entry expires.

   Clear cached deny decisions whenever policy changes can permit traffic, including block removal and allow/bypass activation, for both IPv4 and IPv6.

2. **High — An unsigned tag alias bypasses the minimum release version.**  
   [scripts/upgrade_verify.sh:313](https://github.com/firewalla/firewalla/blob/6a554550c9b1b2517279562e5615cefaccd15473/scripts/upgrade_verify.sh#L313)

   Signature verification authenticates the tag object, but the version comparison reads the external Git ref name. Someone able to modify repository refs can point a new, higher-version ref at an old signed tag object without invalidating its signature.

   A signed fixture reproduced this: version `1.0` was rejected against floor `2.0`, then accepted after the identical tag object was also referenced as `goldpro-alph-v99.0`.

   Extract the version from the verified tag object’s internal `tag` header and reject mismatches between that header and the ref name.

3. **Medium — Domain-specific DNAT cannot intercept ordinary DNS-over-TCP connections.**  
   [extension/dnsmasq/dnsmasq.js:1697](https://github.com/firewalla/firewalla/blob/6a554550c9b1b2517279562e5615cefaccd15473/extension/dnsmasq/dnsmasq.js#L1697)

   The TCP rule requires the DNS question bytes to appear in the packet. The initial SYN contains no DNS question, while NAT determines translation when establishing the connection; later query packets do not re-evaluate this rule. Consequently, with DNS Booster disabled, a TCP query for `fire.walla` to an external resolver bypasses local dnsmasq. The IPv6 rule has the same limitation. This follows [Netfilter’s documented NAT behavior](https://www.netfilter.org/documentation/HOWTO/NAT-HOWTO-5.html).

   Handle domain selection at the DNS layer, with any TCP redirection established during connection setup.

### 💡 Suggestions

- Add regression coverage for block→unblock on an existing connection, aliased signed tags, and TCP queries for `fire.walla` with DNS Booster disabled.
- Run Linux networking integration tests before deployment. Validation here included syntax checks, the signed-tag fixture, and connmark-mask checks; hardware behavior was not exercised.

**Issue requirement coverage**

- **Met by inspection — [#9107](https://github.com/firewalla/firecommit/issues/9107) and [#9299](https://github.com/firewalla/firecommit/issues/9299):** The documented lifecycle fixes cover final-reference deactivation, MSP exception cleanup, and stopping independent cloud-cache polling.
- **Met by inspection — [#9381](https://github.com/firewalla/firecommit/issues/9381):** AP live statistics expose aggregated throughput, Ethernet ports, Wi-Fi bands, and uplink RSSI through the requested response structure.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9738  
Head SHA: `6a554550c9b1b2517279562e5615cefaccd15473`  
Checked at: 2026-09-22 12:24:55 CST

PR #9739, opened 2026-09-22
Orange 1.984 Alpha Deployment

jasonlyc:


j-sallyjin:
## PR Review Summary

Reviewed commit `99dd68f`. Found three actionable defects. Syntax checks passed for 263 changed JavaScript and shell files; hardware integration was not tested.

### ✅ What looks good

- Bloom-filter updates validate checksums and replace files atomically.
- Target-list cleanup now considers both rules and exceptions, including MSP unbinding.
- SSH log parsing validates extracted IP addresses and avoids shell interpolation.

### ⚠️ Issues found

1. **High — Tag aliases bypass the minimum release version.**  
   [scripts/upgrade_verify.sh:311–314](https://github.com/firewalla/firewalla/blob/99dd68f03c914dbddfb6fb27eccf79d245e84a23/scripts/upgrade_verify.sh#L311-L314)

   The version comes from the mutable tag reference, while the signature protects the tag object’s embedded name. An attacker controlling repository refs can alias an older, legitimately signed tag as `orange-alpha-v9.999.999`, allowing that older commit past the version floor without the signing key. A focused reproduction confirmed rejection before aliasing and acceptance afterward, with signature acceptance mocked for the unchanged object.

   **Fix:** Extract the version from the verified tag payload and require its embedded name to match the selected reference.

2. **High — Intel queue saturation suppresses known-threat alarms.**  
   [net2/DNSManager.js:167–176](https://github.com/firewalla/firewalla/blob/99dd68f03c914dbddfb6fb27eccf79d245e84a23/net2/DNSManager.js#L167-L176)

   Once inline-intel migration is ready, this branch skips lookup unconditionally. However, `appendNewFlow()` returns `null` when its queue reaches 2,000 entries, and BroDetect still persists those flows without a category. `FlowMonitor.checkFlowIntel()` then skips them—even when cached intelligence already identifies the destination as malicious.

   A focused reproduction recovered the cached `intel` category before migration readiness but performed zero lookups afterward.

   **Fix:** Distinguish successful enrichment from skipped/failed enrichment, and retain lookup or retry behavior for incomplete snapshots.

3. **Medium — Large SSH logs silently disable guessing detection.**  
   [sensor/AuthLogPlugin.js:47–48](https://github.com/firewalla/firewalla/blob/99dd68f03c914dbddfb6fb27eccf79d245e84a23/sensor/AuthLogPlugin.js#L47-L48)

   `execFileAsync()` buffers the entire journal output using its default limit. When exceeded, the catch converts the failure to `null`, discarding all attempts in that interval. Testing the plugin with synthetic output produced an alarm for 31 failures, but none for 16,000 failures: approximately 1.15 MB triggered `ERR_CHILD_PROCESS_STDIO_MAXBUFFER`.

   **Fix:** Stream and aggregate journal records with bounded memory, and report collection failures instead of silently treating them as empty results.

### 💡 Suggestions

- Add regression coverage for tag aliases, missing intel snapshots under queue pressure, and oversized SSH logs.
- **Issue requirement coverage — commit-linked [firecommit#9107](https://github.com/firewalla/firecommit/issues/9107):**
  - **Met:** Last-reference deletion and MSP-unbind cleanup now deactivate target lists and unregister their cloud cache.
  - **Partially met:** Authorization revocation while a reference remains still lacks explicit suppression or substantial backoff. Fetch failures return without removing refresh registrations. Cover this lifecycle case before treating recurring unauthorized polling as fully resolved.

### Verdict

REQUEST_CHANGES

---
Repo: `firewalla/firewalla`  
PR: https://github.com/firewalla/firewalla/pull/9739  
Head SHA: `99dd68f03c914dbddfb6fb27eccf79d245e84a23`  
Checked at: 2026-09-22 12:10:57 CST

